PR #891, opened 2026-05-08
fix(tui): hide INPUTS header when workflow has no inputs

### LINKED ISSUE #873: feat(tui): hide workflow picker INPUTS header when workflow has no inputs
## Summary

The workflow picker's input phase always renders an `INPUTS` section header (with an optional `n / m` field counter), even when the selected workflow declares zero inputs. For input-less workflows the header sits above an empty scrollbox, taking up vertical space in the picker for no information gain. Hide the header — and its surrounding spacers — when `workflow.inputs.length === 0`.

The relevant code is in `packages/atomic-sdk/src/components/workflow-picker-panel.tsx` inside `InputPhase` (the header row at lines ~1033–1045, gated today only by `isStructured` for the counter text). `isStructured` already exists at line 944 (`workflow.inputs.length > 0`) and is the right condition to gate the entire header row + its leading and trailing spacer `<box height={…} />` siblings on.

## Functional Goals

- When the focused workflow has no inputs, the `INPUTS` row at `workflow-picker-panel.tsx:1033` is not rendered.
- The leading `<box height={2} />` (line 1031) and trailing `<box height={1} />` spacer (line 1046) are also skipped so the scrollbox does not gain a phantom gap above it.
- When the focused workflow has at least one input the panel renders identically to today (header text, `i / n` counter, spacing).
- No change to keyboard navigation, focus model, or to the `confirm` phase (`ConfirmModal` at line 1089) — only the structural inputs phase is affected.

## Non-Goals

- Reworking the `ConfirmModal` layout or the workflow summary card above the header.
- Re-styling the `INPUTS` header for the populated case.
- Changing the empty-state copy elsewhere in the picker (e.g. when no workflows match the filter).

## Cross-platform considerations

Purely a TUI render-tree change — no IO, process spawn, path, or FFI surface is touched. Behavior is identical across:

- **Linux**: no platform-specific concerns.
- **macOS**: no platform-specific concerns.
- **Windows**: no platform-specific concerns; OpenTUI handles the rendering identically.

## Cross-agent considerations

Provider-agnostic. The picker reads `workflow.inputs` from the workflow definition (`PickerWorkflow`) before any agent is selected, so the change does not touch `packages/atomic-sdk/src/providers/{claude,copilot,opencode}.ts` or any session/runtime/hook code.

- **Claude Code**: unaffected.
- **GitHub Copilot CLI**: unaffected.
- **OpenCode**: unaffected.

## Test Plan

- **Unit** — extend `tests/sdk/components/workflow-picker-panel.test.tsx`:
  - With a workflow whose `inputs` is `[]`: the rendered `InputPhase` does not contain the literal `INPUTS` text, and the scrollbox sits flush below the workflow summary card (no phantom spacer rows above it).
  - With a workflow whose `inputs.length > 0`: the `INPUTS` header and `1 / n` counter are present, matching today's snapshot.
- **Integration** — manual TUI smoke: launch the picker (`bun packages/atomic/src/index.ts`), focus a builtin input-less workflow, and confirm the header is gone; focus a workflow with inputs and confirm the header re-appears unchanged.
- **Cross-platform smoke** — covered by the existing `linux-x64` / `darwin-arm64` / `windows-x64` matrix in `.github/workflows`; no new entries needed.
- **Cross-agent matrix** — not applicable (no provider code changes).

prantikmedhi:
Fixes #873.

When a workflow declares zero inputs, the picker was still rendering the `INPUTS` header (plus spacer rows) above an empty prompt scrollbox. This gates the whole header row + spacers on `workflow.inputs.length > 0`, so input-less workflows keep the same layout but without the dead space.

Tests: updated the existing workflow picker panel test to assert the header is absent for input-less workflows.

### DIFF
--- a/packages/atomic-sdk/src/components/workflow-picker-panel.tsx
+++ b/packages/atomic-sdk/src/components/workflow-picker-panel.tsx
@@ -1028,22 +1028,26 @@ function InputPhase({
         </text>
       </box>
 
-      <box height={2} />
+      {isStructured ? (
+        <>
+          <box height={2} />
 
-      <box flexDirection="row" height={1}>
-        <text>
-          <span fg={theme.textDim}>
-            <strong>INPUTS</strong>
-          </span>
-        </text>
-        <box flexGrow={1} />
-        <text>
-          <span fg={theme.textDim}>
-            {isStructured ? `${focusedFieldIdx + 1} / ${fields.length}` : ""}
-          </span>
-        </text>
-      </box>
-      <box height={1} />
+          <box flexDirection="row" height={1}>
+            <text>
+              <span fg={theme.textDim}>
+                <strong>INPUTS</strong>
+              </span>
+            </text>
+            <box flexGrow={1} />
+            <text>
+              <span fg={theme.textDim}>
+                {`${focusedFieldIdx + 1} / ${fields.length}`}
+              </span>
+            </text>
+          </box>
+          <box height={1} />
+        </>
+      ) : null}
 
       <scrollbox
         ref={scrollboxRef}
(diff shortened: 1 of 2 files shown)


lavaman131:
Thank you for your contribution! Will check this in and mark as resolved.

PR #900, opened 2026-05-09
fix(sdk): externalize @opentui/core + add create-atomic-cli scaffolder

### LINKED ISSUE #898: Workflow capsules bundled with `Bun.build` crash on compiled-host `_orchestrator-entry` due to OpenTUI's platform-native dynamic import
## Summary

Third-party compiled hosts that ship standalone workflow capsules (built via `Bun.build({ format: "esm", target: "bun" })` and dropped on disk somewhere outside `node_modules`) crash silently when the SDK's auto-dispatch path re-exec's the host as `_orchestrator-entry` and dynamic-imports the capsule. The crash is not a workflow bug — it's OpenTUI's platform-native loader failing to resolve at runtime from the capsule's location.

## Repro

A consumer with this shape:

- A compiled host built via `bun build --compile` that imports `@bastani/atomic-sdk` and uses `runWorkflow`.
- One or more workflow capsules bundled separately via `Bun.build({ format: "esm", target: "bun" })`, written to a runtime location like `~/.<host-name>/workflows/<wf>.mjs` and pointed to via the `source` field in a `WorkflowDefinition`.
- The host invokes `runWorkflow({ workflow, ... })`. The SDK's `resolveDispatcher()` correctly auto-defaults to `process.execPath` (the host binary), creates the tmux session, and the orchestrator pane runs `<host-binary> _orchestrator-entry <name> <agent> <inputsB64> <capsule-path>`.

Direct repro of the crash:

```
$ <host-binary> _orchestrator-entry <name> <agent> e30= /path/to/<wf>.mjs

[atomic-sdk:_orchestrator-entry] ResolveMessage: Cannot find module
  '@opentui/core-linux-x64/index.ts'
  from '/path/to/<wf>.mjs'
```

Because the error is not an `InvalidWorkflowError`, `lib/auto-dispatch.ts` writes to stderr and `process.exit(1)`. tmux's `pane-exited` hook kills the whole session within ~500ms. The parent foreground attach then errors with `tmux switch-client failed: no current client` (or `open terminal failed: not a terminal`, depending on stdio). To the user it looks like the binary "exits silently" — and `--detach` appears to "work" only because the session-kill happens after the detach banner prints.

## Root cause

The capsule's import chain pulls in the runtime:

```
@bastani/atomic-sdk/workflows
  → primitives/run.ts
    → static `import "../lib/auto-dispatch.ts"`
      → runtime/orchestrator-entry.ts
        → runtime/executor.ts
          → @opentui/core
```

Because `Bun.build` follows the static graph, the capsule ends up containing a copy of OpenTUI's runtime — including its template-literal native loader:

```js
await import(`@opentui/core-${process.platform}-${process.arch}/index.ts`)
```

That import is bare-specifier, runtime-resolved. Bun walks up from the **importer's** directory (`~/.<host-name>/workflows/`), not from where `@bastani/atomic-sdk` lives in the consumer's `node_modules`. None of those parent dirs has a `node_modules/@opentui/core-*`, so resolution fails.

Note that this is **not** fixed by ensuring `@opentui/core` is in the consumer's deps — atomic-sdk already declares it, so it's installed in the consumer's `node_modules/@opentui/`. The problem is that the bundled capsule lives on disk in a location far from any `node_modules` tree, and Node/Bun module resolution is a directory walk from the importer, not a global lookup.

The host binary itself doesn't hit this because `bun build --compile` correctly handles `optionalDependencies` and embeds the matching platform-native package into the binary's bunfs. The issue is specific to the separately-bundled capsule that the SDK then dynamic-imports from disk.

## Recommended fix: adopt OpenTUI's runtime-plugin embedder pattern

OpenTUI already ships the exact mechanism for this case:

```
@opentui/core/runtime-plugin           — build-time BunPlugin
@opentui/core/runtime-plugin-support   — runtime registration helper
```

```ts
export interface CreateRuntimePluginOptions {
  core?: RuntimeModuleEntry;
  additional?: Record<string, RuntimeModuleEntry>;
  rewrite?: RuntimePluginRewriteOptions;
}
export declare function createRuntimePlugin(input?: CreateRuntimePluginOptions): BunPlugin;
```

Two changes inside atomic-sdk:

### 1. Expose a build-time plugin for capsule consumers

```ts
// packages/atomic-sdk/src/build/runtime-plugin.ts
import { createRuntimePlugin } from "@opentui/core/runtime-plugin";

export function atomicRuntimePlugin() {
  return createRuntimePlugin({
    rewrite: {
      nodeModulesRuntimeSpecifiers: true,
      nodeModulesBareSpecifiers: true,
    },
  });
}
```

Re-exported as `@bastani/atomic-sdk/build-plugin`. Consumers add one line to their capsule build:

```ts
await Bun.build({
  entrypoints: [...],
  plugins: [atomicRuntimePlugin()],
  // everything else unchanged
});
```

The plugin rewrites `@opentui/core` and the platform-specific specifiers in the capsule into runtime-resolved module IDs. The capsule no longer contains OpenTUI bytes — it just contains a stub that asks the runtime for module ID `<runtime>:@opentui/core`.

### 2. Register the host's loaded OpenTUI in the SDK's runtime entry

`lib/auto-dispatch.ts` (or a sibling module imported alongside it) calls `ensureRuntimePluginSupport({ core: () => import("@opentui/core") })` once at module load. When the host binary loads, its already-resolved OpenTUI gets registered. When the capsule's rewritten import fires, it lands on the host's instance — which is correctly platform-bundled because `bun build --compile` handles `optionalDependencies` correctly for the host.

This is invisible to the consumer. They never touch `@opentui/core` directly; they don't add it to their `dependencies`. The embedder plumbing is internal to atomic-sdk.

## Why this beats the alternatives

- **"Vendor OpenTUI inside `dist/`"** → 6× install bloat (all platform-native packages), version skew between host and capsule copies of OpenTUI in the same process.
- **"Document a `node_modules` install dance for capsule consumers"** → pushes the SDK's runtime concerns onto every consumer.
- **"Reorder `resolveWorkflowDefinition` to check the host registry first"** is a useful complementary fix — it lets compiled hosts that call `hostLocalWorkflows([...])` skip the dynamic-import path entirely — but doesn't solve the problem for hosts that legitimately want disk-resident capsules (hot-reload of new workflows without recompiling, third-party-distributed workflows, etc.). The runtime-plugin pattern fixes both shapes.

## Future-proofing for widget APIs

Once the embedder pattern is in place, a future `@bastani/atomic-sdk/widgets` (or similar) that lets workflow authors import OpenTUI directly works the same way: types are erased at compile time, runtime imports route through the host's instance via the rewritten specifier. No "do I bundle widgets here or not" decisions, no per-API-surface bookkeeping.

## Deliverables

- [ ] `packages/atomic-sdk/src/build/runtime-plugin.ts` — thin wrapper over `createRuntimePlugin` with sensible defaults.
- [ ] Subpath export `@bastani/atomic-sdk/build-plugin` in `package.json#exports`.
- [ ] `ensureRuntimePluginSupport` registration call in the host runtime entry path.
- [ ] Build-graph guard test: bundle a capsule using only `defineWorkflow` + `atomicRuntimePlugin()` and assert the output contains no `@opentui/core` strings and no `core-${process.platform}` template-literal import.
- [ ] Docs note in the SDK README: compiled-binary hosts that ship standalone capsules must use `atomicRuntimePlugin()` in their capsule build.

lavaman131:
## Summary

Closes #898. Fixes the OpenTUI runtime-plugin identity assertion crash in Bun-compiled capsules, ships a new \`@bastani/create-atomic-cli\` project scaffolder, and patches two environment-leak bugs surfaced by the expanded test suite.

### fix(sdk): OpenTUI runtime-plugin install (#898)
- **\`auto-dispatch.ts\`**: installs \`ensureRuntimePluginSupport\` in the \`_orchestrator-entry\` subprocess, gated by a \`globalThis\` sentinel so re-entered capsules skip the second call and avoid OpenTUI's identity assertion
- **\`orchestrator-entry.ts\`**: adds \`isOpenTuiResolveError()\` to detect \`@opentui/core*\` resolve failures and rethrows with a clear actionable message directing authors to add \`external: ["@opentui/core", ...]\` to their \`Bun.build()\` call
- New e2e tests spawn child Bun processes (per RFC §5.4) to keep parent process plugin state clean; includes lint-guard test for sync \`require()\` safety

### feat(create): \`@bastani/create-atomic-cli\` scaffolder
- New \`packages/create-atomic-cli\` package — run \`bun create @bastani/atomic-cli <name>\` to scaffold either:
  - \`.atomic/workflows/<name>/\` skeleton (atomic-workflow template)
  - Standalone Commander-embed CLI (standalone-cli template)
- Handles \`settings.json\` merging, supports \`-y\` / \`--yes\` for non-interactive use, and mirrors \`bun init\`'s CLI shape
- Adds \`examples/compiled-cli\` as a reference implementation

### fix(file-discovery): strip \`GIT_*\` env vars before spawning subprocesses
- \`file-discovery.ts\`: filters \`GIT_*\` env vars from the child environment passed to \`git ls-files\` and \`rg\`, preventing git hooks (e.g. \`pre-push\`) from injecting \`GIT_DIR\` and causing child processes to walk the parent repo instead of \`cwd\`
- Fixes 6 spurious test failures in \`preflight.real-spawn.test.ts\` and \`file-discovery.test.ts\` under real \`git push\`

### fix(deps): lock file regenerated
- \`bun.lock\` updated to include the \`examples/compiled-cli\` workspace entry and \`@img/sharp\` platform binaries pulled in transitively

### docs
- \`workflow-creator\` skill updated to teach \`bun create @bastani/atomic-cli\` as the primary scaffold flow; second-pass audit adds scaffolded layout coverage and fixes stale cross-references
- All example READMEs normalized to a consistent structure; top-level \`examples/README.md\` pattern catalog added
- \`packages/atomic-sdk/docs/disk-resident-workflows.md\` added

## Test plan

- [x] \`bun run test:coverage\` clean across 157 files / 2546 tests
- [x] \`bun run typecheck\` clean
- [x] \`bun run lint\` clean
- [x] pre-push hook (\`prek\` → \`bun run test:coverage\`) passes under real \`git push\` (GIT_DIR env-leak case)
- [ ] Manual: scaffold a fresh workflow via \`bun create @bastani/atomic-cli\` and verify end-to-end with \`atomic workflow refresh\`
- [ ] Manual: build a capsule that imports an OpenTUI React workflow and run it on a compiled host — confirm runtime-plugin path resolves with no \`_orchestrator-entry\` crash


lavaman131:
Closing — pivoting to explore an atomic 2.0 daemon refactor (see specs/2026-05-09-ui-server-bun-native.md). The original scope of this PR is subsumed by the broader rewrite. Remote branch preserved for reference.

PR #1052, opened 2026-05-25
fix(mcp): remove eager OAuth callback init from session startup

### LINKED ISSUE #1045: Suppress non-blocking MCP OAuth initialization errors in graph orchestrator panel
## Problem

The graph orchestrator panel currently surfaces `MCP OAuth initialization failed` messages inline when an MCP server requires login and the user is not authenticated.

This error is non-blocking: the orchestrator can continue, but the visible message looks like an execution failure and confuses the user.

## Example

In the orchestrator panel, repeated entries appear as:

Image

```text
MCP OAuth initialization failed
```

This occurs when the user is using an MCP server that needs OAuth and they are not logged in to that server.

## Expected behavior

- MCP servers should be lazy loaded like in the main chat as well and use the same suppression logic as in the main chat MCP extension
- Suppress this specific non-blocking OAuth initialization failure from the graph orchestrator panel.
- Avoid presenting it as an orchestrator/task error.
- If we need to preserve observability, keep it in debug logs or a lower-noise diagnostics surface.

## Notes

- This should only suppress the known non-blocking auth-initialization message.
- Actual blocking MCP errors should still be surfaced clearly.

flora131:
## Summary

Stops the MCP extension from eagerly starting the OAuth callback server on every session start. Each workflow stage creates a child `AgentSession` that loads the MCP extension; the previous eager `initializeOAuth()` call in the `session_start` handler attempted to bind the OAuth callback server immediately, and any port-bind failure was written to `console.error` — making a non-blocking infrastructure detail appear as a workflow/stage execution error. The fix removes that eager initialization entirely. Explicit auth flows (`/mcp-auth`, MCP panel auth, `settings.autoAuth` paths) already call `ensureCallbackServer()` on demand, so no auth capability is lost.

Also adds a deterministic `StatusWriter.flush()` drain method to eliminate timing-based `sleep()` calls from integration tests, removing a source of intermittent CI flakiness.

Closes #1045

## Changes

- **`packages/mcp/index.ts`**: Removes the eager `initializeOAuth()` call (and its `console.error` fallback) from the `session_start` handler, making OAuth callback server startup fully lazy. Adds explicit TypeScript types to `execute` handler parameters (`_toolCallId: string`, `_signal: AbortSignal | undefined`, `_onUpdate: AgentToolUpdateCallback<...>`, `_ctx: ExtensionContext`).
- **`packages/workflows/src/extension/status-writer.ts`**: Adds `flush(): Promise<void>` to the `StatusWriter` interface and its no-op stub, backed by a `drainPromise` sentinel via `ensureDrain()` that serializes write cycles without re-entrancy hazards.
- **`test/unit/mcp-oauth-startup.test.ts`**: New regression test asserting that MCP session startup leaves the OAuth callback server stopped and emits no `MCP OAuth initialization failed` to `console.error`, even when ports are occupied.
- **`test/integration/runtime-tunables.test.ts`**: Replaces 8 `sleep(N)` calls with `writer.flush()` throughout the `statusFile` suite, making disk-write assertions deterministic on slow CI runners.
- **`packages/mcp/CHANGELOG.md`**: Adds `### Fixed` entry under `[Unreleased]`.
- **`specs/`**: RFC and root-cause analysis covering prior art, design rationale, and alternatives considered for issue #1045.

## Why

Workflow stages create child `AgentSession`s that each load the MCP extension. On every session start, `initializeOAuth()` tried to bind the OAuth callback server. Port-bind failures were caught and immediately re-emitted as `console.error("MCP OAuth initialization failed")`. In orchestration contexts, that stderr line surfaced in stage/subagent output and made a non-blocking port collision look like a workflow execution error.

Since explicit auth flows (`/mcp-auth`, MCP panel, `settings.autoAuth`) already call `ensureCallbackServer()` before starting any browser auth sequence, removing the eager startup call eliminates the noise without affecting real auth flows or masking genuine errors.

## Validation

```sh
AGENT=1 bun test test/unit/mcp-oauth-startup.test.ts test/unit/mcp-init-statusbar.test.ts test/unit/stage-chat-view.test.ts
bun run typecheck
bun run lint && bun run test:unit
```


flora131:
# Implementation Notes

Task: Address the following gh issue: https://github.com/flora131/atomic/issues/1045

## Running Notes

- Iteration 1 implemented the preferred source-suppression path: MCP no longer eagerly calls `initializeOAuth()` during `session_start`, so the non-blocking `MCP OAuth initialization failed` `console.error` path was removed instead of merely downgraded.
- Kept `shutdownOAuth()` cleanup during MCP session restart/shutdown because explicit `/mcp-auth` and MCP panel auth flows can still start the callback server lazily.
- Added workflow-visible sanitization as an exact-prefix-only defense for `MCP OAuth initialization failed` followed by LF or CRLF. The original workflow failure diagnostic message is preserved; only `userMessage` is sanitized.
- Explicit/blocking errors remain visible: `MCP initialization failed: ...`, explicit `OAuth authentication failed ...`, and same-line/colon variants of the benign message are not stripped.
- No separate MCP adapter lifecycle test was added. The implementation subagent judged a full fake extension startup harness too brittle for this small removal; validation instead used focused workflow-failure tests, existing MCP status/stage tests, typecheck, and diff/source inspection.
- Validation passed: `bun test test/unit/workflow-failures.test.ts test/unit/mcp-init-statusbar.test.ts test/unit/stage-chat-view.test.ts` reported 79 passing tests, and `bun run typecheck` passed.
- Validation observation: if historical text consists only of `MCP OAuth initialization failed` plus a newline and no real trailing error, the sanitizer returns an empty workflow-facing message. This is not blocking for iteration 1 because the source emission was removed, but it may be worth a future UX decision if historical standalone entries need special rendering.

PR #1086, opened 2026-05-28
fix(workflows): warn before session switch stops in-flight runs

### LINKED ISSUE #1082: Warn before `/new` kills running workflows and clears workflow history
## Summary

The `/new` slash command should warn the user before creating a new session when there are running workflows. Creating a new session kills all in-flight workflows and clears workflow history (since workflows are tied to a single session).

## Motivation

Workflows are scoped to a single session. Compaction stays within the same session, so it's safe. But `/new` starts a brand new session, which:

- Kills any workflows currently in flight
- Clears workflow history for the previous session

Today this happens silently and users can accidentally lose long-running workflow work by running `/new` without realizing the consequences.

## Proposed behavior

When the user runs `/new`:

1. Check whether there are any workflows currently running in the active session.
2. If **yes**, show a confirmation prompt warning that proceeding will:
   - Kill all running workflows in the current session
   - Clear workflow history tied to the session
   - Require explicit confirmation before the new session is created
3. If **no running workflows**, behave exactly as today (no extra prompt — don't add friction when there's nothing to lose).

The warning should *only* surface when there are actively running workflows. Completed / killed / errored workflows on their own should not trigger the prompt (they would be addressable by improvements tracked separately).

## Acceptance criteria

- [ ] `/new` with at least one running workflow shows a confirmation warning describing the consequences and requires the user to confirm before proceeding.
- [ ] `/new` with no running workflows behaves exactly as it does today (no prompt).
- [ ] Confirming the warning kills the workflows and starts the new session as expected.
- [ ] Declining the warning leaves the current session, workflows, and history intact.

flora131:
## Summary

Adds a `session_before_switch` guard that intercepts `/new` and `/resume` session transitions when workflows are still in flight, prompting the user to confirm before proceeding. Without this, switching sessions silently killed active runs and cleared session-scoped workflow history with no opportunity to cancel.

## Changes

- **`session_before_switch` handler** registered in the extension factory; fires only for `reason: "new"` and `reason: "resume"` — other transitions (e.g. `"fork"`) pass through silently.
- **In-flight detection** counts runs without `endedAt` from the store, matching the same set that `session_start` would stop on switch.
- **Confirmation dialog** displays a titled prompt describing both consequences (stops in-flight runs, clears session-scoped history); action and message labels are tailored to the transition reason.
- **`{ cancel: true }` return** when the user declines, accompanied by an info toast; `undefined` return when the user confirms so the switch proceeds normally.
- **Fail-open semantics**: skips the prompt when there are no in-flight runs, when `ctx.ui.confirm` is absent, or when the confirm call throws — ensuring headless/automation callers are never wedged.
- **Tests** (`test/unit/extension.test.ts`): 7 new unit tests covering confirm path, cancel path, plural noun rendering, confirm-throws fail-open, no in-flight runs, unrelated switch reason, and missing confirm UI.
- **Changelog** updated under `[Unreleased] > Fixed` in `packages/workflows/CHANGELOG.md`.

## Breaking Changes

None.

## Test Plan

```sh
bun test test/unit/extension.test.ts
bun run typecheck
bun run lint
bun run test:unit
```

Closes #1082

### DIFF
--- a/packages/workflows/CHANGELOG.md
+++ b/packages/workflows/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ## [Unreleased]
 
+### Fixed
+
+- Warn before starting a new session when workflows are still running, allowing users to cancel before those runs are killed and current-session workflow history is cleared ([#1082](https://github.com/flora131/atomic/issues/1082)).
+
 ## [0.8.18] - 2026-05-27
 
 ### Changed
--- a/packages/workflows/CHANGELOG.md
+++ b/packages/workflows/CHANGELOG.md
@@ -8,7 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ### Fixed
 
-- Warn before starting a new session when workflows are still running, allowing users to cancel before those runs are killed and current-session workflow history is cleared ([#1082](https://github.com/flora131/atomic/issues/1082)).
+- Warn before starting or resuming another session when workflows are still in flight, allowing users to cancel before those runs are killed and current-session workflow history is cleared ([#1082](https://github.com/flora131/atomic/issues/1082)).
 
 ## [0.8.18] - 2026-05-27
 
--- a/packages/workflows/src/extension/index.ts
+++ b/packages/workflows/src/extension/index.ts
@@ -3260,22 +3260,25 @@ function factory(pi: ExtensionAPI): void {
       const reason = typeof event === "object" && event !== null && "reason" in event
         ? (event as { readonly reason?: string }).reason
         : undefined;
-      if (reason !== "new") return undefined;
+      if (reason !== "new" && reason !== "resume") return undefined;
 
-      const runningWorkflowCount = store.runs().filter((run) => run.endedAt === undefined).length;
-      if (runningWorkflowCount === 0) return undefined;
+      const inFlightWorkflowCount = inFlightRunCount();
+      if (inFlightWorkflowCount === 0) return undefined;
 
-      const confirmNewSession = ctx?.ui?.confirm;
-      if (typeof confirmNewSession !== "function") return undefined;
+      const confirmSessionSwitch = ctx?.ui?.confirm;
+      // Headless/non-interactive callers intentionally fail open so automation cannot wedge.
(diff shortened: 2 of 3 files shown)


flora131:
# Implementation Notes

Task: Implement GitHub issue https://github.com/flora131/atomic/issues/1082 in its own isolated git worktree. Fetch/read the issue details with gh as needed, follow repo instructions (Bun only), use TDD where applicable, and leave the completed changes in the worktree with tests/typecheck run. Use a branch/worktree name specific to issue 1082.

## Running Notes

- Spec read from `/Users/tonystark/Documents/projects/atomic-issue-1082/specs/2026-05-28-implement-github-issue-https-github-com-flora131-atomic-issues-1082-in-its-own-i.md` before delegation.
- Preflight found a Bun/TypeScript monorepo with documented Bun-only setup. `bun install --frozen-lockfile` was run successfully before implementation.
- Preflight found the worktree on detached HEAD; implementation created/switched to branch `fix/issue-1082-new-workflow-warning`.
- Issue #1082 summary from `gh`: warn before `/new` kills running workflows and clears workflow history; only active/running workflows should trigger friction.
- TDD was followed by adding failing extension tests before implementing the `session_before_switch` handler.
- Implementation decision: the prompt keys on `endedAt === undefined`, matching existing `killAllRuns()` semantics, so pending/running/paused non-ended runs are treated as in-flight while terminal runs with `endedAt` do not prompt.
- Implementation decision: the pre-switch handler is observational only and does not kill runs or clear the workflow store; existing `session_start` cleanup remains responsible after confirmation.
- Tradeoff: if `ctx.ui.confirm` is unavailable, the handler preserves existing non-interactive/headless behavior and allows `/new` to proceed rather than introducing an unanswerable prompt.
- UX decision: declining the confirmation returns `{ cancel: true }` and emits an optional info notification that the new session was cancelled and workflows remain active.
- Changelog updated under `packages/workflows/CHANGELOG.md` Unreleased/Fixed with issue #1082 reference.
- Validation passed: `bun test test/unit/extension.test.ts` (6 tests), `bun test test/unit/background-status-kill.test.ts` (17 tests), and `bun run typecheck`.
- Validation cleanup removed generated subagent report artifacts from the worktree. It also removed the untracked spec file copy from the worktree; this did not affect tracked source changes.
- Final tracked changes are limited to `packages/workflows/src/extension/index.ts`, `test/unit/extension.test.ts`, and `packages/workflows/CHANGELOG.md`.
- No blockers remain.

Issue #1088, opened 2026-05-28
Refactor subagent foreground guard tests away from ESM namespace spyOn


flora131:
Follow-up from PR #1081 review: test/unit/subagents-foreground-guard-propagation.test.ts currently uses spyOn against ESM namespace imports. Bun supports this today, but it may be fragile if module bindings become read-only. Consider adding dependency-injection seams to createSubagentExecutor so foreground guard propagation can be tested without monkey-patching ESM exports.

PR #1091, opened 2026-05-28
fix(subagents): inject executor runtime dependencies

flora131:
## Summary

Introduces a \`SubagentExecutorRuntimeDeps\` dependency-injection seam into the foreground subagent executor so that \`runSync\`, \`executeAsyncChain\`, \`executeAsyncSingle\`, \`isAsyncAvailable\`, and \`formatAsyncStartedMessage\` can be overridden per-executor instance. Fixes unreliable \`spyOn\` usage against ESM module namespaces in tests; production defaults are unchanged.

### Core DI seam (\`subagent-executor.ts\`)
- Exports \`SubagentExecutorRuntimeDeps\` interface with all five overridable functions
- Adds internal \`ResolvedExecutorDeps\` type that narrows the optional \`runtime\` field to fully-resolved after construction
- All execution paths (\`runAsyncPath\`, \`runChainPath\`, \`runParallelPath\`, \`runSinglePath\`, \`resumeAsyncRun\`) route through \`deps.runtime.*\` instead of module-level imports
- Tightens \`SubagentParamsLike\`: \`action\` constrained to \`SUBAGENT_ACTIONS\`, \`agentScope\` typed as \`string\`; adds missing \`chainName\` and \`config\` fields
- Fixes \`inspectSubagentStatus\` call to pass a narrowed params object instead of the full \`SubagentParamsLike\`

### Chain execution (\`chain-execution.ts\`)
- Extracts \`ChainForegroundControl\` type (was duplicated inline across two interfaces)
- Adds \`RunSyncDependency\` type alias
- Threads injected \`runSync\` through \`ChainExecutionParams\` and \`ParallelChainRunInput\` so sequential and parallel chain children use the overridden function

### Type consolidation (\`types.ts\`)
- Adds \`SubagentToolResult = AgentToolResult<Details> & { isError?: boolean }\` alias — consolidates the result type used throughout and removes scattered re-imports of \`AgentToolResult\` from \`pi-agent-core\`

### Supporting fixes
- **\`agent-management.ts\`**: Migrates all handler return types to \`SubagentToolResult\`; adds \`MutableDefinition\` narrowing type guard for \`resolveTarget\`
- **\`run-status.ts\`**: Migrates \`inspectSubagentStatus\` return type to \`SubagentToolResult\`
- **\`settings.ts\`**: Adds \`cwd?: string\` to \`ParallelStep\` interface
- **\`run-history.ts\`**: Replaces \`Boolean(entry)\` with \`entry !== undefined\` for correct TypeScript narrowing in the filter predicate
- **\`chain-clarify.ts\`**: Fixes key casing \`pageup\`/\`pagedown\` → \`pageUp\`/\`pageDown\` to match upstream key-matching convention

### Tests (\`subagents-foreground-guard-propagation.test.ts\`)
- Replaces all \`spyOn\` calls against ESM module namespaces with typed mock functions passed via \`runtime\`
- Removes the \`afterAll\` spy-restore block (no longer needed)
- Derives test types directly from exported signatures (\`ExecutorForTest\`, \`ExecutorDepsForTest\`, etc.)
- Tightens mock signatures from rest-arg destructuring to explicit typed parameters
- Adds \`assertNoErrorFlag\` helper and \`makeUiContext\`/\`makeModelRegistry\` helpers for typed test contexts

## Why

\`spyOn\` against ESM namespace exports is unreliable — module bindings are live and not always reassignable. Dependency injection at the executor level is the idiomatic fix: tests pass fakes through the public API instead of mutating module state.

## Validation

\`\`\`sh
AGENT=1 bun test test/unit/subagents-foreground-guard-propagation.test.ts
bun run lint
bun run test:unit
\`\`\`

Refs #1088

### DIFF
--- a/packages/subagents/CHANGELOG.md
+++ b/packages/subagents/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
+- Hardened workflow-stage subagent guard propagation tests with an internal executor runtime DI seam ([#1088](https://github.com/flora131/atomic/issues/1088)).
 - Capped subagent fanout spawned from workflow stages to a single child level with a workflow-specific nested-subagent error.
 
 ## [0.24.3] - 2026-05-14
--- a/packages/subagents/src/agents/agent-management.ts
+++ b/packages/subagents/src/agents/agent-management.ts
@@ -1,6 +1,5 @@
 import * as fs from "node:fs";
 import * as path from "node:path";
-import type { AgentToolResult } from "@earendil-works/pi-agent-core";
 import { CONFIG_DIR_NAME } from "@bastani/atomic";
 import type { ExtensionContext } from "@bastani/atomic";
 import {
@@ -20,7 +19,7 @@ import {
 import { serializeAgent } from "./agent-serializer.ts";
 import { serializeChain } from "./chain-serializer.ts";
 import { discoverAvailableSkills } from "./skills.ts";
-import type { Details } from "../shared/types.ts";
+import type { SubagentToolResult } from "../shared/types.ts";
 
 type ManagementAction = "list" | "get" | "create" | "update" | "delete";
 type ManagementScope = "user" | "project";
@@ -34,7 +33,7 @@ interface ManagementParams {
 	config?: unknown;
 }
 
-function result(text: string, isError = false): AgentToolResult<Details> {
+function result(text: string, isError = false): SubagentToolResult {
 	return { content: [{ type: "text", text }], isError, details: { mode: "management", results: [] } };
 }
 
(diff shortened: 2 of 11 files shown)


flora131:
# Implementation Notes

Task: Implement GitHub issue https://github.com/flora131/atomic/issues/1088 in this repository. Follow project instructions in AGENTS.md, use Bun for commands, use TDD where applicable, update tests/docs/changelog as appropriate, and leave a concise implementation summary. Use an isolated git worktree under ../atomic-issue-1088 for all changes.

## Running Notes

- Preflight found this is a Bun monorepo workspace. `node_modules/` was initially absent; delegated setup ran `bun install` successfully. No generated artifacts/submodules were required.
- Implemented issue #1088 with an internal executor runtime DI seam for `runSync`, `executeAsyncChain`, `executeAsyncSingle`, `isAsyncAvailable`, and `formatAsyncStartedMessage`; production defaults remain unchanged when no override is supplied.
- Refactored `test/unit/subagents-foreground-guard-propagation.test.ts` to remove `spyOn` and ESM namespace module patching. The test injects fakes through `createSubagentExecutor({ runtime: ... })`.
- Spec deviation/tradeoff: the test still uses a top-level dynamic import through a minimal module interface rather than a static source import. A static import pulled excluded `packages/subagents` source into the root `tsconfig` typecheck graph and surfaced package-local type errors; dynamic import no longer has module-load-order coupling because no spies/mocks are installed before import.
- Added minimal `runSync` dependency propagation through `executeChain()` / `chain-execution.ts` so sequential and parallel chain foreground children can use injected execution without module mocks.
- Added `cwd?: string` to `ParallelStep` settings typing and a local `ChainForegroundControl` type to satisfy typecheck for existing chain fields used by the code once the test import/type surface changed.
- Added a topmost `packages/subagents/CHANGELOG.md` `[Unreleased]` / `Fixed` entry for #1088.
- Removed an unrelated untracked `progress.md` artifact created by subagent progress tracking. Left the untracked spec file untouched.
- Validation passed after fixes: `AGENT=1 bun test test/unit/subagents-foreground-guard-propagation.test.ts`, `AGENT=1 bun test test/unit/subagents-depth-guard.test.ts test/unit/subagents-async-config.test.ts`, `bun run typecheck`, and `git diff --check`.

PR #1093, opened 2026-05-28
feat(cursor-provider): add experimental bundled Cursor AI provider

### LINKED ISSUE #1031: Add Cursor OAuth provider extension
## Problem

Atomic/pi does not currently offer Cursor as an OAuth-backed provider. Contributors who want to use Cursor-hosted authentication or Cursor-specific model access need to configure something manually, and there is no first-party provider extension that demonstrates the integration path.

## Requested Capability

Add support for Cursor as an OAuth provider, most likely as a pi extension, including:

- A Cursor provider extension that follows the existing custom provider extension conventions.
- OAuth login/refresh handling wired into Atomic/pi auth storage.
- Provider metadata so Cursor appears with the expected display name in provider selection flows.
- Documentation for installing/enabling the extension and signing in with Cursor.
- Tests covering auth storage/provider registration behavior where practical.

## Expected Outcome

Users should be able to install or enable the Cursor provider extension, authenticate through Cursor OAuth, and use Cursor as a provider in Atomic/pi without custom local configuration.

## Implementation Notes

Relevant areas to inspect:

- `packages/coding-agent/docs/custom-provider.md`
- `packages/coding-agent/examples/extensions/custom-provider-*`
- `packages/coding-agent/src/core/auth-storage.ts`
- `packages/coding-agent/src/core/provider-display-names.ts`
- `packages/coding-agent/src/modes/interactive/components/oauth-selector.ts`

This should start with discovery of Cursor's current OAuth/provider API surface, then keep the extension small and focused on the Cursor integration.

flora131:
## Summary

Adds `@bastani/cursor-provider` — an experimental, unofficial bundled Atomic extension that lets users authenticate with Cursor, discover available models, and route chat requests through a localhost OpenAI-compatible proxy backed by Cursor's private HTTP/2 + protobuf API.

### New `packages/cursor-provider` package
- **OAuth authentication** (`auth.ts`) — Cursor OAuth login and token refresh flow
- **Model discovery** (`models.ts`, `model-mapping.ts`) — calls Cursor's private `AgentService/GetUsableModels` gRPC endpoint; maps discovered models to pi's `ProviderConfig` format
- **Localhost OpenAI-compatible proxy** (`proxy.ts`, `cursor-bridge.ts`, `h2-bridge.mjs`) — per-process bearer-secret-protected proxy bound to `127.0.0.1`; translates Cursor MCP tool-call events into OpenAI `delta.tool_calls` and strips Cursor-native filesystem/shell tools
- **Protobuf definitions** (`proto/agent_pb.ts`) — generated types for Cursor's private agent protocol
- **Startup credential hydration** — rehydrates stored OAuth credentials on `session_start` so saved accounts can rediscover models without re-running `/login cursor`
- **Debug logging** (`debug.ts`) — opt-in via `ATOMIC_CURSOR_PROVIDER_DEBUG=1`; redacts token-like values

### Wiring into `@bastani/atomic`
- `builtin-packages.ts` — registers `cursor-provider` as a built-in workspace package (bundled at build time alongside `workflows`, `subagents`, `mcp`, `web-access`, `intercom`)
- `scripts/copy-builtin-packages.ts` — copies all cursor-provider files including `h2-bridge.mjs` and `proto/` into the dist bundle
- `package.json` (root + coding-agent) — adds workspace dependency and test script

### Direct `/login <provider>` dispatch (`interactive-mode.ts`)
- `/login cursor` now jumps directly to the Cursor OAuth flow without opening the provider selector
- New `loginProviderByIdOrName` method resolves a provider by id or display name and dispatches to the appropriate login dialog (OAuth, API key, or Bedrock)

### `before_provider_request` provider metadata (`extensions/types.ts`, `runner.ts`, `sdk.ts`)
- `BeforeProviderRequestEvent` gains optional `provider` and `modelId` fields
- `emitBeforeProviderRequest` accepts a `metadata` object forwarded into the event
- Used by the cursor provider's `before_provider_request` handler to inject `pi_session_id` into request payloads

### CI & tests
- `test.yml` — adds `bun run test:cursor-provider` step and smoke checks for cursor-provider built artifacts
- `publish.yml` — adds cursor-provider to both the shell and Node.js smoke-check gates; adds `cursor-provider` to the expected-builtins assertion
- Unit and integration tests for auth, model discovery, proxy, cursor-bridge, direct login, and built-in packaging

### Docs
- `packages/coding-agent/docs/providers.md` — documents Cursor setup, caveats, and runtime behavior
- `packages/cursor-provider/README.md` — package-level README with usage and runtime notes

## Breaking Changes

None. Existing provider flows and extension loading are unchanged. Users who want to opt out of the Cursor extension can pass `--no-extensions`.

## Notes

- Uses Cursor private APIs; not affiliated with or endorsed by Cursor. May break without notice.
- Requires a `node` executable at runtime for the HTTP/2 child bridge (`h2-bridge.mjs`).
- `ATOMIC_CURSOR_PROVIDER_PORT=<port>` can override the ephemeral proxy port for local diagnostics.

Closes #1031


flora131:
- Preflight only; no repository changes made.
- Did not run `bun install` because lockfile, `node_modules/`, and `packages/coding-agent/dist/` were already present, indicating the checkout is initialized.
- If future validation needs a clean-state check, prefer `bun install --frozen-lockfile` per CI/docs.

# Iteration 9 implementation notes

- Preflight subagent confirmed the checkout is already initialized for the Bun 1.3.14 + TypeScript workspace: `bun.lock`, `node_modules/`, and `packages/coding-agent/dist/` are present. No setup command was needed and no initialization blockers were found.
- Implemented successful tree-navigation Cursor state invalidation by adding a provider-side `clearSession(sessionId)` bridge type guard and wiring `session_tree` to clear only the current Atomic session id. `session_before_tree` remains non-mutating because it is cancellable and now logs a pending boundary instead of claiming state was cleared.
- Added `NodeH2CursorBridge.clearSession(sessionId)` to delete one in-memory `StoredSession`; `close()` continues to clear all sessions, active child bridges, and timers. The reset intentionally drops checkpoint/blob state while preserving the existing deterministic conversation-id strategy for future fresh sessions.
- Updated the parent H2 bridge wrapper to pipe child stderr, cap it at 8 KiB, sanitize control characters, redact bearer/current access-token text, and surface sanitized stderr as the error for nonzero child exits.
- Updated `h2-bridge.mjs` to inspect HTTP/2 `:status`; non-2xx or missing/invalid statuses suppress stdout payload forwarding, emit a sanitized path/status diagnostic only, and exit nonzero. Response bodies, request bodies, headers, and tokens are intentionally not included.
- Added tests for provider `session_tree` clearing, no clear on `session_before_tree`, compatibility with bridge stubs lacking `clearSession`, `clearSession` dropping stale checkpoint state, unary/streaming nonzero child stderr propagation, and direct local HTTP/2 non-2xx status handling.
- Subagent output report files created in the repo root during orchestration were removed after validation to avoid leaving extra artifacts.

Validation:

- `AGENT=1 bun run test:cursor-provider` — passed: 57 pass, 0 fail, 192 expect calls across 6 files.
- `AGENT=1 bun run typecheck` — passed (`tsc --noEmit`).
- `git status --short --untracked-files=all` — still shows broad modified/untracked files from the multi-iteration Cursor provider work; no new untracked `research/` scratch artifacts were detected by the validator.

Blockers: none.

# Iteration 10 implementation notes

- Preflight was delegated and corroborated by analysis: the checkout was already initialized for the Bun/TypeScript workspace (`bun.lock`, `node_modules/`, `packages/coding-agent/dist/`, and copied builtin cursor-provider artifacts present). No setup command was needed.
- Replaced remaining provider-runtime `Bun.env` usage with `process.env` and replaced `Bun.serve()` in the Cursor provider proxy with a `node:http` loopback server bound to `127.0.0.1`, preserving the existing `startCursorProxy()` API and `handleCursorProxyRequest(Request, deps)` seam.
- The new HTTP adapter buffers incoming non-GET/HEAD request bodies for the local JSON proxy path, while preserving streaming Fetch `Response.body` output for SSE responses. Adapter errors return a sanitized JSON 500 without request bodies or tokens.
- Startup hydration for expired Cursor OAuth credentials now triggers auth-storage refresh, re-reads refreshed credentials, and applies/discovers/registers models through the same path as login/refresh. Token-fingerprint checks avoid duplicate model discovery when the same process already applied the refreshed token via the OAuth refresh hook.
- Added a fallback for cases where storage readback is unavailable/stale but the OAuth `getApiKey()` side effect populated the current access token; model discovery is run for that token.
- Updated Cursor provider package metadata to include Node `>=20.6.0` alongside Bun `>=1.3.14`, and updated README runtime/env var wording for Node-loaded published Atomic installs, the Node HTTP/2 bridge executable requirement, and `ATOMIC_CURSOR_PROVIDER_PORT`.
- Added/updated regression tests for Node/no-Bun import via a Node+jiti smoke path, real `startCursorProxy()` loopback behavior, cross-process expired refresh hydration registration, and avoiding double model discovery on same-process refresh.
- The first implementer subagent invocation failed without details; a follow-up TypeScript-focused implementation subagent completed the work successfully.

Validation:

- `AGENT=1 bun run test:cursor-provider` — passed: 60 pass, 0 fail, 202 expect calls across 6 files.
- `AGENT=1 bun run typecheck` — passed (`tsc --noEmit`).
- `git status --short --untracked-files=all` — still shows broad modified/untracked files from the multi-iteration Cursor provider work, including the cursor-provider package files and earlier core/test wiring.
- `issues.md` does not exist; no unresolved validation issue file was needed.

Blockers: none.

PR #1131, opened 2026-05-30
feat(coding-agent): clarify requirements then implement workflows directly by default

flora131:
## Summary

Improves workflow-creation guidance in the system prompt so agents gather requirements first, then implement workflows directly — reserving the \`/goal\` workflow only for cases where the user explicitly requests that long-running, reviewer-gated process.

## Changes

- **System prompt** (`packages/coding-agent/src/core/system-prompt.ts`):
  - When a user asks to create a workflow, the agent must ask detailed clarifying questions (purpose, inputs, stages, handoffs, validation, success criteria) and then implement the workflow directly by reading docs/examples.
  - Replaces the previous instruction that defaulted to the \`goal\` workflow for all workflow creation.
  - The \`goal\` workflow is reserved for explicit user requests, or when the agent explains the tradeoff and the user opts in. When \`goal\` *is* used, the objective must be tightly scoped with concrete done criteria and validation steps, and the run must be monitored via workflow status/connect controls rather than manual polling loops.
- **Docs** (`packages/coding-agent/docs/workflows.md`): Adds a note clarifying that Atomic does not use \`/goal\` by default for first-time workflow creation.
- **Tests** (`packages/coding-agent/test/system-prompt.test.ts`): Updates assertions to verify the new guidance strings are present in the generated system prompt when the \`workflow\` tool is available.
- **Changelog** (`packages/coding-agent/CHANGELOG.md`): Documents the guidance change under \`### Changed\` in \`[Unreleased]\`.

## Why

Without explicit guidance, agents defaulted to invoking the \`goal\` workflow for any workflow-creation request. This is slow, token-heavy, and unnecessary for straightforward cases. Routing through clarifying questions first and implementing directly produces faster, more accurate results. The \`goal\` workflow remains available for complex, reviewer-gated work when the user deliberately chooses it.

## Notes

- No breaking changes — the instruction is additive/corrective within the existing \`workflowGuidance\` conditional block.
- The instruction is omitted (along with all other workflow guidance) when the \`workflow\` tool is explicitly excluded.
- `AGENT=1 bun test packages/coding-agent/test/system-prompt.test.ts` passes.


lavaman131:
Created follow-up issue for the system prompt behavior discussed here:

https://github.com/flora131/atomic/issues/1138

flora131:
Observed while validating this PR:

There was an issue where the model did not want to wait for the `/goal` workflow to complete. Instead, it paused the goal workflow and attempted to write the workflow itself. Once the user interrupted the chat and explicitly asked it to wait, it properly unpaused the goal workflow. The chat then started a polling loop along the lines of `sleep 120`, followed by checking the workflow status to decide whether to continue waiting or stop.

While this workaround was running, we hit a Codex max-context-exceeded event, so the model had to auto-compact while `/goal` was still running. That suggests that if we need this behavior to work reliably, we likely need a more robust or different implementation.

It also feels like `/goal` takes the wrong amount of time for writing workflows, which is something else to consider. I do not think this should be default behavior for that reason. The goal workflow was still running with the prompt:

> Create a reusable Atomic workflow that reviews a pull request, runs focused validation, writes a runbook, and summarizes remaining manual checks.

After 30 minutes, the workflow still had not completed. We eventually stopped because `/goal` was still running after 40+ minutes. Perhaps using `/goal` with additional constraints could make sense, but a first-time user probably does not want to wait this long or spend this many tokens each time workflows are written.

PR #1144, opened 2026-05-30
fix(workflows): keep completed HIL prompt archives scrollable

### LINKED ISSUE #1140: Completed human-in-the-loop workflow stages are not scrollable when re-attached
## Summary

After a workflow stage has completed, re-attaching to a completed human-in-the-loop stage no longer allows the stage/context UI to scroll through the full output.

## What happened

A workflow included a human-in-the-loop stage with a large amount of output to review. When the stage was first presented, the UI allowed scrolling through the output so the full review context was visible.

After the human-in-the-loop stage finished, going back and attaching to that same completed stage opened the stage/context UI again, but scrolling no longer worked. As a result, only part of the prior output was visible and important review context could not be reached.

## Expected behavior

Completed human-in-the-loop stages should preserve the same scrollable review experience when re-attached as they had when first shown. A user should be able to scroll through all retained output/context for the stage, even after the stage has completed.

## Actual behavior

When attaching to the completed human-in-the-loop stage, the context UI is not scrollable, preventing access to all of the output that was available during the original review.

## Impact

This makes it difficult or impossible to inspect prior human review context after a workflow progresses, especially for long outputs that do not fit on screen.

## Notes

- Area: workflows TUI / stage attachment UI
- Specific scenario: completed human-in-the-loop stage
- Regression/contrast: scrolling worked when the human-in-the-loop stage was initially active, but not after re-attaching once it had finished

lavaman131:
## Summary

Fixes completed human-in-the-loop (HIL) prompt archives so that reattached read-only prompt nodes remain scrollable via keyboard and mouse-wheel input after a stage completes. Previously, scroll events were swallowed by the read-only input guard before reaching the archived prompt card, making long prompts and their responses unreachable.

### `packages/workflows` — HIL prompt archive scrolling

- **Route scroll input to read-only prompt archives**: Keyboard scroll keys (`pageDown`, `home`, `end`, arrow keys) and mouse-wheel events (`\x1b[<65;...M`) are now dispatched to `_handlePromptScrollInput` before the general read-only input guard fires, so completed prompt cards respond to navigation.
- **Extract `_resetPromptScroll()` helper**: Consolidated `promptScrollOffset = 0` / `promptMaxScroll = 0` resets into a single private method called from `_syncPromptState`, `resolvePrompt`, and the new-prompt branch — preventing stale offsets from carrying over to reattached archives.
- **Fix `_syncPromptState` change detection**: Broadened the "had live prompt" check to also cover `promptEditor !== null` and `promptEditorPromptId !== null`, ensuring clearing editor state always triggers a re-render even when `promptState` was already null.
- **Refactor `_isReadOnlyArchive` call site**: Moved `readOnlyArchive` assignment to after `_syncPromptState` and introduced `readOnlyPromptArchive` (requires `stage.promptFootprint`) to gate prompt-scroll routing only on stages that actually have an archived prompt card.
- **Regression tests**: Added two unit tests in `stage-chat-view.test.ts` — one for keyboard (`pageDown`, `home`, `end`) and one for mouse-wheel — verifying top/bottom content visibility before and after scroll in a completed stage.
- **Changelog**: Added `[Unreleased] ### Fixed` entry referencing issue #1140.

### `packages/subagents` — lazy npm root probe

- **Avoid unnecessary `getGlobalNpmRoot()` call**: Refactored `configuredPiIntercomPackageDir` to lazily resolve the global npm root only when processing a `"user"`-scoped settings file. Project-scoped entries no longer trigger the npm probe, eliminating a redundant subprocess call on every settings scan.

## Closes

Closes #1140


lavaman131:
# Implementation Notes

Task: Implement GitHub issue https://github.com/flora131/atomic/issues/1140 in this repository using a separate reusable git worktree at ../atomic-issue-1140. Inspect the issue details and relevant code, implement the requested changes, add/update tests and docs where appropriate, run relevant validation with Bun commands only, then commit and push the branch and create/update a PR if appropriate. Follow the repository guidance in AGENTS.md. Do not publish, create a release, or create a tag.

## Running Notes

- Preflight confirmed the checkout is initialized: Bun workspace TypeScript monorepo, `bun.lock` and `node_modules` present, `bun --version` is 1.3.14, and `bun pm ls --depth 0` succeeded. No setup command was needed.
- Started from detached `HEAD` at `f81ab4c9`; implementation subagent created/switched to branch `fix/issue-1140-completed-hil-scroll` before editing.
- Implementation stayed localized to `StageChatView`, `test/unit/stage-chat-view.test.ts`, and `packages/workflows/CHANGELOG.md` as planned.
- Decision: reused existing `promptScrollOffset`/`promptMaxScroll` and `_handlePromptScrollInput()` for completed prompt archives instead of adding separate read-only archive scroll state. This matches the spec’s minimal localized design.
- Decision: kept initial re-attach viewport behavior unchanged; the fix restores scrollability after attach, not persisted scroll position across attach sessions.
- Preserved live prompt editor PageUp/PageDown handling, mounted custom UI routing, and generic transcript scrolling for non-prompt read-only archives.
- Added regression coverage for keyboard and mouse-wheel scrolling of completed prompt archives after re-attach.
- Validation passed: `bun test test/unit/stage-chat-view.test.ts`, `bun test test/unit/workflow-attach-pane.test.ts`, `bun run typecheck`, and `bun run test:unit` (`1756 pass`). Commit hooks also ran and passed, including `bun run lint` and `bun run test:unit`.
- Committed tracked changes only; left the untracked spec file uncommitted as requested.
- Commit pushed: `4197b2f67cdef747c55b2cbc86ae05050c908d4a` (`fix(workflows): keep completed HIL prompt archives scrollable`). PR created: https://github.com/flora131/atomic/pull/1144.
- No blockers. No publish, release, or tag actions were performed.

flora131:
The scrolling works properly — this fix works.

Here is the PR 1144 workflow code so it can be reused for future testing:

```ts
import { defineWorkflow } from "@bastani/workflows";

const DEFAULT_LINE_COUNT = 72;

type PromptKind = "input" | "confirm" | "select" | "editor";

function normalizeLineCount(value: unknown): number {
  if (typeof value !== "number" || !Number.isFinite(value)) return DEFAULT_LINE_COUNT;
  return Math.max(24, Math.min(180, Math.floor(value)));
}

function longPrompt(kind: PromptKind, lineCount: number): string {
  const upper = kind.toUpperCase();
  const lines = Array.from({ length: lineCount }, (_, index) => {
    const line = index + 1;
    return `${upper} PROMPT LINE ${String(line).padStart(3, "0")}: PR 1144 archived prompt scroll verification filler; keep this line visible only after scrolling.`;
  });
  return [
    `PR1144 ${upper} PROMPT TOP MARKER`,
    `This ${kind} prompt intentionally exceeds the stage-chat viewport. Scroll while it is active, answer it, then reattach to the completed prompt node and verify the archive still scrolls.`,
    "",
    ...lines,
    "",
    `PR1144 ${upper} PROMPT BOTTOM MARKER`,
  ].join("\n");
}

function selectOptions(lineCount: number): readonly string[] {
  return [
    "alpha-short-choice",
    [
      "bravo-long-choice: PR1144 SELECT CHOICE TOP MARKER",
      ...Array.from({ length: Math.max(8, Math.floor(lineCount / 3)) }, (_, index) => `select choice detail line ${index + 1}`),
      "bravo-long-choice: PR1144 SELECT CHOICE BOTTOM MARKER",
    ].join(" | "),
    "charlie-final-choice",
  ] as const;
}

function editorInitial(lineCount: number): string {
  return [
    "PR1144 EDITOR INITIAL TOP MARKER",
    "Edit this document, optionally add a response marker, then submit it. The completed archive should show both the long initial value and the saved response after scrolling.",
    "",
    ...Array.from({ length: lineCount }, (_, index) => {
      const line = index + 1;
      return JSON.stringify({
        line,
        check: "completed HIL editor archive remains scrollable after reattach",
        action: "scroll active editor, submit, reattach, scroll archived editor prompt",
      });
    }),
    "",
    "PR1144 EDITOR INITIAL BOTTOM MARKER",
  ].join("\n");
}

export default defineWorkflow("pr1144-hil-archive-scroll")
  .description("Manual PR 1144 workflow: creates long ctx.ui input/confirm/select/editor prompt nodes so completed HIL archives can be reattached and scrolled with keyboard and mouse wheel.")
  .input("line_count", {
    type: "number",
    default: DEFAULT_LINE_COUNT,
    description: "Number of filler lines generated for each long HIL prompt. Defaults to 72.",
  })
  .run(async (ctx) => {
    const lineCount = normalizeLineCount(ctx.inputs.line_count);

    const inputResponse = await ctx.ui.input(longPrompt("input", lineCount));
    const confirmed = await ctx.ui.confirm(longPrompt("confirm", lineCount));
    const selected = await ctx.ui.select(longPrompt("select", lineCount), selectOptions(lineCount));
    const edited = await ctx.ui.editor(editorInitial(lineCount));

    return {
      lineCount,
      inputResponseLength: inputResponse.length,
      confirmed,
      selected,
      editedLength: edited.length,
      manualVerification: [
        "For each completed HIL prompt node, reattach from the graph and verify PageDown/Home/End plus mouse wheel reveal TOP and BOTTOM markers.",
        "Confirm Ctrl+D still returns to the graph and Escape still closes the read-only archive.",
      ],
    };
  })
  .compile();
```

PR #1149, opened 2026-05-31
fix(workflows): guard HIL prompt input against accidental resolution

### LINKED ISSUE #1148: Workflow select prompt can auto-advance while navigating options
## Summary

While manually verifying PR 1144's HIL/archive-scroll workflow, the `ctx.ui.select(...)` prompt appears to sometimes resolve and advance to the next `ctx.ui.editor(...)` prompt even though the user has **not pressed Enter** to select an option.

The observed behavior was:

1. Run the PR1144 test workflow.
2. Answer the initial `input` and `confirm` prompts.
3. Reach the `select` stage whose options include `alpha-short-choice`.
4. Use only the Up/Down arrow keys to move between select options.
5. Before pressing Enter, the select stage greys out as completed/resolved.
6. The workflow automatically proceeds to the editor stage.

This feels like the select prompt is receiving an unintended submit/default/cancel event, or that input/focus routing is resolving the prompt while the user is only navigating.

## Expected behavior

For a fresh run, Up/Down arrow keys in the select prompt should only move the highlighted option. The select prompt should remain pending until the user explicitly submits with Enter or intentionally cancels/skips.

The workflow should not advance from `select` to `editor` while the user is only navigating options with arrow keys.

## Actual behavior

The select stage greys out and the workflow advances to the editor stage before Enter is pressed.

The selected-looking option was the first option, `alpha-short-choice`, which may indicate either:

- the default highlighted option was submitted unexpectedly,
- a cancel/default path resolved the select prompt,
- a buffered/repeated key event was interpreted as submit/cancel,
- or the stage/chat focus routing caused the prompt to resolve unintentionally.

## Why this appears unexpected from the workflow logic

The PR1144 workflow is linear:

```ts
const inputResponse = await ctx.ui.input(longPrompt("input", lineCount));
const confirmed = await ctx.ui.confirm(longPrompt("confirm", lineCount));
const selected = await ctx.ui.select(longPrompt("select", lineCount), selectOptions(lineCount));
const edited = await ctx.ui.editor(editorInitial(lineCount));
```

So it is expected that `editor` starts immediately **after** `select` resolves. The questionable part is that `select` appears to resolve before explicit submission.

From code inspection, this does not look like intended behavior:

- `packages/workflows/src/tui/prompt-card.ts` initializes select prompts with `selectedIndex: 0`, so the first option is pre-highlighted by design.
- Arrow key navigation should update selection but not submit.
- `Enter` should submit.
- `Esc`/`Ctrl+C` can resolve with defaults, where select defaults to the first choice.
- `packages/workflows/src/runs/foreground/executor.ts` records a pending prompt and waits for it; there does not appear to be an intentional timer that auto-answers fresh prompts.

## Reproduction workflow

This is the exact test workflow used for the PR1144 verification. Save it as `.atomic/workflows/pr1144-hil-archive-scroll.ts` and run it from Atomic's workflow UI.

```ts
import { defineWorkflow } from "@bastani/workflows";

const DEFAULT_LINE_COUNT = 72;

type PromptKind = "input" | "confirm" | "select" | "editor";

function normalizeLineCount(value: unknown): number {
  if (typeof value !== "number" || !Number.isFinite(value)) return DEFAULT_LINE_COUNT;
  return Math.max(24, Math.min(180, Math.floor(value)));
}

function longPrompt(kind: PromptKind, lineCount: number): string {
  const upper = kind.toUpperCase();
  const lines = Array.from({ length: lineCount }, (_, index) => {
    const line = index + 1;
    return `${upper} PROMPT LINE ${String(line).padStart(3, "0")}: PR 1144 archived prompt scroll verification filler; keep this line visible only after scrolling.`;
  });
  return [
    `PR1144 ${upper} PROMPT TOP MARKER`,
    `This ${kind} prompt intentionally exceeds the stage-chat viewport. Scroll while it is active, answer it, then reattach to the completed prompt node and verify the archive still scrolls.`,
    "",
    ...lines,
    "",
    `PR1144 ${upper} PROMPT BOTTOM MARKER`,
  ].join("\n");
}

function selectOptions(lineCount: number): readonly string[] {
  return [
    "alpha-short-choice",
    [
      "bravo-long-choice: PR1144 SELECT CHOICE TOP MARKER",
      ...Array.from({ length: Math.max(8, Math.floor(lineCount / 3)) }, (_, index) => `select choice detail line ${index + 1}`),
      "bravo-long-choice: PR1144 SELECT CHOICE BOTTOM MARKER",
    ].join(" | "),
    "charlie-final-choice",
  ] as const;
}

function editorInitial(lineCount: number): string {
  return [
    "PR1144 EDITOR INITIAL TOP MARKER",
    "Edit this document, optionally add a response marker, then submit it. The completed archive should show both the long initial value and the saved response after scrolling.",
    "",
    ...Array.from({ length: lineCount }, (_, index) => {
      const line = index + 1;
      return JSON.stringify({
        line,
        check: "completed HIL editor archive remains scrollable after reattach",
        action: "scroll active editor, submit, reattach, scroll archived editor prompt",
      });
    }),
    "",
    "PR1144 EDITOR INITIAL BOTTOM MARKER",
  ].join("\n");
}

export default defineWorkflow("pr1144-hil-archive-scroll")
  .description("Manual PR 1144 workflow: creates long ctx.ui input/confirm/select/editor prompt nodes so completed HIL archives can be reattached and scrolled with keyboard and mouse wheel.")
  .input("line_count", {
    type: "number",
    default: DEFAULT_LINE_COUNT,
    description: "Number of filler lines generated for each long HIL prompt. Defaults to 72.",
  })
  .run(async (ctx) => {
    const lineCount = normalizeLineCount(ctx.inputs.line_count);

    const inputResponse = await ctx.ui.input(longPrompt("input", lineCount));
    const confirmed = await ctx.ui.confirm(longPrompt("confirm", lineCount));
    const selected = await ctx.ui.select(longPrompt("select", lineCount), selectOptions(lineCount));
    const edited = await ctx.ui.editor(editorInitial(lineCount));

    return {
      lineCount,
      inputResponseLength: inputResponse.length,
      confirmed,
      selected,
      editedLength: edited.length,
      manualVerification: [
        "For each completed HIL prompt node, reattach from the graph and verify PageDown/Home/End plus mouse wheel reveal TOP and BOTTOM markers.",
        "Confirm Ctrl+D still returns to the graph and Escape still closes the read-only archive.",
      ],
    };
  })
  .compile();
```

## Suggested investigation areas

- Whether an extra `Enter`, `Esc`, or `Ctrl+C` event is being delivered after select navigation.
- Whether key repeat/buffered terminal input can be routed to the select prompt as confirm/cancel.
- Whether focus changes between graph/stage-chat/prompt overlay can resolve the active prompt.
- Whether continuation/replay or restored prompt answers are being applied unexpectedly to a fresh-looking run.
- Whether a custom keybinding could bind `tui.select.confirm` to an arrow key or otherwise make arrow navigation submit.

## Notes

If this only happens under continuation/replay, then the UI should make that explicit because it currently looks like the select prompt auto-selected an option. If this happens in a fresh run with only Up/Down arrow keys, it is likely a TUI/input-routing bug.

lavaman131:
## Summary

Guards workflow Human-in-the-Loop (HIL) prompt input so navigation keys, scroll input, and ambiguous Escape-prefix sequences cannot accidentally resolve pending prompts. Also closes a repeated-Enter race condition where the keystroke that submitted a prompt answer would immediately attach the next pending prompt.

Closes #1148

### Input routing fixes

- **Escape handling**: Separates `Escape` from `Ctrl+C` — bare Escape (and its kitty/CSI-u variants) is now consumed as a no-op on prompt cards and host-backed prompt editors via the new `isPromptEscapeInput` export. `Ctrl+C` is the explicit skip/cancel path. Prevents split terminal escape-prefix sequences from accidentally resolving a prompt to its default response.
- **Select prompt movement**: Replaces delegation to pi-tui's `SelectList` with a deterministic `selectMovementDelta` handler covering arrow keys, left/right aliases, and page-up/down via configurable keybindings (`tui.select.pageUp` / `tui.select.pageDown`). Navigation no longer risks silently submitting a selection.
- **Page navigation keybindings**: Adds `selectPageUp` / `selectPageDown` actions to `TUI_ACTION` and default fake-keybinding fixtures.
- **Host editor Escape**: The host-backed prompt editor path (`promptEditor`) now ignores bare Escape the same way the card renderer does, forwarding only non-Escape input to the editor.

### Repeated-Enter suppression

- Introduces `StageChatDetachReason` (`"user"` | `"prompt-resolved"`) so `StageChatView` can distinguish an explicit Ctrl+D detach from an auto-detach after a prompt answer.
- `WorkflowAttachPane` sets a one-shot `suppressNextGraphSubmit` flag when detaching due to prompt resolution via Enter, consuming the first graph-mode Enter so it cannot attach the next pending prompt unintentionally. Ctrl+C skips and non-Enter submissions (confirm `y`/`n`, remapped select confirm) do not consume the next Enter.

### Custom keybinding propagation

- Passes host `piKeybindings` through `WorkflowAttachPane` → `GraphView` so run-level prompt cards honor custom select keybindings, consistent with stage-level cards.

### UX / hints

- Updated all prompt hint strings from `esc/ctrl+c` → `ctrl+c` to accurately reflect that Escape is no longer an active skip trigger.

### Refactoring

- Extracts `matchesAnyKey`, `applyDeleteRange`, and `takeRows` helpers to reduce repetition in `prompt-card.ts` and `stage-chat-view.ts`.
- Replaces `normalizeSelectKeyData` (which translated keys for pi-tui's `SelectList`) with the self-contained `selectMovementDelta` function.

## Files changed

| File | Change |
|---|---|
| `src/tui/prompt-card.ts` | Escape no-op, `selectMovementDelta`, `isPromptEscapeInput` export |
| `src/tui/stage-chat-view.ts` | `StageChatDetachReason`, host editor Escape guard, `takeRows` helper |
| `src/tui/workflow-attach-pane.ts` | `suppressNextGraphSubmit` flag, `piKeybindings` propagation |
| `src/tui/keybindings-adapter.ts` | `selectPageUp` / `selectPageDown` actions added to `TUI_ACTION` |
| `test/support/fake-keybindings.ts` | Default fixtures for new page-nav keybindings |

## Tests

- **`prompt-card.test.ts`**: Verifies Ctrl+C variants cancel while Escape variants return `noop`; custom keybinding navigation and submission; bare Escape does not resolve select or editor prompts.
- **`stage-chat-view.test.ts`**: End-to-end coverage for structured select/editor prompts remaining pending under navigation/scroll/Escape input; host editor consuming Escape without resolving.
- **`workflow-attach-pane.test.ts`**: Regression tests asserting the first graph-mode Enter after an Enter-submitted prompt answer is consumed; Ctrl+C skip, confirm `y`/`n`, and remapped select confirm all leave the next Enter free to attach normally; `piKeybindings` forwarded to run-level graph prompt cards.
- **`overlay-graph.test.ts`**: Verifies remapped select keybindings are honored for run-level graph-mode prompt cards.

## Test results

- Focused 5-file suite: **150 passed**
- `bun run typecheck`: **passed**
- `AGENT=1 bun run test:unit`: **1761 passed**


lavaman131:
# Finalize Report — GitHub Issue #1148 Iteration 3

## Summary

Finalized iteration 3 on branch `fix/1148-hil-prompt-input-routing`. The iteration-3 source/test/changelog changes were committed, pushed to origin, and the existing PR was updated rather than duplicated.

## Repository / artifact handling

- Read `AGENTS.md` and the issue spec at `specs/2026-05-31-implement-github-issue-https-github-com-flora131-atomic-issues-1148-in-this-repo.md`.
- Inspected `git status`, current diffs, and the full branch diff via `git diff origin/main...HEAD`.
- Removed untracked `implementation-report.md` as requested.
- Did not commit workflow artifacts. The spec file was left untracked because no repo convention requiring specs to be committed was found. Existing untracked `analysis-report.md` and `preflight-report.md` were also left uncommitted.
- Did not publish, release, tag, or version-bump.

## Validation

All requested/practical validation was run with Bun-only commands:

- `AGENT=1 bun test test/unit/workflow-attach-pane.test.ts test/unit/stage-chat-view.test.ts test/unit/prompt-card.test.ts test/unit/overlay-graph.test.ts`
  - Result: `207 pass`, `0 fail`.
- `bun run typecheck`
  - Result: passed (`tsc --noEmit`).
- `AGENT=1 bun test test/unit/background-runner-hil.test.ts test/unit/store-pending-prompt.test.ts`
  - Result: `27 pass`, `0 fail`.
- `AGENT=1 bun run test:unit`
  - Result: `1767 pass`, `0 fail`.

Pre-commit/pre-push hooks also ran successfully and included `bun run lint` and `bun run test:unit`.

## Commit

- Commit: `72a0b02f0912aee85beb9b85f48e2dee501571c4`
- Message: `fix(workflows): narrow prompt enter suppression`
- AI attribution included in the commit body: `AI-Assisted-By: OpenAI ChatGPT`

Committed files:

- `packages/workflows/CHANGELOG.md`
- `packages/workflows/src/tui/graph-view.ts`
- `packages/workflows/src/tui/stage-chat-view.ts`
- `packages/workflows/src/tui/workflow-attach-pane.ts`
- `test/unit/workflow-attach-pane.test.ts`

## Branch / push / PR

- Branch: `fix/1148-hil-prompt-input-routing`
- Push: `git push -u origin fix/1148-hil-prompt-input-routing` succeeded; branch tracks `origin/fix/1148-hil-prompt-input-routing`.
- Existing PR updated: https://github.com/flora131/atomic/pull/1149
- PR status: `OPEN`
- PR title: `fix(workflows): guard HIL prompt input against accidental resolution`
- No duplicate PR was created.

## Notable decisions / tradeoffs

- Kept the internal field name `suppressNextGraphSubmit` to minimize churn, but narrowed behavior to raw Enter-origin prompt submissions only.
- Used explicit detach metadata from `StageChatView` to `WorkflowAttachPane` so Ctrl+C skips, `y`/`n` confirms, and remapped select-confirm keys do not consume the next graph-mode Enter.
- For host prompt editors, Enter-origin is tracked with a transient flag around `promptEditor.handleInput(data)`. This preserves repeated-Enter protection for synchronous raw-Enter submits and conservatively avoids suppressing graph Enter for asynchronous/non-Enter submits.
- The PR body was updated to describe iteration 3 behavior and the latest validation results.

## Final state notes

- Tracked working tree is clean after commit/push.
- Untracked workflow/report artifacts remain intentionally uncommitted: `analysis-report.md`, `preflight-report.md`, the issue spec under `specs/`, and the task output report `finalize-report.md` once written.

## Supervisor consolidation

- Preflight found the checkout initialized for this Bun/TypeScript monorepo: Bun `1.3.14`, `bun.lock`, workspace manifests, and `node_modules` were present; no submodules or setup blockers were found. A dry-run `bun install --frozen-lockfile --dry-run` invoked the existing prepare hook but did not require dependency changes.
- Analysis confirmed the iteration-3 fix should remain local to `StageChatView` and `WorkflowAttachPane`; no persisted prompt schema or public workflow API changes were needed.
- Implementation followed the spec's selected approach: prompt detach now carries in-memory metadata, and graph Enter suppression is armed only for raw-Enter prompt submissions.
- Validation completed successfully with targeted prompt/TUI tests, HIL/store tests, full unit tests, typecheck, and hooks. No blockers remain for this iteration.

PR #1158, opened 2026-05-31
fix(tui): extract ReactiveWidget, fix remount flicker and clock freeze

### LINKED ISSUE #1150: Async subagent UI causes TUI flicker while running
## Summary

Async subagent activity can introduce visible TUI flicker. When one or more async subagents are running, the async subagent status/widget can flicker and the main chat box area constantly flickers/repaints.

## Symptoms

- The async subagent widget itself visibly flickers while async runs are active.
- The chat input/chat box area appears to constantly flicker even when the user is not changing it.
- The behavior is similar in character to the whole-screen flicker fixed in #1122, but appears tied to async subagent UI rendering rather than workflow status cards.

## Expected behavior

Async subagent status updates should render stably. TUI regions should not repaint/flicker continuously unless their displayed content actually changes.

## Suspected cause / direction

This likely needs the same methodology used in #1122: avoid render-time nondeterminism in components that are re-rendered frequently.

In #1122, workflow status cards flickered because a `dynamicTextRenderComponent` callback recomputed elapsed time with `Date.now()` on every TUI render, causing scrollback bytes to change on every frame. The fix captured wall-clock time once when the chat entry was created and threaded that stable timestamp through the renderer.

For async subagents, investigate whether the widget/status rendering or chat-surface rendering:

- calls `Date.now()` or computes elapsed/running text during every render;
- changes spinner/progress text in regions that force broader diffs than intended;
- emits unstable output from a `dynamicTextRenderComponent` or equivalent render callback;
- updates off-viewport or layout-affecting content that causes the chat box/input region to redraw.

The likely fix is to capture volatile values at the time the status entry/update is produced, or otherwise isolate live-changing values so they do not cause unrelated TUI regions to repaint.

## Reference

- Related fix: #1122

lavaman131:
## Summary

Introduces a shared `ReactiveWidget` lifecycle abstraction in `@bastani/atomic` that drives mount/unmount/update decisions, coalesced render requests, and timer-based refreshes for TUI widgets. All three packages adopt it, eliminating async widget remount flicker, scrollback spinner clock drift, and startup latency from the npm skill probe.

Closes #1150.

### ReactiveWidget Abstraction (`coding-agent/src/core/extensions/reactive-widget.ts`) *(new)*
- **`installReactiveWidget<TSnapshot, TTheme>`**: Generic controller that owns the full widget lifecycle — subscribes to snapshot changes, decides `mount / unmount / update / none` via `decideReactiveWidgetAction`, coalesces `requestRender` calls through a microtask queue, and schedules timer-based clock refreshes.
- **`decideReactiveWidgetAction`**: Pure helper comparing previous `ReactiveWidgetRenderState` to next preview lines, so the decision logic is independently testable.
- All related types (`ReactiveWidgetComponent`, `ReactiveWidgetUi`, `ReactiveWidgetController`, etc.) exported from `@bastani/atomic`.

### Workflows Store-Widget Installer (`tui/store-widget-installer.ts`)
- Replaced ~90 lines of inline lifecycle logic with a single `installReactiveWidget` call, picking up coalesced renders and proper dispose semantics for free.
- `WidgetFactory`, `WidgetAction`, `WidgetRenderState`, `TimerApi`, `TimerHandle` re-exported as aliases over the shared types.

### Async Widget Lifecycle (`subagents/src/tui/render.ts`)
- **`LiveWidgetComponent`**: Constructor now takes a live `getJobs: () => AsyncJobState[]` getter instead of a captured snapshot, so the mounted component always reads the latest job state without a remount.
- **`renderWidget`**: Tracks `widgetMounted` / `mountedWidgetCtx` state. `setWidget` fires only on mount and unmount; visible-to-visible updates call `requestRender()` instead, eliminating remount-driven flicker.
- **`stopWidgetAnimation`**: Explicitly calls `setWidget(WIDGET_KEY, undefined)` on full teardown when a widget is currently mounted.

### Chatbox Result Row Clock Freeze (`subagents/src/extension/index.ts`, `tui/render.ts`)
- **`renderSubagentResult` / `renderSingleCompact` / `renderMultiCompact` / `resultGlyph` / `runningGlyph`**: Accept `now?: number`; when provided, the spinner uses the captured timestamp so chat scrollback rows are byte-stable across host re-renders.
- **`renderResult`**: Captures a stable `subagentResultSnapshotNow` once per tool-result row via `context.state`.
- **`createSlashResultComponent`**: Captures `lastSnapshotNow` per snapshot version.
- Removed the `syncResultAnimation` polling timer from both `renderResult` and `createSlashResultComponent` — the frozen clock replaces the need for steady re-render scheduling.

### npm Skill Discovery Hardening (`agents/skills.ts`)
- Timeout reduced 5 000 ms → 2 500 ms.
- Stderr suppressed via `stdio: ["ignore", "pipe", "ignore"]`.
- `windowsHide: true` prevents a spurious console window on Windows.
- `clearSkillCache()` now also resets `cachedGlobalNpmRoot` so the probe is retried after cache clear.
- Exported `__setGlobalNpmRootExecSyncForTest` for unit-test injection.

### Tests
- **`reactive-widget.test.ts`** *(new)*: Full lifecycle coverage for `installReactiveWidget` — mount, unmount, update, coalesced renders, clock ticks, subscribe/dispose, and stale-error handling.
- **`subagents-render-stability.test.ts`**: Extended with regressions for visible→visible skip, context-switch unmount/remount, repeated-empty single unmount, and captured `opts.now` byte-stability.
- **`subagents-skills-npm-probe.test.ts`** *(new)*: Verifies failure caching, `clearSkillCache` retry, and expected spawn options (timeout, stdio, `windowsHide`).

## Validation

```sh
bun test test/unit/reactive-widget.test.ts
bun test test/unit/subagents-render-stability.test.ts
bun test test/unit/subagents-skills-npm-probe.test.ts
bun test test/unit/subagents-async-widget-visibility.test.ts
bun test test/unit/store-widget-installer.test.ts
bun test test/unit/chat-surface.test.ts test/unit/run-detail-render.test.ts
bun run typecheck
```

Pre-push hooks ran `bun run lint` and `bun run test:unit`.


lavaman131:
# Implementation Notes

Task: Address GitHub issue #1150 in flora131/atomic.

Title: Async subagent UI causes TUI flicker while running
URL: https://github.com/flora131/atomic/issues/1150
Labels: bug

Work in the provided git worktree. Read the issue and existing TUI/subagent code before editing. Implement the fix with tests where appropriate, use Bun commands only, update changelog/docs if warranted, and run targeted validation. Do not publish or create a PR unless explicitly instructed. Final output should include the root cause, files changed, validation run, and any follow-up needed.

## Running Notes

- Preflight: checkout was a Bun workspace and initially lacked `node_modules`; delegated setup ran `PREK_DISABLE_INSTALL=1 bun install --frozen-lockfile` successfully. No submodules or generated artifacts were required.
- Root cause confirmed by analysis subagent: async `renderWidget()` remounted with `ctx.ui.setWidget(...)` on every non-empty status update, and the interactive host disposes/rebuilds extension widgets on each `setWidget`, causing below-editor/chat-input flicker under frequent async updates.
- Implementation: `LiveWidgetComponent` now reads jobs through a live getter; async widget lifecycle tracks mounted state so hidden→visible mounts once, visible→visible updates `latestWidgetJobs` and requests an in-place render, and visible→hidden unmounts once. The 80ms spinner ticker and `belowEditor` placement were preserved.
- Tradeoff/decision: kept `async-job-tracker.ts` unchanged to minimize scope; `renderWidget()` now owns visible in-place `requestRender()`, while existing tracker render requests remain compatible.
- Teardown decision: `stopWidgetAnimation()` now best-effort clears the mounted widget if a live UI context exists, then resets widget state; cleanup failures are swallowed to avoid shutdown/reload crashes.
- Tests added in `test/unit/subagents-render-stability.test.ts` for no visible→visible remount with updated content from the original mounted component, and single unmount on repeated empty updates.
- Changelog: added an Unreleased Fixed entry in `packages/subagents/CHANGELOG.md` referencing #1150.
- Validation passed: `bun test test/unit/subagents-render-stability.test.ts`; `bun test test/unit/subagents-async-widget-visibility.test.ts`; `bun test test/unit/store-widget-installer.test.ts`; `bun test test/unit/chat-surface.test.ts test/unit/run-detail-render.test.ts`; `bun run typecheck`.
- Worktree status after validation: modified tracked files are `packages/subagents/CHANGELOG.md`, `packages/subagents/src/tui/render.ts`, and `test/unit/subagents-render-stability.test.ts`; the spec file remains untracked as workflow input.

PR #1282, opened 2026-06-07
fix(coding-agent): resume queued work after auto-compaction

### LINKED ISSUE #1280: Agent stops responding after auto-compaction until user prompts again
## Summary

After an agent session auto-compacts, the agent can stop responding instead of continuing the in-progress task. The user has to manually send another prompt to get the agent to continue.

## Expected behavior

When auto-compaction completes, the agent should automatically resume the current turn/task and continue responding without requiring user intervention.

## Actual behavior

The agent appears to stall after auto-compaction. It does not continue the response/task until the user manually prompts it again.

## Impact

- Long-running tasks can hang unexpectedly.
- Users may think the agent is still working when it is actually waiting for input.
- Automated or unattended workflows become unreliable because manual nudging is required.

## Reproduction notes

1. Start a sufficiently long agent task/session that triggers auto-compaction.
2. Wait for auto-compaction to complete.
3. Observe that the agent does not continue the previous task/response.
4. Send a manual follow-up prompt.
5. Observe that the agent resumes only after the manual prompt.

## Possible area to investigate

The continuation/resume signal after auto-compaction may not be re-entering the active turn loop, or the post-compaction state may incorrectly mark the session as waiting for user input.

lavaman131:
## Summary

Fixes two timing races in `AgentSession._runAutoCompaction()` that caused the agent to stall after automatic context compaction. Overflow retry context is now cleaned synchronously before `compaction_end` listeners can start a new run, and the post-compaction continuation probe is deferred to sample live queue state after listener-initiated queue flushing.

Closes #1280

## Changes

- Extracts `_dropTrailingOverflowAssistantErrorIfPresent()` and runs it synchronously after rebuilding compacted context but before emitting `compaction_end`, so the trailing overflow assistant error cannot leak into listener-started runs
- Extracts `_schedulePostAutoCompactionContinuationProbe()` to defer the queue/streaming state sample until after `compaction_end` listeners have had a chance to flush queued messages — replaces the flawed immediate-snapshot approach (rejected in round 1 review)
- Extracts `_resumeAfterAutoCompaction()` as the sole gate for calling `agent.continue()` after compaction
- Extends threshold auto-compaction to resume queued active-turn work if present (previously always waited for manual user input)
- Adds 4 regression tests: async listener queue flushing, streaming suppression of deferred probe, overflow cleanup timing before listener-started streaming, and idle threshold compaction does not resume
- Adds issue #1280 changelog entry and implementation RFC/spec

## Validation

```sh
bun --cwd=packages/coding-agent run test -- test/agent-session-auto-compaction-queue.test.ts test/agent-session-compaction.test.ts
bun run typecheck
git diff --check main
```

Pre-push hooks: `bun run lint`, `bun run test:unit`

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed auto-compaction so queued in-progress work resumes without requiring a manual follow-up prompt ([#1280](https://github.com/bastani-inc/atomic/issues/1280)).
+
 ## [0.8.26-alpha.6] - 2026-06-06
 
 ### Changed
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -2101,7 +2101,7 @@ export class AgentSession {
 	 *
 	 * Two cases:
 	 * 1. Overflow: LLM returned context overflow error, remove error message from agent state, compact, auto-retry
-	 * 2. Threshold: Context over threshold, compact, NO auto-retry (user continues manually)
+	 * 2. Threshold: Context over threshold, compact, resume queued active-turn work if present; otherwise wait for user
 	 *
 	 * @param assistantMessage The assistant message to check
 	 * @param skipAbortedCheck If false, include aborted messages (for pre-prompt check). Default: true
@@ -2184,6 +2184,49 @@ export class AgentSession {
 		if (shouldCompact(contextTokens, contextWindow, settings)) {
 			await this._runAutoCompaction("threshold", false);
 		}
+	}
+
+	/**
+	 * Internal: remove the trailing overflow error from retry context if it is still present.
+	 */
+	private _dropTrailingOverflowAssistantErrorIfPresent(): void {
+		const messages = this.agent.state.messages;
+		const lastMsg = messages[messages.length - 1];
+		if (lastMsg?.role === "assistant" && (lastMsg as AssistantMessage).stopReason === "error") {
+			this.agent.state.messages = messages.slice(0, -1);
+		}
+	}
+
+	/**
+	 * Internal: schedule a live post-event continuation probe after compaction_end listeners can flush queues.
+	 */
+	private _schedulePostAutoCompactionContinuationProbe(
+		reason: "overflow" | "threshold",
+		willRetry: boolean,
+	): void {
+		setTimeout(() => {
+			if (this.isCompacting || this.isStreaming) {
+				return;
+			}
+
+			if (reason === "overflow" && willRetry) {
+				this._resumeAfterAutoCompaction();
+				return;
+			}
+
+			if (!this.agent.hasQueuedMessages()) {
+				return;
+			}
+
+			this._resumeAfterAutoCompaction();
+		}, 100);
+	}
+
+	/**
+	 * Internal: resume generation after successful auto-compaction only when active work remains.
+	 */
+	private _resumeAfterAutoCompaction(): void {
+		this.agent.continue().catch(() => {});
 	}
 
 	/**
@@ -2306,6 +2349,9 @@ export class AgentSession {
 			const newEntries = this.sessionManager.getEntries();
 			const sessionContext = this.sessionManager.buildSessionContext();
 			this.agent.state.messages = sessionContext.messages;
+			if (reason === "overflow" && willRetry) {
+				this._dropTrailingOverflowAssistantErrorIfPresent();
+			}
 
 			// Get the saved compaction entry for the extension event
 			const savedCompactionEntry = newEntries.find((e) => e.type === "compaction" && e.summary === summary) as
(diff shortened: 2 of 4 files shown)


lavaman131:
# Implementation Notes

Task: Implement GitHub issue #1280: https://github.com/bastani-inc/atomic/issues/1280

Title: Agent stops responding after auto-compaction until user prompts again

Objective: Fix the Atomic agent bug where, after an automatic session compaction completes during an in-progress turn/task, the agent automatically resumes/continues that same turn without requiring a manual follow-up prompt from the user.

Issue summary:
- Expected: auto-compaction completes, then the current turn/task continues responding automatically.
- Actual: after auto-compaction the agent appears to stall and only resumes after the user sends another prompt.
- Impact: long-running/unattended tasks can hang unexpectedly.
- Possible area: continuation/resume signal after auto-compaction may not re-enter the active turn loop, or post-compaction state may incorrectly mark the session as waiting for user input.

Scope and constraints:
- Work in a separate git worktree at ../atomic-auto-compaction-resume-1280.
- Use base branch main.
- Use Bun commands only for development/tests (no npm/yarn/pnpm/node except release-only npm publish, which is irrelevant here).
- Focus on the auto-compaction continuation/resume flow; avoid unrelated rewrites.
- Preserve manual compaction and normal waiting-for-user behavior.
- If debugging reveals multiple issues, track them in issues.md during the run and delete it once resolved.

Concrete done criteria:
1. Locate the auto-compaction completion path and active turn/session state handling responsible for resuming generation.
2. Implement a minimal, robust fix so auto-compaction queues or triggers continuation of the interrupted active turn instead of leaving the agent waiting for user input.
3. Add or update tests that fail before the fix and verify automatic continuation after auto-compaction without a user nudge.
4. Ensure existing behavior for manual compaction, explicit user waits, and normal session state transitions remains intact.
5. Update the appropriate package CHANGELOG.md under the [Unreleased] ### Fixed section if the changed package has a changelog entry convention.
6. Validate with targeted tests plus bun run typecheck; run broader bun test/bun run test:unit if practical.
7. Create a pull request when implementation and validation are complete, with a concise summary, test evidence, and a link to issue #1280.

## Running Notes

- Preflight: repository is a Bun monorepo (`packageManager: bun@1.3.14`). Fresh worktree was not initialized, so `bun install` was run successfully; it installed dependencies and `prek` hooks via the root prepare script with no tracked/staged setup changes.
- Discovery: the relevant completion path is `AgentSession._runAutoCompaction()` in `packages/coding-agent/src/core/agent-session.ts`; manual `AgentSession.compact()` is a separate abort-and-wait path and was left unchanged.
- Implementation decision: extracted an explicit private `_resumeAfterAutoCompaction()` gate in `AgentSession` rather than rewriting compaction flow. The gate is invoked only after successful auto-compaction and schedules one deferred `agent.continue()` when `willRetry` or `agent.hasQueuedMessages()` is true.
- Tradeoff: continued relying on `Agent.hasQueuedMessages()` as the authoritative public queue signal instead of reaching into core private queue internals. No optional event metadata such as `willContinue` was added in this iteration to keep the external event surface unchanged.
- Behavior preserved: threshold auto-compaction with no queued agent messages does not resume; manual compaction still does not call `agent.continue()`; failed/aborted auto-compactions do not reach the resume gate.
- Tests: added a no-queue regression test in `packages/coding-agent/test/agent-session-auto-compaction-queue.test.ts` alongside the existing queued-continuation test.
- Changelog: updated `packages/coding-agent/CHANGELOG.md` under `[Unreleased]` → `### Fixed` with issue #1280.
- Validation: `bun --cwd=packages/coding-agent run test -- test/agent-session-auto-compaction-queue.test.ts` passed (1 file, 7 tests). `bun run typecheck` passed (`tsc --noEmit`). No `issues.md` remains.
- Note: the literal spec command form `bun --cwd packages/coding-agent run test -- ...` is invalid for this Bun version; validation used `--cwd=packages/coding-agent`.
- Deferred: no PR was created in iteration 1 per orchestrator instruction to ignore PR submission requests.

## Iteration 2 Notes

- Revised spec change: iteration 1 sampled `agent.hasQueuedMessages()` immediately after `compaction_end`, which could miss messages flushed asynchronously by `InteractiveMode` listeners. Iteration 2 replaces that immediate snapshot with a deferred post-event continuation probe.
- Implementation decision: added `_schedulePostAutoCompactionContinuationProbe(reason, willRetry)` in `packages/coding-agent/src/core/agent-session.ts`. It runs after successful auto-compaction and samples live `this.isCompacting`, `this.isStreaming`, and `this.agent.hasQueuedMessages()` inside the timer callback.
- Behavior decision: the probe returns without continuing if compaction or streaming has already started by probe time. Overflow retry still resumes even without queued messages, but trailing overflow assistant-error cleanup now happens inside the probe immediately before continuing.
- Test decision: added a regression where `compaction_end` observes no queued messages, then an async listener queues follow-up work before the deferred probe; this would fail under the iteration-1 immediate-snapshot implementation. Added streaming-suppression coverage and kept no-queue idle threshold coverage.
- Changelog: existing `[Unreleased]` fixed entry for #1280 remains appropriate; no additional changelog entry was added.
- Validation: `bun --cwd=packages/coding-agent run test -- test/agent-session-auto-compaction-queue.test.ts` passed (1 file, 9 tests). `bun run typecheck` passed (`tsc --noEmit`). No `issues.md` exists.
- Deferred: no PR was created in iteration 2 per instruction. Full repository/unit test suites were not run beyond targeted tests plus typecheck.

## Iteration 3 Notes

- Revised spec change: overflow retry-context cleanup must not be owned by the deferred probe because a `compaction_end` listener can start streaming before the probe fires, causing the probe's streaming guard to skip cleanup.
- Implementation decision: kept `_dropTrailingOverflowAssistantErrorIfPresent()` as an in-memory-only cleanup helper, but moved its call to immediately after `this.agent.state.messages = sessionContext.messages` and before `session_compact` / `compaction_end` for `reason === "overflow" && willRetry`.
- Behavior decision: `_schedulePostAutoCompactionContinuationProbe()` now owns only continuation. It still samples live `isCompacting`, `isStreaming`, and queued messages at callback time, but no longer mutates retry context.
- Test decision: added regression coverage that mocks rebuilt context with a trailing overflow assistant error and verifies a `compaction_end` listener observes cleaned in-memory context even when it starts streaming before the deferred probe. The probe then suppresses `agent.continue()` because streaming is already true.
- Changelog: existing `[Unreleased]` fixed entry for #1280 remains appropriate; no additional entry was added.
- Validation: targeted queue test passed (10/10); related compaction tests passed (2 executed, 5 skipped); `bun run typecheck` passed. No `issues.md` exists.
- Deferred: no PR was created in iteration 3 per instruction. No manual end-to-end interactive compaction session was performed.

PR #1326, opened 2026-06-10
fix(read): block oversized text file reads above 50k chars

### LINKED ISSUE #1323: Block oversized file reads with incremental-read guidance
## Summary

File-read tool calls should refuse to return file contents when the requested read would exceed roughly 20k tokens. Instead of dumping the oversized content into context, Atomic should return a clear error-style prompt telling the agent that the read is too large and that it must read only the needed context incrementally.

## Problem

Agents can accidentally request an entire large file and flood the transcript with low-value content. This wastes tokens, degrades model performance, and often hides the useful context inside a huge blob.

For file reads, writing the content into context is usually the wrong fallback when the file is too large. The agent already has a stable source path and can inspect smaller ranges as needed.

## Proposed behavior

- Estimate the token size of file-read results before returning content.
- If the requested read would be `> 20k` tokens, block the read result.
- Return a concise error-style message to the agent that includes:
  - the file path
  - the estimated token size, if available
  - the configured threshold
  - clear guidance to read smaller ranges or targeted snippets
  - examples of incremental reads when applicable
- Do not include the oversized file contents in the returned tool result.
- Prefer guidance that nudges the agent toward targeted context discovery, such as searching symbols, reading line ranges, or chunking the file.

Example returned message:

```text
File read blocked: requested content is too large (~31,200 tokens; limit: 20,000).
Path: /path/to/large-file.ts

Read only the needed context incrementally. For example:
- search for relevant symbols first
- read a smaller line range
- request chunks around the matching lines
```

## Acceptance criteria

- File reads at or below the threshold continue returning normally.
- File reads above `20k` tokens return a clear "too much read" error/prompt.
- Oversized file contents are not inserted into model context.
- The prompt tells the agent how to recover by incrementally reading needed context.
- Tests cover both allowed reads and blocked oversized reads.

lavaman131:
## Summary

Adds a 50,000-character safety gate to the Read tool that blocks text file reads where the selected range exceeds `READ_TOOL_MAX_RESULT_CHARS` (matching upstream's `DEFAULT_MAX_RESULT_SIZE_CHARS`), returning actionable incremental-read guidance instead of injecting oversized content into model context. Fixes [#1323](https://github.com/bastani-inc/atomic/issues/1323).

## Key Changes

- **Character gate** (`read.ts`): Checks `selectedContent.length > READ_TOOL_MAX_RESULT_CHARS` (50,000) **after** `offset`/`limit` selection and **before** `truncateHead`, so a targeted range inside a large file is allowed while a whole-file read of that same file is blocked
- **No content leakage**: Blocked results return guidance-only text — the oversized content is never assigned to a tool-result content block
- **Structured metadata**: Adds `ReadToolDetails.oversizedRead` (`OversizedReadDetails`) exposing `blocked`, `path`, `chars`, `maxChars`, `startLine`, `requestedLimit`, `totalFileLines`, `firstLineBytes`, and `byteGuidance` for extension hooks and tests
- **Actionable guidance**: Blocked message includes the file path, char count, threshold, and concrete `grep`/`read` examples for incremental recovery
- **Byte-slice guidance**: When the selected content is a single oversized line (or the first selected line exceeds `DEFAULT_MAX_BYTES`), the guidance switches to `sed`-based byte-window examples instead of useless line-pagination examples (`byteGuidance: true`)
- **Shell-safe paths**: Paths in `sed` guidance lines are single-quoted with proper escaping to prevent command injection
- **TUI visibility**: `formatReadResult` now renders blocked guidance in collapsed (non-error) mode, keyed off `details.oversizedRead.blocked`, and skips syntax highlighting even for source-extension paths (e.g. `.ts`) so the message is readable as tool output
- **Spec document**: Added `specs/2026-06-10-implement-github-issue-1323-...` documenting the design and deviation from the token-based approach described in the issue

## Test Coverage

Six cases added/updated in `packages/coding-agent/test/tools.test.ts`:

- **Threshold boundary**: file at exactly 50,000 chars passes through to the existing truncation path
- **Blocked + no-leak**: read above threshold returns guidance and omits a sentinel string embedded in the file content; structured metadata is validated
- **Incremental range**: small `offset`/`limit` slice of a large file is allowed and returns the expected lines
- **Single-line byte guidance**: single-line oversized file produces `sed`-based byte-window examples instead of line-pagination guidance
- **Shell-escape safety**: paths with special characters (`$(...)`, `'`) are correctly single-quoted in `sed` guidance
- **Rendering**: oversized `.ts` file renders as plain tool output (not syntax-highlighted TypeScript)
- **Byte-truncation regression**: existing byte-limit test adjusted to use UTF-8 multi-byte chars that stay within the 50k char threshold but still exceed the 50 KB byte limit

## Design Notes

- Threshold constant `READ_TOOL_MAX_RESULT_CHARS = 50_000` matches `DEFAULT_MAX_RESULT_SIZE_CHARS` from upstream; no tokenizer dependency added
- Gate runs after `offset`/`limit` selection so targeted ranges inside large files remain accessible
- Gate runs before `truncateHead` so the oversized content is never constructed as a "helpful" first chunk
- Existing 2,000-line / 50 KB truncation rail remains in place for all allowed reads

## Validation

- `bun run lint` ✅
- `bun run test:unit` ✅
- `bun run typecheck` ✅


lavaman131:
# Implementation Notes – Issue #1323: Block Oversized Reads

## Summary
The Read tool now guards against inserting large file contents into model context by estimating tokens for the **selected text range** (post offset/limit, pre-truncateHead) and rejecting reads above 20,000 estimated tokens.

## Token Estimation
- Uses `Math.ceil(text.length / 4)` — a fast char-count heuristic (≈ 4 chars/token).
- Applied to `selectedContent` after offset/limit slicing, before `truncateHead`.
- Threshold: `READ_TOOL_MAX_RESULT_TOKENS = 20_000` (strict `>`; exactly 20,000 is allowed).

## Decision: Estimate on selectedContent, not post-truncation
The check runs on the full selected slice (before truncateHead) to ensure that even if truncation would reduce output, the intent to read a huge range is blocked. This prevents leakage of an implied "big context window" from a large-but-truncatable file.

## No-Leak Guarantee
When blocked, `content` is set exclusively to the `buildOversizedReadMessage` string. The selected file content is never stored in any returned field.

## OversizedReadDetails shape
```ts
interface OversizedReadDetails {
  blocked: true;
  path: string;          // absolute path
  estimatedTokens: number;
  maxTokens: number;     // 20_000
  startLine: number;     // 1-indexed display line
  requestedLimit?: number;
  totalFileLines: number;
}
```
Added to `ReadToolDetails` as optional `oversizedRead` field (null when not blocked).

## Render-side behavior
`formatReadResult` checks `oversizedRead?.blocked` and:
1. Forces `expanded`-like display (message shown even without expanded flag).
2. Renders `buildOversizedReadMessage` instead of file content for the block message.

## Guidance message format
```
File read blocked: requested selected range is too large (estimated X tokens; threshold: 20,000).
Path: <absolutePath>

Read only the needed context incrementally. Examples:
- Search for relevant symbols first: grep({ ... })
- Read a smaller line range: read({ "path": ..., "offset": <startLine>, "limit": 200 })
- Read a targeted snippet around a match: read({ "path": ..., "offset": <max(startLine,120)>, "limit": 80 })
```

## Tradeoffs / Residual Risks
- **Heuristic accuracy**: 4 chars/token is an approximation. For typical ASCII code, 1 char ≈ 0.25 tokens is reasonable; for dense non-ASCII (e.g. CJK), a char may still be ~1 token so the estimate undershoots actual token count. The threshold is conservative (20K) and real model context windows are larger.
- **No escape via limit**: A user/agent could request `limit=1` on a 1M-line file and see the first line — this is intentional; the check is based on the *selected range* size, not the whole file.
- **Image path unchanged**: Image reads bypass this check entirely (correct — images are sized separately).
- **Existing truncation unchanged**: Allowed reads still go through `truncateHead` with the existing 2000-line / 50KB limits.

## Tests Added (tools.test.ts)
1. `should allow text reads at the oversized-read token threshold` — exactly 20,000 tokens (80,000 chars), verifies not blocked but still truncated.
2. `should block text reads above the oversized-read token threshold without leaking content` — sentinel pattern confirms no content leakage.
3. `should allow small ranges from files that would be oversized as whole-file reads` — `offset=100, limit=3` on a 1,000-line oversized file is allowed.

## Changelog
Entry added under `packages/coding-agent/CHANGELOG.md` → `## [Unreleased]` → `### Fixed`, referencing `#1323`.

## Validation Results (2026-06-10)

### Commands Run
- `bun run typecheck`: **PASSED** — `tsc --noEmit` exited cleanly, zero errors.
- `bun run test:unit`: **PASSED** — 2181 tests pass, 0 fail (6.54s via vitest/bun runner).

### Test Coverage Observations
- **Threshold boundary**: The threshold-allowed test creates exactly 80,000 chars (20,000 tokens), relies on `Math.ceil(80000/4)=20000`, and the `>` (not `>=`) condition correctly allows it. Truncation still fires because the 50KB byte limit / 2000 line limit is hit first.
- **No-leak sentinel**: The blocked test embeds `DO_NOT_LEAK_1323_CONTENT` and `z`×100 prefix; both are confirmed absent from output; `result.details?.oversizedRead?.blocked === true` and `result.details?.truncation === undefined`.
- **Small incremental range**: `offset=100, limit=3` on a 1000-line large file — only 3 lines ~309 chars ≈ 78 estimated tokens — correctly allowed through and shows continuation hint for remaining 898 lines.
- **Byte-truncation regression guard**: Existing test updated from 500 lines to 300 lines so it stays below 20K tokens and exercises the 50KB byte-truncation path (not the oversized-read block path). `expect(output).not.toContain("File read blocked")` was added to make this intent explicit.

### Render Behavior
`formatReadResult` now exposes blocked messages to the UI without requiring `expanded` mode (condition `!oversizedReadBlocked` added). The rendered message comes from `buildOversizedReadMessage(oversizedRead)` — same source as model content — keeping UI and model context consistent.

### No Staged Files
No source edits were made during this validation pass.

PR #1340, opened 2026-06-11
fix(workflows): clear ctx.ui errors in headless and partial-adapter runs

### LINKED ISSUE #1339: Bug: Headless/background runs crash on ctx.ui.custom (TypeError: ctx.ui.custom is not a function)
# Summary

 A workflow stage that renders an interactive component via ctx.ui.custom(...) crashes the whole run in headless/background (async) mode:

 ```
   TypeError: ctx.ui.custom is not a function
 ```

Headless runs have no attached TUI, so ctx.ui/ctx.ui.custom is undefined, but the stage assumes interactivity is always present and throws a terminal error.

# Repro

 1. Workflow stage calls await ctx.ui.custom(...) and awaits a user selection.
 2. Run it in headless/background (async) mode.
 3. It reaches that stage and fails.

# Expected

 In headless mode, don't hard-crash — fall back to a non-interactive path, or surface a clear error ("interactive selection unavailable in headless mode")
 instead of a raw TypeError.

# Actual

 Earlier stages complete; the run dies the moment the interactive stage calls ctx.ui.custom.

lavaman131:
## Summary

Prevents headless/background workflow runs from hard-crashing with a raw `TypeError: ctx.ui.custom is not a function` when a stage calls any interactive `ctx.ui.*` primitive in non-interactive mode. Extends the same protection to interactive runs where the UI adapter is partial (missing optional methods like `custom`).

Closes #1339

### Core fix — `packages/workflows/src/runs/foreground/executor.ts`

- **`makeRejectingUIContext(msg)`** (extracted): shared implementation for all "rejecting" UI contexts, parameterized by a message factory so the two error-path variants never drift.
- **`makeHeadlessUnavailableUIContext()`** (new): used in `non_interactive` mode when no `ui` adapter is provided. Every interactive primitive (`input` / `confirm` / `select` / `editor` / `custom`) rejects with a clear, actionable error:
  ```
  atomic-workflows: interactive ctx.ui.custom is unavailable in headless
  (non-interactive) mode; run the workflow in interactive mode or remove
  the interactive prompt from this stage
  ```
  instead of surfacing a raw `TypeError` that kills the whole run.
- **`normalizeUIContext`** hardened with `typeof === "function"` guards on every adapter method, so partial/missing adapter methods fall back to the same clear rejection rather than `x is not a function`.
- **`run()`** updated: headless runs (`executionMode === "non_interactive"` + no `ui` adapter) now receive `makeHeadlessUnavailableUIContext()` instead of falling through to `normalizeUIContext(undefined)`.
- Interactive background runs continue to broker `ctx.ui.custom` through `awaiting_input` prompt nodes; earlier completed stages remain completed.

### Tests — `test/unit/headless-custom-ui.test.ts` (new)

Four regression tests covering the full bug surface (all red before the fix, green after):

1. **Headless run fails with a clear error, not a TypeError** — asserts `run.status === "failed"`, error matches `ctx.ui.custom` and `headless`, and the earlier `warmup` stage stays `completed`.
2. **All five interactive primitives get the headless treatment** — iterates `input` / `confirm` / `select` / `editor` / `custom` in `non_interactive` mode.
3. **Interactive background run brokers `ctx.ui.custom` as `awaiting_input`** — resolves the brokered request via `stageUiBroker.registerHost`, asserts `run.status === "completed"` with expected output.
4. **Partial UI adapters degrade to clear errors** — passes an empty `{}` adapter to `run()`, asserts no `is not a function` in the error.

### Manual verification workflow — `.atomic/workflows/hil-custom-dummy.ts` + `.atomic/workflows/lib/table-selector.ts` (new)

A hands-on repro/verification workflow whose `ctx.ui.custom<TableChoice>` mounts a themed, sortable, responsive table-selector TUI widget. The widget factory lives in `lib/` (not at the top level) to avoid the discovery validator rejecting it as an invalid workflow export — see documentation change below.

### Docs — `packages/coding-agent/docs/workflows.md` + `packages/workflows/README.md`

- Documents the **helper-module subdirectory pattern**: every runtime export of a top-level workflow file is validated as a workflow definition; non-workflow helpers (widget factories, shared utilities) must live in a subdirectory such as `.atomic/workflows/lib/` that discovery does not scan.
- Adds a cross-reference from the *Programmatic Usage* section to *Workflow Locations*.
- Updates the headless error message example in both docs to match the new, more actionable wording.

### Changelog — `packages/workflows/CHANGELOG.md`

Added `### Fixed` entry under `[Unreleased]` referencing #1339.

## Validation

- `bun run typecheck`, `bun run lint`, and `bun run test:unit` (2227 pass / 0 fail) all green
- Pre-commit hooks (prek) ran lint + unit tests on every commit
- **Interactive**: `/workflow hil-custom-dummy` → warmup completes → focus the `custom` awaiting-input node in the graph viewer and drive the table selector
- **Headless**: `atomic -p '/workflow hil-custom-dummy skip_warmup=true'` now surfaces the clear headless-unavailable error instead of crashing with a raw TypeError


lavaman131:
Thanks for the thorough review — all four items addressed, plus a rebase onto `main` (the `ctx.exit()` #1341 changes). Force-pushed.

### 1. Docs drift ✅
Updated both `packages/workflows/README.md` and `packages/coding-agent/docs/workflows.md` to quote the new headless wording (`interactive ctx.ui.confirm is unavailable in headless (non-interactive) mode; …`) in a dedicated commit.

### 2. Which message fires in production headless? → the new one (branch is reachable, not dead) ✅
The host passes **no** `ui` adapter headlessly. In `runs/background/runner.ts` the detached runner explicitly destructures `ui` out and never forwards it to `RunOpts`:

```ts
const { jobs: _jobs, cancellation: _cancellation, ui: _ui, store: storeOverride, ...restOpts } = opts;
const execOpts: RunOpts = {
  ...restOpts,                       // ui is NOT here → opts.ui === undefined
  usePromptNodesForUi: opts.executionMode !== "non_interactive",
  ...
};
```

So for **every** detached run `opts.ui === undefined`, and in headless mode `usePromptNodesForUi` is `false`. That means the new `executionMode === "non_interactive" && opts.ui === undefined` branch **does fire in production headless** → users see the new headless message, confirming the docs in #1 were stale (now fixed). With a *partial* adapter it would fall to `normalizeUIContext` (old message), but the runner never feeds one in production — partial adapters only arrive via direct foreground `run()` / SDK callers, which the per-method guards cover.

### 3. Changelog root-cause framing ✅
You're right — pre-PR, the headless-no-adapter path already went through `normalizeUIContext(undefined)` → `makeUnavailableUIContext()`, a graceful rejection (with `custom` already guarded). The genuine raw `is not a function` `TypeError` was only reachable for **partial** adapters via `input`/`confirm`/`select`/`editor`. I reworded the CHANGELOG and the PR description to split the two: (a) partial adapters → guarded against the real `TypeError`, (b) headless-no-adapter → clearer message replacing the generic one. I also re-checked the #1339 path: the crash a workflow author hits is `WorkflowUIContext.custom` from the executor (this file), not `wiring.ts`/overlay adapters — the stage extension UI context isn't even bound in `non_interactive` mode (`shouldBindStageUiContext` returns `false`).

### 4. DRY ✅
Collapsed both factories into a single parameterized `makeRejectingUIContext(msg)`; `makeUnavailableUIContext` and `makeHeadlessUnavailableUIContext` are now thin wrappers, so the two messages can't drift.

### Minor nits
- **table-selector `selected` index** — agreed, it's index-sticky not meaning-sticky; acceptable for the dummy repro widget, leaving as-is.
- **test polling deadlines** — noted; the 2000ms deadlines with 5ms poll intervals are comfortably generous. Full `bun run test:unit` is green after the rebase.

PR #1351, opened 2026-06-12
fix(workflows): sanitize release-docs git subprocess env

lavaman131:
## Summary

Strips repository-local Git environment variables (`GIT_DIR`, `GIT_WORK_TREE`, `GIT_INDEX_FILE`, and others) from subprocesses in the release-docs workflow and test fixtures, preventing Git hook runners (e.g. `prek`) from silently redirecting commands at the wrong repository.

## Root Cause

Git honors repository-local env vars over `cwd` — hook runners like `prek` export these vars into the hook process environment, so any transitively spawned Git command (including `git init` on a temp directory) operates on the hook-invoking repo instead of the intended one. In the worst case, `git init` persists `core.worktree` into the shared `.git/config` of the real repo, corrupting it.

## Changes

- **`.atomic/workflows/lib/release-docs.ts`**: Pass `env: createGitEnvironment()` to `execFileSync` inside `runCommand()`, stripping all repository-local Git env vars so every Git subprocess always targets its explicit `cwd`.
- **`test/unit/release-docs-workflow.test.ts`**:
  - Apply the same `createGitEnvironment()` sanitization to `runGit()` in test fixtures, preventing fixture-repo `git init` calls from corrupting the real repo under hook runners.
  - Extract a `commitAll()` helper to deduplicate repeated commit invocations across tests.
  - Add a regression test that spawns a child Bun process with ambient hook-style env vars (`GIT_DIR`, `GIT_WORK_TREE`, `GIT_INDEX_FILE` pointing at a detached decoy repo) and asserts `currentBranchName()` still returns the correct branch of the target repo.

## Validation

- `bun run typecheck`
- `bun run lint`
- `AGENT=1 bun run test:unit`
- Pre-commit and pre-push hooks passed during commit/push

### DIFF
--- a/.atomic/workflows/lib/release-docs.ts
+++ b/.atomic/workflows/lib/release-docs.ts
@@ -1,5 +1,6 @@
 import { execFileSync } from "node:child_process";
 import { join } from "node:path";
+import { createGitEnvironment } from "../../../packages/coding-agent/src/utils/git-env.js";
 
 export type StaleDocTask = {
   id: string;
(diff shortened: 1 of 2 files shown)


lavaman131:
Addressed the non-blocking feedback in `aae1df73 fix(workflows): use public atomic git env export`:

- Switched both release-docs implementation and regression test imports to the public `@bastani/atomic` barrel for `createGitEnvironment`.
- Added an explicit test comment noting `process.execPath` is intentionally the Bun runtime for the child TypeScript script.

The branch is pushed and clean.

PR #1393, opened 2026-06-15
fix(compaction): preserve thinking blocks during overflow

### LINKED ISSUE #1386: Verbatim Compaction can corrupt the latest assistant message's `thinking` blocks
## Summary

After context compaction runs, an Anthropic-backed session can become permanently unusable. Every subsequent request fails with:

```
400 {"type":"error","error":{"type":"invalid_request_error",
"message":"messages.27.content.7: thinking or redacted_thinking blocks in the latest assistant message
cannot be modified. These blocks must remain as they were in the original response."},
"request_id":"req_011Cc3WDBJ3zUGKrg3p53vxx"}
```

Verbatim Compaction deletes individual content blocks `(entryId, blockIndex)` inside messages. Its validation/protection layer does **not** know about Anthropic's extended-thinking replay invariant: when extended thinking is enabled, the `thinking`/`redacted_thinking` blocks in the **latest assistant message** must be sent back **byte-for-byte and in the same position** as originally returned. A `content_block` deletion that (a) removes a `thinking` block, or (b) removes any neighboring block in an assistant message that contains a thinking block, mutates that assistant turn's content array. The Anthropic provider faithfully replays the mutated structure on the next request, and Anthropic rejects it.

Because the bad deletion is persisted as a `context_compaction` entry, the corruption is **sticky**: the active context is rebuilt from that entry on every turn, so the session stays broken until the entry is removed / restored from the `.compact.bak` snapshot.

- **Severity:** High (one accepted compaction target bricks the session)
- **Affected versions:** observed on `@bastani/atomic` 0.8.28; root cause present on `main`
- **Provider lib:** `@earendil-works/pi-ai@0.79.3`
- **Scope:** Anthropic (extended thinking) and any same-model path that replays signed/redacted thinking blocks

## Reproduction

1. Run an Anthropic model with extended thinking enabled (interleaved thinking + tool use), so assistant turns contain `thinking` (and possibly `redacted_thinking`) blocks alongside `text`/`toolCall` blocks.
2. Grow the conversation until auto-compaction triggers (`contextTokens > contextWindow - reserveTokens`), or run `/compact`. A context-overflow recovery pass (`critical_overflow` mode) makes it more likely, since its prompt instructs the planner to "remove old reasoning traces first."
3. The deletion planner (or an extension `deletionRequest`) selects a `content_block` target that is, or sits next to, a `thinking` block in the assistant message that is the **latest assistant turn** of the rebuilt context.
4. Next request → `400 ... thinking or redacted_thinking blocks in the latest assistant message cannot be modified`.
5. Every following turn fails identically, because the persisted `context_compaction` entry keeps re-applying the bad deletion during `buildSessionContext()`.

### 1. Thinking blocks are treated as ordinary, deletable content blocks

`packages/coding-agent/src/core/compaction/context-compaction.ts`

`textFromContentBlock()` makes thinking text searchable/readable by the planner (~L394–399):

```ts
if (record.type === "thinking" && typeof record.thinking === "string") return record.thinking;
```

`contentBlocksForEntry()` records every block by index and sets `protected` **only** from the containing entry (~L445–460). There is no special handling for `type: "thinking"`, redacted thinking, or `thinkingSignature`:

```ts
return {
  entryId,
  blockIndex,
  type: ...,
  text,
  tokenEstimate: estimateContentBlockTokens(block, text),
  protected: protectedEntry,   // <-- inherited from entry only
  toolCallId: getToolCallIdFromBlock(block),
};
```

The `critical_overflow` prompt even prioritizes deleting reasoning traces (~L1857):

```ts
return `... Evict in priority order: remove old reasoning traces first, then old user/custom/summary context, ...`;
```

### 2. "Latest assistant message" is not a protection concept

`isProtectedEntry()` protects user/custom/branch-summary/error/failed-bash entries and the **last five context-eligible entries** (~L503–532), but never identifies "the latest assistant message" specifically:

```ts
if (recentEntryIds.has(entry.id)) return true;
if (message.role === "user") return true;
// ...no rule for "latest assistant" or "assistant contains thinking"
```

An assistant turn with tool calls can be followed by many tool-result / bash / custom entries, so it can be the latest *assistant* turn while falling **outside** the last-five-entries window — i.e., unprotected.

### 3. Validation has no thinking / latest-assistant safety check

`validateContextDeletionRequest()` (~L924–1014) checks unknown IDs, protected targets, duplicates, "only block" deletions, overlap with entry deletion, "would remove every content block", tool-call/result orphaning, all-context, and task-bearing context. It does **not** check:

- `block.type === "thinking"` / `"redacted_thinking"`
- whether the entry is the **latest assistant message**
- whether a (non-thinking) deletion would reposition a sibling thinking block

```ts
if (deletion.kind === "content_block") {
  // ...index/exists/protected/only-block checks only
}
// ...reconcileToolDependencies() can ADD assistant content-block deletions here...
const reconciledTargets = reconcileToolDependencies(transcript, deletedTargets);
// ...no thinking-safety re-check on reconciledTargets
```

Note: `reconcileToolDependencies()` (~L771–779) can introduce assistant `content_block` deletions that the planner never directly requested, so any guard must run on the **final reconciled targets**, not just the raw request.

### 4. Rebuild blindly index-filters assistant content

`packages/coding-agent/src/core/session-manager.ts`

`buildSessionContext()` → `buildContextDeletionFilteredPath()` (~L429–447) → `filterMessageContentBlocks()` (~L388–405) uses a generic index filter for assistant messages (~L384):

```ts
function filterContentArray<T>(content: T[], deletedBlocks: ReadonlySet<number>): T[] {
  return content.filter((_, index) => !deletedBlocks.has(index));
}
// case "assistant": const content = filterContentArray(message.content, deletedBlocks); ...
```

`Array.filter` preserves survivor order but **reindexes** survivors. Anthropic's constraint is positional + byte-exact, so even dropping an *earlier non-thinking* block in the latest assistant turn corrupts the thinking block's position.

### 5. The provider replays the mutated thinking blocks

`@earendil-works/pi-ai@0.79.3`

- `transform-messages.ts` (~L97–110) keeps same-model thinking blocks with signatures (and same-model redacted thinking) — not "latest only".
- `anthropic.ts convertMessages()` maps internal thinking back to API `thinking` (with `signature`) / `redacted_thinking` (~L1071–1102).

So whatever `buildSessionContext()` produces for the latest assistant turn is sent verbatim to Anthropic — including the corrupted content array — triggering the 400.

### End-to-end path

`agent-session.ts _applyContextVerbatimCompaction()` (prepare ~L2057, validate ~L2128, persist + rebuild ~L2156–2164) → `appendContextCompaction()` persists the bad targets → `buildSessionContext()` re-applies them every turn → provider replays corrupted latest-assistant thinking → 400, repeatedly.

## Why existing protections miss it

Protection today is entry-/task-oriented (recent 5 entries, user/custom/summary, errors, failed bash, tool-call/result consistency). It does not model the **provider replay invariant** for assistant thinking blocks. Thinking blocks inherit protection only from their entry; the latest assistant message is never specifically protected from *content-block* deletion; filtering is type-blind; and tool-dependency reconciliation can add assistant content-block deletions after the raw checks.

### A. Validation guard (prevents new corruption) — primary

In `validateContextDeletionRequest()` (`context-compaction.ts`), add an Anthropic-thinking-safety pass applied to the **final `reconciledTargets`**:

1. Reject any `content_block` deletion whose target block is `type === "thinking"` / `"redacted_thinking"`.
2. Reject any **partial** `content_block` deletion of an assistant message that contains a thinking/redacted-thinking block (deleting a neighbor repositions the thinking block).
3. Reject any partial `content_block` deletion of the **latest retained assistant message**.

Whole-entry deletion of such assistant messages can remain allowed (subject to existing tool/task validation), since that removes the message entirely rather than mutating its thinking blocks in place.

### B. Rebuild guard (recovers already-broken sessions) — defensive

In `session-manager.ts` `filterMessageContentBlocks()` / `buildContextDeletionFilteredPath()`, when applying persisted deletions, **skip** partial content-block filtering for:

- the latest retained assistant message, and
- any assistant message containing thinking/redacted-thinking blocks,

i.e., keep that message verbatim instead of partially filtering it. Prefer skipping (and logging a recoverable diagnostic) over throwing, so existing sessions with a bad persisted `context_compaction` entry become usable again without manual JSONL surgery.

Both fixes are needed: (A) stops producing bad entries; (B) repairs sessions that already contain one.

## Suggested tests

1. `validateContextDeletionRequest()` rejects a `content_block` deletion of a `thinking` block in an unprotected assistant entry.
2. It rejects deletion of a non-thinking block in the latest assistant message when that message contains thinking.
3. It rejects unsafe assistant content-block deletions **introduced by `reconcileToolDependencies()`** (validate the reconciled set).
4. `buildSessionContext()` preserves the latest assistant content verbatim when a persisted "bad" `context_compaction` entry exists (rebuild guard).
5. A fixture with `redacted: true` thinking is protected identically to a normal signed thinking block.

## Residual risks / notes

- Fixing only validation leaves already-corrupted sessions broken (they keep re-applying persisted targets) — hence the rebuild guard.
- If the rebuild guard silently skips unsafe deletions, compaction stats/`deletedTargets` may no longer exactly match the active context; log this. Still preferable to a permanently dead session.
- `pi-ai` currently replays *all* same-model historical thinking, not just the latest turn. If Anthropic ever enforces immutability beyond the latest assistant message, the broader "no partial filtering of any assistant message containing thinking" rule already covers it.

## Key references (paths relative to `packages/coding-agent/src`)

- `core/compaction/context-compaction.ts`: `textFromContentBlock` (~L394), `contentBlocksForEntry` (~L439–460), `isProtectedEntry` (~L503–532), `reconcileToolDependencies` (~L739–779), `validateContextDeletionRequest` (~L924–1014), `critical_overflow` prompt (~L1857).
- `core/session-manager.ts`: `filterContentArray` (~L384), `filterMessageContentBlocks` (~L388–405), `buildContextDeletionFilteredPath` (~L429–447), `buildSessionContext` (~L467–520).
- `core/agent-session.ts`: `_applyContextVerbatimCompaction` (~L2057, L2128, L2156–2164).
- `@earendil-works/pi-ai@0.79.3`: `providers/transform-messages.ts` (~L97–110), `providers/anthropic.ts convertMessages` (~L1071–1102).

lavaman131:
## Summary

Prevents critical-overflow context compaction from deleting, partially filtering, or instructing removal of assistant `thinking`/`redacted_thinking` content blocks. Also repairs unsafe persisted deletion filters during session rebuild so existing affected sessions recover without mutating Anthropic thinking-bearing assistant turns.

Closes #1386.

### New: `thinking-blocks.ts`

Extracted shared detection utilities:
- `isAssistantThinkingBlockType(type)` — type guard for `"thinking"` / `"redacted_thinking"`
- `contentArrayHasAssistantThinkingBlock(content)` — scans a content array
- `messageHasAssistantThinkingContentBlock(message)` — checks a full message

### Core compaction (`context-compaction.ts`)

- Adds `assistantEntryHasThinkingContentBlock` helper to detect thinking-bearing `CompactableTranscriptEntry` values
- Adds `assertNoAssistantThinkingDeletionTargets` — throws if any reconciled deletion target touches an assistant entry containing thinking blocks (whole-entry or per-block); runs post-reconciliation in standard mode
- Adds `assertNoLatestRetainedThinkingAssistantContentBlockDeletionTargets` — throws during `critical_overflow` if a content-block deletion targets the latest retained thinking-bearing assistant entry
- Updates `validateContextDeletionRequest` to call both assertions after tool-dependency reconciliation
- Updates `createContextDeletionTool` grep path to skip (rather than match) assistant entries and blocks belonging to thinking-bearing entries, emitting `"assistant_thinking_entry"` / `"assistant_thinking_block"` skip reasons
- Adds `has_assistant_thinking_blocks`, `role`, and `type` fields to `StoredTranscriptEntry`, `StoredContentBlock`, and their SQL row types
- Rewrites `contextCompactionModePrompt` for `critical_overflow`: removes the instruction to evict reasoning traces first; adds an explicit invariant prohibiting deletion of thinking-bearing assistant entries or their sibling blocks
- Switches `prepareContextCompaction` to use `buildEffectiveContextDeletionFilters` instead of raw `buildContextDeletionFilters`

### Session rebuild (`session-manager.ts`)

- Exports `buildEffectiveContextDeletionFilters` — replays all compaction entries and skips any deletion target whose assistant message carries a thinking block
- When a content-block target for the latest retained thinking-bearing assistant is skipped, collects paired `toolCall` IDs and finds their `toolResult` entries so active context has no dangling tool calls
- Later valid content-block deletion filters for restored multi-block tool-result entries are still applied (provenance-aware repair)
- Replaces the default argument of `buildContextDeletionFilteredPath` and the call in `branch-summarization.ts` with `buildEffectiveContextDeletionFilters` so every session rebuild is safe

### Tests

- `context-compaction-deletion-tool.test.ts` — grep-path skips on thinking entries and blocks, skip reason assertions
- `context-compaction.test.ts` — `validateContextDeletionRequest` rejection of thinking-block targets (whole-entry and per-block), critical-overflow prompt wording invariant
- `build-context.test.ts` — persisted-filter repair: stale whole-entry deletion, stale block deletion, paired tool-result restoration, provenance-aware preservation of later valid block filters on restored tool results

### Docs & changelog

- `docs/compaction.md` — extends the critical-overflow description to document the thinking-block verbatim-preservation invariant and the session-rebuild repair behavior
- `CHANGELOG.md` — `### Fixed` entry under `## [Unreleased]` referencing #1386

## Breaking Changes

None. All changes are additive guard-rails and internal behavior fixes; no public API surfaces changed.

## Validation

```sh
bun test packages/coding-agent/test/session-manager/build-context.test.ts \
  packages/coding-agent/test/context-compaction.test.ts \
  packages/coding-agent/test/context-compaction-deletion-tool.test.ts
# Result: 73 pass, 0 fail

bun run typecheck
# Result: tsc --noEmit — passed
```

Pre-push/pre-commit hooks ran successfully (`bun run lint`, `bun run test:unit`).


lavaman131:
Updated per review/user clarification: critical_overflow may now evict old/non-latest thinking-bearing context, while deterministic validation/rebuild safeguards protect the latest retained thinking-bearing assistant message from content-block mutation. I also added a clear context_delete self-correction error for non-deletable latest thinking blocks.\n\nValidation rerun:\n- `bun test packages/coding-agent/test/session-manager/build-context.test.ts packages/coding-agent/test/context-compaction.test.ts packages/coding-agent/test/context-compaction-deletion-tool.test.ts` — 75 pass, 0 fail\n- `bun run typecheck` — passed\n- `git diff --check` — passed

lavaman131:
Addressed the latest review feedback.\n\nChanges in the pushed update:\n- Rebuild now applies accepted old/non-latest thinking content-block deletions, so critical_overflow actually reclaims that context instead of no-oping.\n- The latest-retained-assistant definition is aligned between validation and rebuild: only the absolute latest retained assistant is protected when it contains thinking/redacted_thinking.\n- Paired tool results restored because of a skipped latest-assistant partial filter are protected from later stale whole-entry deletion, preventing dangling retained tool calls. Later valid content-block trimming of those result entries still works.\n- Added regression tests for old thinking block rebuild eviction and later stale whole-result deletion.\n\nValidation rerun:\n- `bun test packages/coding-agent/test/session-manager/build-context.test.ts packages/coding-agent/test/context-compaction.test.ts packages/coding-agent/test/context-compaction-deletion-tool.test.ts` — 77 pass, 0 fail\n- `bun run typecheck` — passed\n- `git diff --check` — passed

lavaman131:
Addressed the latest review feedback.\n\nUpdates pushed:\n- Removed the redundant effective-filter recomputation in `prepareContextCompaction`: it now computes effective filters once and passes them to `buildContextDeletionFilteredPath`, whose optional parameter is documented as precomputed effective filters.\n- Updated branch summarization to use `buildContextDeletionFilteredPath(entries)` directly so persisted filter repair is consistently applied.\n- Simplified the hoisted `let block` / `??=` into a local `const block` in the content-block branch.\n- Restored the cheaper early-return ordering in `contentBlocksForEntry`.\n- Added a comment clarifying why standard-mode grep distinguishes thinking blocks from non-thinking sibling blocks in thinking-bearing assistant messages.\n\nValidation rerun:\n- `bun test packages/coding-agent/test/session-manager/build-context.test.ts packages/coding-agent/test/context-compaction.test.ts packages/coding-agent/test/context-compaction-deletion-tool.test.ts` — 77 pass, 0 fail\n- `bun run typecheck` — passed\n- `git diff --check` — passed\n- Commit/push hooks also passed, including `bun run lint` and `bun run test:unit`.

lavaman131:
Addressed the latest review feedback.\n\nUpdates pushed:\n- Corrected the PR description to state that latest-assistant protection is implemented by effective deletion-filter repair, not a `filterMessageContentBlocks` no-op guard.\n- Added a warning doc comment to `buildContextDeletionFilters` explaining it returns raw/unrepaired filters and production rebuilds should use `buildEffectiveContextDeletionFilters` or `buildContextDeletionFilteredPath(path)`.\n- Extracted thinking/redacted-thinking detection into shared `src/core/thinking-blocks.ts` and wired both compaction validation and session rebuild through it.\n- Added a regression test that whole-entry deletion of the latest thinking-bearing assistant is allowed during `critical_overflow`, while partial latest content-block deletion remains rejected.\n\nValidation rerun:\n- `bun test packages/coding-agent/test/session-manager/build-context.test.ts packages/coding-agent/test/context-compaction.test.ts packages/coding-agent/test/context-compaction-deletion-tool.test.ts` — 78 pass, 0 fail\n- `bun run typecheck` — passed\n- `git diff --check` — passed\n- Commit/push hooks also passed, including `bun run lint` and `bun run test:unit`.

Issue #1464, opened 2026-06-21
open-claude-design drops Playwright annotation feedback between preview and refinement stages


flora131:
## Summary

`open-claude-design` captures interactive Playwright annotation feedback in `preview-display-*` stages, but that feedback is not reliably propagated into the subsequent `user-feedback-*` and `apply-changes-*` stages. As a result, a user can annotate a preview, see the feedback captured in the preview-display stage summary, but the next iteration applies only internal critique/QA changes and produces a visually near-identical design.

Observed in a real run while redesigning the Atomic website.

## Version / scope

Confirmed on `@bastani/atomic@0.9.0-alpha.2`.

Evidence:
- Local executable: `/Users/norinlavaee/.bun/bin/atomic`
- `atomic --version` returned `0.9.0-alpha.2`
- Installed package metadata: `/Users/norinlavaee/.bun/install/global/node_modules/@bastani/atomic/package.json` had `"version": "0.9.0-alpha.2"`
- Affected workflow run: `207579f1-c448-43fa-8a86-6ba408c6db63`

I have **not** re-run the scenario on `0.9.0-alpha.3`, but I compared the relevant `open-claude-design` workflow files between alpha2 and alpha3 and they appear unchanged:

- `packages/workflows/builtin/open-claude-design.ts`
- `packages/workflows/builtin/open-claude-design-runner.ts`
- `packages/workflows/builtin/open-claude-design-phases.ts`
- `packages/workflows/builtin/open-claude-design-utils.ts`
- `packages/workflows/builtin/open-claude-design.d.ts`

Alpha3 release notes also appear unrelated to this workflow. Treat this as **confirmed on alpha2 and likely still present in alpha3/latest prerelease unless fixed outside these workflow files**.

## What happened

Run: `207579f1-c448-43fa-8a86-6ba408c6db63`
Workflow: `open-claude-design`

The first generated preview was shown through Playwright interactive annotation. The user left 5 concrete annotations:

1. Hero background: simplify it to a black → grey gradient with solid texture.
2. Top-left masthead text: too light on the background, ensure WCAG/a11y standards across the page.
3. Copy button font: too generic, improve with better design craft.
4. `∀ ❯ Start a loop` CTA: good, keep it.
5. Pixel-art `∀` / overall vibe: make it more polished, closer to Apple-level polish.

The annotations were captured successfully in:

- Stage: `preview-display-initial`
[...]

PR #1470, opened 2026-06-22
feat(open-claude-design)!: discovery, init, reference discovery, live QA

flora131:
## Summary

Restructures the builtin `open-claude-design` workflow around the accessible `impeccable` skill (`/skill:impeccable …`). The workflow now opens with a **discovery interview**, always bootstraps project design context via `init`, runs design-system onboarding + gallery reference discovery + reference import in a single concurrent context phase, and uses `impeccable live` for in-browser element-level QA. Also fixes annotation feedback being silently dropped between preview and refinement stages, and bundles a pending impeccable skill v3.8.0 refresh.

## Breaking Changes

Three top-level inputs are removed; the discovery stage now gathers them interactively.

| Removed input | Replacement |
|---|---|
| `reference` | Provide in `prompt` or answer the discovery interview |
| `output_type` | Confirmed during discovery (`prototype`/`wireframe`/`page`/`component`/`theme`/`tokens`) |
| `design_system` | Auto-detected and bootstrapped by the always-run `init` stage |

**Remaining inputs:** `prompt`, `discover_references` (default `true`), `max_refinements`.

**Migration:** drop `reference=…`, `output_type=…`, `design_system=…` from existing invocations, or fold them into `prompt`. The discovery interview will ask for anything it can't infer.

```diff
- /workflow open-claude-design prompt="Design a billing page" reference=https://stripe.com/billing output_type=page
+ /workflow open-claude-design prompt="Design a billing page like Stripe's"

- /workflow open-claude-design prompt="Generate spacing and color tokens" output_type=tokens design_system=./DESIGN.md
+ /workflow open-claude-design prompt="Generate spacing and color tokens"
```

### `open-claude-design` workflow — new phase order

1. **Discovery** (`/skill:impeccable shape`): new first stage interviews the user for a confirmed brief, output type, and references via the structured question tool, returning `{ brief, output_type, references }`. User-supplied references take **precedence over `DESIGN.md`/`PRODUCT.md`** (DESIGN.md fills gaps; PRODUCT.md governs register/voice). A `REFERENCE_PRECEDENCE` note threads into the import, generator, and apply-changes prompts. Headless runs infer a defensible brief without blocking.

2. **Init (always runs)** (`/skill:impeccable init`): detects `PRODUCT.md`/`DESIGN.md` (root, `.agents/context/`, or `docs/`, case-insensitive), creates whichever is missing, and reconciles existing files against the discovery brief without clobbering them. Reuses discovery answers so nothing is asked twice. Best-effort; never blocks the run.

3. **Combined context phase (single concurrent fan-out)**: design-system onboarding (`ds-locator` → `ds-analyzer` → `ds-patterns`), gated gallery **reference-discovery** (Awwwards, recent.design, Dribbble, Monet, Motionsites — `playwright-cli` clicks into standout work, records a scroll-through video for animation fidelity, falls back to full-page screenshot + real URL, persists to `<artifact_dir>/references.md`; web-search fallback if browser is unavailable), and per-reference **import** (`web-capture-<n>` / `file-parser-<n>`) all run in one `ctx.parallel` fan-out before `design-system-builder` synthesizes the design system. Set `discover_references=false` to skip the gallery pass.

4. **Generation** → 5. **Live QA** → 6. **Export gate**

### Live QA (`/skill:impeccable live`)

`preview-display-*` stages drive `/skill:impeccable live` against the static `preview.html` — pick elements in the browser, annotate them, compare three on-brand variants, accept one that is written back in place. Degrades to `playwright-cli show --annotate`, then a manual file path.

### New modules

- `open-claude-design-setup.ts`: `runDiscovery`, `detectDesignContextFiles`, `ensureProjectDesignContext`, `buildReferenceDiscoveryPrompt`, `persistReferencesBrief`, `buildLivePreviewDisplayPrompt`
- `open-claude-design-feedback.ts`: feedback parser, merged-brief builder, guardrail
- `discoveryDecisionSchema` / `discoveryDecisionFromResult` / `REFERENCE_PRECEDENCE` added to utils

### New unit tests

Added coverage for: discovery schema/parse, always-run init + reconcile + discovery-context threading, combined context fan-out, reference-discovery prompt/persist, live-QA prompt, trimmed input contract, full-run init path, `live_changes` capture/threading/persistence, feedback parser, merged-brief ordering, guardrail, and artifact persistence.

### Annotation feedback dropped between preview and refinement ([#1464](https://github.com/bastani-inc/atomic/issues/1464))

`preview-display-*` results were awaited-and-discarded, so a user could annotate a preview and watch the next iteration apply only internal critique while the design stayed visually unchanged.

**Fix:** Captured annotations are now parsed (tolerant of heading/bullet/bold/backtick label styles and placeholder values), carried through the loop as `previewFeedbackHistory`, written to `<artifact_dir>/feedback/iteration-<n>.md/.json`, and ordered **above** reviewer decisions and impeccable critique in the merged refinement brief. `user-feedback-*` receives them in a dedicated `<user_annotations>` block and will not approve export while honest feedback is unaddressed. A guardrail asserts annotation threading before each apply stage and fails loudly if it's missing.

### Artifact dir pollution in test runs

`prepareArtifactDir` now writes to the OS tmpdir when `NODE_ENV=test`, preventing per-run folders from appearing in `specs/design/`.

### Export gate + terminal annotation hardening

- Export gate refuses `ready_for_export` while unaddressed annotations exist (forces an apply pass via `assertUserAnnotationsThreaded`)
- Final refinement iteration and post-export `final-display` are read-only (no longer solicit `user_notes`/`live_changes` they cannot apply)
- `assertUserAnnotationsThreaded` now also enforces accepted `live_changes` threading
- Deterministic browser-centric early exit: when `playwright-cli` is unavailable the run calls `ctx.exit()` with install instructions instead of generating a design no one can review
- `copyAnnotationArtifacts` refuses to copy a model-supplied snapshot path that resolves outside the project/artifact dir

## Bundled impeccable skill v3.8.0 refresh

- Reorganizes live-mode helpers under `scripts/live/` and shared utilities under `scripts/lib/`
- Adds `detector/design-system.mjs` (design-system detection) and inline-ignore support (`scripts/detector/shared/inline-ignores.mjs`)
- Removes deprecated standalone scripts (`cleanup-deprecated.mjs`, `impeccable-paths.mjs`)
- Significant expansion of `live-browser.js`, `live-server.mjs`, `scripts/context.mjs`, and detector rules
- New `scripts/live/` modules: `manual-apply.mjs`, `manual-edit-routes.mjs`, `svelte-component.mjs`, `sveltekit-adapter.mjs`, `ui-core.mjs`, `vocabulary.mjs`

## Notes

- All checks pass: `typecheck`, `lint`, `check:file-length`, `test:unit` (2546 pass / 0 fail), pre-commit/pre-push hooks.
- Design spec: `specs/2026-06-21-open-claude-design-init-references-live-qa.md`.
- The impeccable skill changes were pre-existing in the working tree and are intentionally bundled in this PR.


flora131:
Thanks for the thorough review — addressed in `cbde5080b`.

**🟠 Medium (the blocker) — early-break export gate.** Fixed deterministically. `refineOpenClaudeDesign` now tracks `unappliedMeaningfulFeedback` (seeded from the initial preview, set when a `preview-display-*` round captures meaningful notes/live-changes, cleared after each apply). The loop only honors an immediate `ready_for_export=true` when that flag is `false`; otherwise it falls through to a forced apply pass — which runs `assertUserAnnotationsThreaded` — so captured annotations can never be dropped at the gate. Added `builtin-workflows-open_claude_design-02.test.ts` with two cases: (1) `ready_for_export=true` while the initial preview holds unaddressed annotations now forces `apply-changes-1` (and threads "I don't like this background" / "Apple website") with no spurious second apply, and (2) the no-annotations fast-path still exports immediately without an apply.

**🟡 Low — guardrail vs `liveChanges`.** `assertUserAnnotationsThreaded` now enforces `liveChanges` threading in addition to `userNotes`, matching `buildUserAnnotationsSection`/`persistPreviewFeedback`. New unit test covers a live-changes-only feedback round (throws when dropped, passes when threaded).

**🔵 Nits**
- Dropped `"a"`/`"n"` from `PLACEHOLDER_TOKENS` so a one-character real note survives (`"na"` stays for `N/A`); added an assertion.
- Fixed the `staff design enginer` prompt typo.

**Deliberately left as-is (flagging the rationale):**
- **`extractField` over-capture (Low).** The `preview-display-*` output labels are a fixed, model-controlled set and all live in `FIELD_LABELS`, so a stray unknown label between `user_notes` and the next field shouldn't occur in practice. Terminating on any `word:`-style line would truncate legitimate multi-line notes (which are frequently bullet lists containing colons), so I left the over-capture-leaning behavior rather than risk under-capture. Happy to tighten it if you'd prefer a stricter contract.
- **`ensurePlaywrightCli` global install / `copyAnnotationArtifacts` / `// @ts-nocheck` tests** — kept as conscious choices per your notes (gated best-effort bootstrap, trusted-local artifact copy, mock-heavy test files).

Full suite green locally: typecheck, lint, `check:file-length`, and `test:unit` (2549 pass / 0 fail).

flora131:
Thanks — addressed in `5b7a06492`.

**🟠 1 — terminal-iteration annotations dropped (residual #1464).** Fixed. `buildLivePreviewDisplayPrompt` gained a `final` mode: on the **last** refinement iteration the preview is shown read-only — it no longer solicits `user_notes`/`live_changes` it can't apply and instead tells the user how to re-run — and the loop skips capture/threading on that iteration, so a terminal annotation is never captured-then-orphaned. (Combined with the export-gate guard from the prior round, captured feedback now always goes through an apply pass or is never solicited.)

**🟠 2 — `final-display` solicited then discarded notes.** Fixed. It no longer runs `show --annotate` or asks for `user_notes`/`annotated_snapshot`; it just opens the spec for review and emits a `next_action_hint` on how to re-run for changes.

**Browser-centric early exit (per the maintainer's note that `ctx.exit()` is available).** Added a deterministic guard right after the startup `ensurePlaywrightCli()` check: when the `playwright-cli` browser can't be made available, the run `ctx.exit()`s up front — surfacing the would-be artifact paths and install instructions — instead of generating a design no one can review interactively. Gated via `shouldEarlyExitForBrowser(available, NODE_ENV)` so the unit suite (`NODE_ENV=test`) and runtimes without `ctx.exit` still run to completion.

**🟡 3 — parallel onboarding prompts implied a pipeline.** Reworded the `ds-analyzer` and `ds-patterns` objectives to state each does its own independent repo scan and explicitly that they run in parallel with the locator, so the "evaluate the located evidence" phrasing no longer implies a dependency.

**🟡 4 — `copyAnnotationArtifacts` copied a model-supplied path verbatim.** Now constrained: the resolved source must be within `workflowCwd` or the run's artifact dir, otherwise it's skipped (the notes `.md`/`.json` are still persisted).

**🔵 Nits.** The `extractField` known-label-termination edge and the `// @ts-nocheck` test files are left as-is (acknowledged tradeoffs). The `runOpenClaudeDesignWorkflow` over-indentation / `const designContext = ctx;` alias is cosmetic and from the earlier rewrite — re-indenting the whole body now would balloon this PR's diff far more than it'd clean up, so I left it; happy to do a dedicated cosmetic pass if you'd prefer.

New tests cover the read-only final prompt, the early-exit predicate, and the snapshot-containment check; docs and both changelogs updated. Full suite green locally: typecheck, lint, `check:file-length`, `test:unit` (2552 pass / 0 fail).

PR #1523, opened 2026-06-26
fix(subagents): prevent async widget remount flicker

### LINKED ISSUE #1517: Enhancement: remove flicker from the subagent UI
## Summary
Remove visible flicker from the subagent UI during updates.

## Problem
The subagent UI appears to flicker as state changes, making the experience feel unstable and harder to visually track. This is especially noticeable when subagent status/output is refreshing or re-rendering frequently.

## Desired behavior
- Subagent rows/panels should update without visible flicker.
- Existing content should remain visually stable while new status/output arrives.
- Loading, streaming, and completion transitions should avoid remount-like flashes when possible.

## Possible areas to inspect
- React keys or component identity changes causing unnecessary remounts.
- Conditional rendering paths that briefly remove/re-add subagent content.
- Streaming/status update frequency causing layout or opacity flashes.
- CSS transitions or skeleton/loading state swaps that create visible flicker.

flora131:
## Summary

Fixes a TUI flicker where the async subagent status widget would briefly blank out during reset-and-hydrate cycles (e.g. on Atomic host updates), even when active background runs were still present. Two independent bugs combined to cause the issue.

## Root Cause

1. **Transient empty frame on reset**: `resetJobs` in `async-job-tracker.ts` was calling `rerenderWidget(ctx, [])`, publishing an empty widget frame *before* `hydrateActiveJobs` could re-populate it with still-running jobs.

2. **Object identity comparison failure**: Widget ownership (`mountedWidgetCtx !== ctx`) was compared by reference. When Atomic delivers a fresh `ExtensionContext` wrapper for the same logical session (e.g. after a host update), the new wrapper is not `===` the stored one, so the guard incorrectly treated it as a different owner and allowed the blank frame to clear the widget.

### `packages/subagents/src/runs/background/async-job-tracker.ts`
- Removed the premature `rerenderWidget(ctx, [])` call from `resetJobs`; now only updates `lastUiContext` and lets the subsequent `hydrateActiveJobs` drive the first render.

### `packages/subagents/src/tui/render-widget.ts`
- Introduced `getWidgetOwnerKey(ctx)` — a stable string key derived from the resolved session file path (falling back to session ID, then `"session:unknown"`) combined with the resolved `cwd`. Two `ExtensionContext` objects with the same session and working directory compare as the same logical owner.
- Replaced all `mountedWidgetCtx !== ctx` (object-identity) guards with `mountedWidgetOwnerKey !== ownerKey` (value-equality) guards.
- Added `mountedWidgetOwnerKey` tracking alongside `mountedWidgetCtx` in the singleton widget state.
- Fixed teardown pointer: visible→visible updates from fresh same-owner wrappers now refresh `mountedWidgetCtx` so teardown always targets the freshest wrapper rather than a stale one.

### Tests
- `test/unit/subagents-render-widget-owner.test.ts` — new file covering: same-owner fresh wrappers must not remount; different-owner wrappers must remount; `ctx.cwd`-relative vs. absolute session files resolve to the same owner; running/status/terminal state transitions do not blank until no jobs remain.
- `test/unit/subagents-async-widget-visibility.test.ts` — two new cases: widget stays visible across `reset → hydrate` when active runs exist; widget correctly unmounts when `hydrate` finds no active jobs after reset.
- `test/unit/subagents-render-stability-widget-lifecycle.ts` — updated fixtures and test descriptions to reflect logical-owner semantics.

### Changelog
- Added `### Fixed` entry under `## [Unreleased]` in `packages/subagents/CHANGELOG.md`.

## Validation

```sh
bun test test/unit/subagents-render-widget-owner.test.ts test/unit/subagents-async-widget-visibility.test.ts
bun test test/unit/subagents-render-stability-widget-lifecycle.ts
bun run test:unit
bun run typecheck
bun run check:file-length
```

Closes #1517


flora131:
# Implementation Notes

Task: Implement GitHub issue #1517: remove visible flicker from the subagent UI during updates. Work only in the provided git worktree. Scope likely packages/subagents UI/rendering, any Atomic TUI integration that displays subagent rows/panels, tests, docs, and changelog as needed. Requirements: investigate subagent UI update/render lifecycle and identify sources of flicker/remount-like flashes during frequent status/output refreshes; preserve stable row/panel identity and layout while loading, streaming, status updates, and completion transitions occur; avoid conditional remove/re-add paths or key churn that briefly drops existing content; throttle/debounce or stabilize update paths only if it does not hide important real-time state; preserve existing main chat and workflow chat subagent behavior. Add targeted bun:test coverage or rendering regression tests where practical to verify stable component identity/output across running→streaming→terminal transitions and no transient blank/missing subagent rows during updates. Update relevant docs/changelog if behavior is user-visible. Validate with focused tests plus typecheck/lint or closest practical checks. When complete, create a PR for the changes.

## Running Notes

- Record implementation decisions, deviations from the research findings, tradeoffs, blockers, validation notes, and anything else the user should know.

## Setup preflight (2026-06-25)
- Used `bun install` (not `bun install --frozen-lockfile`) because project docs/package metadata prefer Bun and no CI-only frozen-lockfile constraint was required for this local preflight; existing `bun.lock` remained unmodified.
- Bun toolchain matched the repo requirement: `bun --version` reported `1.3.14`; TypeScript resolved from installed dependencies as `6.0.3`.
- Verified setup with `bun run typecheck` and `bun run test:unit`; both completed successfully after dependency installation.

## Validation Review (2026-06-26)

- Reviewed the current git diff for issue #1517. The implementation narrows the reset path in `packages/subagents/src/runs/background/async-job-tracker.ts` so `resetJobs(ctx)` clears internal tracker state but no longer immediately publishes an empty async widget frame. `hydrateActiveJobs(ctx)` remains responsible for re-rendering after it has reloaded active queued/running jobs, and still unmounts the widget when hydration finds no active jobs. This matches the issue requirement to avoid a reset-and-hydrate blank/remount flash while preserving legitimate empty-state unmount behavior.
- Reviewed the added unit coverage in `test/unit/subagents-async-widget-visibility.test.ts`. The new cases cover both sides of the behavior: active reset+hydrate does not call `setWidget(..., undefined)` and reuses the mounted widget with one in-place render, while reset+hydrate with no active runs still emits a single unmount during hydration. The assertions are aligned with the existing widget lifecycle test helpers.
- Reviewed `packages/subagents/CHANGELOG.md`; the `[Unreleased]` fixed entry accurately describes the user-visible flicker fix.

Validation commands run from `/Users/tonystark/Documents/projects/atomic-1517-subagent-ui-flicker`:

- `bun test test/unit/subagents-async-widget-visibility.test.ts` — passed, 7 tests.
- `bun run typecheck` — passed (`tsc --noEmit`).
- `bun run check:file-length` — passed (1837 tracked files checked; max 500).
- `bun run test:unit` — passed, 2725 tests across 291 files.

## TUI / tmux E2E Feasibility

- Confirmed tmux is available (`tmux 3.6b`).
- Performed an isolated terminal smoke scenario using tmux rather than browser automation:
  - Seeded a synthetic active async subagent `status.json` under the extension's default async run root for the current worktree cwd.
  - Launched the local Atomic TUI with `ATOMIC_OFFLINE=1 bun packages/coding-agent/src/cli.ts --approve --offline` inside a detached tmux session.
  - Captured the terminal pane and confirmed the below-editor async subagent widget hydrated on startup, showing the active `worker` background run and updating spinner frames in place.
  - Killed the tmux session and removed the synthetic async run directory after capture.
- This smoke test exercises the app/session-start reset+hydrate path that issue #1517 targets. It is still a terminal capture/manual inspection rather than deterministic frame-by-frame flicker measurement; the focused unit tests provide the deterministic assertion that no blank widget frame is published during active reset+hydrate.
- No blocking prerequisite remained for a basic TUI smoke test. A stronger automated visual E2E would require a purpose-built terminal-frame recorder/assertion harness for the TUI, which this repo does not currently expose as a browser-playback target.

## QA E2E Video

- No Playwright/browser `.webm` was produced because this issue is for a terminal UI widget, not a browser UI. Fabricating a browser video would not validate the affected surface.
- The applicable E2E evidence is the tmux terminal smoke capture described above plus the deterministic Bun unit tests verifying widget mount/unmount calls.

## Implementation Decisions (2026-06-26)

- Implemented the fix at the async job tracker reset/hydrate boundary instead of in the low-level widget renderer. `renderWidget(ctx, [])` still represents a true empty state and can unmount the widget; `resetJobs(ctx)` no longer publishes that transient empty state before hydration recomputes active runs.
- Preserved `state.lastUiContext = ctx` during reset so follow-up hydration/render paths retain the current UI context without forcing a blank below-editor widget frame.
- Added regression coverage for both active and empty reset+hydrate paths: active jobs stay mounted with no `setWidget(..., undefined)` call, while no active jobs still unmount during hydration.
- Updated the subagents changelog. No user-facing docs were changed because this is a behavior-quality fix with no new command, option, or documented workflow.
- No PR was created in this stage per orchestration instruction to ignore PR submission requests.

## Issue #1517 render-widget owner stability implementation

- Replaced render-widget ownership comparisons based on `ExtensionContext` object identity with a stable logical owner key: `sessionManager.getSessionFile()` when present, otherwise `getSessionId()`, plus `path.resolve(ctx.cwd)`.
- Kept `mountedWidgetCtx` as the actual context that received `ui.setWidget(...)`, so later unmounts still target the mounted UI even when same-owner updates arrive through fresh wrapper objects.
- Same-owner active updates now only refresh the singleton widget data and request an in-place render; different owner/session updates still best-effort clear the previous mount before mounting the new one.
- Added focused `bun:test` coverage for same-owner fresh wrappers, same-owner no-active unmount, different-owner remount, and running/status/terminal updates avoiding transient blank frames before the no-active transition.
- Also updated the existing hydration regression test to use separate fresh context wrappers for initial hydrate, reset, and rehydrate, matching the Atomic event-wrapper behavior that triggered #1517.

## Issue #1517 owner stability refinement (2026-06-26)

- Normalized render-widget session-file ownership with `path.resolve(sessionFile)` while preserving the existing resolved cwd component. This keeps relative and absolute wrappers for the same persisted session from being treated as different logical widget owners.
- Updated the no-active reset+hydrate regression to use fresh same-owner UI context wrappers for the initial hydrate, reset, and follow-up hydrate, matching the wrapper churn observed in real Atomic event paths.
- Added focused owner coverage proving an equivalent relative session file followed by an absolute session file updates in place rather than clearing/remounting the widget.
- Validation: `bun test test/unit/subagents-render-widget-owner.test.ts test/unit/subagents-async-widget-visibility.test.ts` passed (12 tests).

## Final Validation After Stable Owner Implementation (2026-06-26)

- Reviewed the current git diff against the updated issue #1517 research and requirements:
  - `packages/subagents/src/tui/render-widget.ts` no longer treats fresh `ExtensionContext` wrapper object identity as ownership. It uses a stable logical owner key derived from normalized `sessionManager.getSessionFile()` when present, otherwise `getSessionId()`, plus normalized `cwd`.
  - Same logical owner with fresh wrapper objects updates the singleton widget data and requests an in-place render without clearing/remounting the below-editor widget.
  - Different logical owners still best-effort unmount the previous mounted UI context before mounting the widget on the new owner/context.
  - Empty updates from stale/different owners are ignored so they cannot clear the active owner's widget; same-owner empty updates still unmount the actual mounted UI context.
  - `resetJobs(ctx)` keeps the latest UI context but no longer publishes `rerenderWidget(ctx, [])`, so reset+hydrate no longer creates a transient empty widget frame before active jobs are rehydrated.
- During full-suite validation, the pre-existing lifecycle regression `visible async widget remounts when the UI context changes` failed because its old assertion encoded object-identity context switching. I updated that test to assert the approved behavior instead: remount occurs when the logical owner changes. This keeps the broader lifecycle suite aligned with the stable-owner implementation.

Validation commands run from `/Users/tonystark/Documents/projects/atomic-1517-subagent-ui-flicker`:

- `bun test test/unit/subagents-render-widget-owner.test.ts test/unit/subagents-async-widget-visibility.test.ts` — passed, 12 tests.
- `bun test test/unit/subagents-render-stability.test.ts` — passed, 25 tests, including the adjusted logical-owner lifecycle regression.
- `bun run typecheck` — passed (`tsc --noEmit`).
- `bun run check:file-length` — passed (1837 tracked files checked; max 500).
- `bun run test:unit` — passed, 2730 tests across 292 files.

TUI / tmux E2E status:

- Reran a focused terminal TUI smoke because the final owner change touched the low-level render-widget ownership path after the earlier smoke.
- Used tmux 3.6b with a synthetic active async subagent `status.json` under the default `ASYNC_DIR` for this worktree cwd, then launched `ATOMIC_OFFLINE=1 bun packages/coding-agent/src/cli.ts --approve --offline` in a detached 120x40 terminal.
- Captured the pane and confirmed the below-editor widget hydrated and displayed the active `worker` background run (`async subagent worker · background`, running step row, output path). Cleaned up the synthetic run directory and killed the tmux session afterward.
- This remains a terminal smoke/manual capture, not a deterministic frame-by-frame visual diff. The deterministic guarantees for no transient blank/remount are covered by the Bun unit tests above.

QA E2E video applicability:

- No Playwright/browser video was produced. Issue #1517 affects a terminal TUI surface, not a browser UI, so a browser `.webm` would not validate the affected behavior.
- Applicable QA evidence is the tmux terminal capture/smoke plus the deterministic widget lifecycle and hydration unit tests.

Issue #1527, opened 2026-06-26
GitHub Copilot Claude can reject replay after repeated subagent runs due to orphaned tool_result


flora131:
## Summary

A production user on the latest stable Atomic release (`@bastani/atomic@0.9.2`) reports that GitHub Copilot Claude models fail after running a subagent job for the second time. The first subagent run succeeds; the second run completes; the next parent-chat continuation fails for all GitHub/Copilot models with an Anthropic tool-result pairing error. Switching to Fireworks GLM 5.2 allows the same chat to continue.

Reported error:

```text
Error: 400 {"type":"error","error":{"type":"invalid_request_error","message":"messages.30.content.0: unexpected `tool_use_id` found in `tool_result` blocks:\ntoolu_01YFHGCXwGqDkeyASQeSKn7Q. Each `tool_result` block must have a corresponding `tool_use` block in the previous message."},"request_id":"req_011CcS7dYJvbJoy4XmCECT6f"}
```

## Production baseline investigated

- Latest stable production release: `@bastani/atomic@0.9.2`
- Git tag: `0.9.2`
- Commit: `749d03393e8430014bec84cc2e68c9d9a5de8e2a`
- `packages/coding-agent/package.json` at that tag pins:
  - `@earendil-works/pi-ai`: `^0.79.10`
  - `@earendil-works/pi-agent-core`: `^0.79.10`

The current checkout is a post-prod prerelease, so the investigation intentionally referenced the stable `0.9.2` code paths.

## Current hypothesis / likely root cause

The active replayed session context can contain a retained `toolResult` whose matching assistant `toolCall` / Anthropic `tool_use` was removed from the immediately preceding assistant message.

Likely bad replay shape:

```text
assistant: text only              # original toolCall block was deleted/filtered
toolResult(toolCallId = A)        # retained; now orphaned
```

or:

```text
user
toolResult(toolCallId = A)        # assistant entry with toolCall A was deleted/filtered
```

For GitHub Copilot Claude models (`api: "anthropic-messages"`), this serializes to Anthropic `tool_result` blocks. Anthropic requires every `tool_result.tool_use_id` to correspond to a `tool_use.id` in the immediately previous assistant message, so the provider rejects the request.

This explains why Fireworks GLM can continue: that model path is more OpenAI-compatible/tolerant and does not enforce Anthropic's strict adjacency invariant in the same way.

## Why the second subagent run is a plausible trigger

Subagent invocations are represented in the parent session as ordinary tool-call turns:

[...]

PR #1528, opened 2026-06-26
fix(session): harden tool-call/result replay integrity and subagent dispatch

flora131:
## Summary

Refs #1527. Prevents orphaned `toolResult` messages from reaching Anthropic/GitHub Copilot providers after context compaction, and hardens subagent dispatch to reject duplicate concurrent calls. The fix spans session-replay reconciliation, LLM serialization guarding, and upstream subagent sync.

### Session replay repair (`packages/coding-agent`)

- **New** `session-manager-tool-dependencies.ts` — reconciles persisted `context_compaction` deletion filters during active-context rebuild so deleting one side of a tool-call/tool-result pair always deletes or preserves the paired side consistently; converges via a fixpoint loop bounded by `path.length * 2` passes.
- **New** `repairOrphanToolResults` in `messages.ts` — final LLM serialization guard that drops any `toolResult` message whose matching assistant `toolCall` is no longer the immediately preceding tool-use group, preventing provider replay failures.
- **Wired** `reconcilePersistedToolDependencyFilters` into `buildEffectiveContextDeletionFilters` in `session-manager-history.ts`.
- **Docs** `compaction.md` updated to document the tool-call/tool-result pair replay invariant and the two-layer guard strategy.

### Upstream subagent hardening (`packages/subagents`)

- **Duplicate dispatch guard** `subagent-executor.ts` wraps `execute` with `executeWithSingleDispatchGuard` — rejects concurrent subagent dispatches within the same turn with a clear error; `fanout-child.ts` and `index.ts` initialize the new `subagentInProgress` flag in safe state.
- **Fanout child history** `subagent-prompt-runtime.ts` preserves nested subagent call/result history for fanout children (`SUBAGENT_FANOUT_CHILD_ENV=1`) and adds `subagent-slash-text-result` to the parent-only strip set.
- **Failed-run diagnostics** `subagent-executor-single.ts` formats failed foreground run output with the error message, captured child output, and output artifact path for easier debugging.
- **Compact tool-call summaries** `prompt-template-bridge.ts` embeds `toolCall` content blocks from delegated run summaries into delegation messages as readable text; fixes a null-safety gap when `agent` or `task` is missing from a prompt template request.
- **Schema simplification** `schemas.ts` removes provider-hostile `allOf` conditionals from the chain step schema.

### Regression coverage

- New `test/unit/subagents-upstream-sync.test.ts`: duplicate dispatch rejection, fanout child history preservation, compact tool-call summary embedding, failed-run output formatting, schema keyword audit, and malformed fanout shape error.
- Extended `test/session-manager/build-context-02.suite.ts`: stale compaction split-pair replay cases (block-level and entry-level deletion).
- New `packages/coding-agent/test/messages.test.ts`: orphan `toolResult` repair in `convertToLlm`.

## Validation

```sh
bun test test/unit/subagents-upstream-sync.test.ts
bunx --bun --no-install vitest --run test/session-manager/build-context-01.suite.ts test/session-manager/build-context-02.suite.ts test/messages.test.ts
bun run typecheck
bun run check:file-length
git diff --check
```

Pre-commit/pre-push hooks: `bun run lint`, `bun run check:file-length`, `bun run test:unit`

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixed
 
 - Fixed custom tool renderer disposal to honor renderer-owned cleanup callbacks, preventing stale animation registry entries after terminal workflow tool rows are finalized ([#1518](https://github.com/bastani-inc/atomic/issues/1518)).
+- Hardened session replay and LLM conversion so persisted context-compaction filters cannot leave orphaned `toolResult` messages after deleting their paired assistant `toolCall`, preventing GitHub Copilot Claude/Anthropic replay failures after repeated subagent runs ([#1527](https://github.com/bastani-inc/atomic/issues/1527)).
 
 ## [0.9.3-alpha.1] - 2026-06-25
 
--- a/packages/coding-agent/docs/compaction.md
+++ b/packages/coding-agent/docs/compaction.md
@@ -60,6 +60,8 @@ Verbatim Compaction never asks a model to rewrite the conversation for the main 
 
 Replay-sensitive assistant messages that contain `thinking` or `redacted_thinking` blocks are all-or-nothing: Atomic may delete an old thinking-bearing assistant entry when dependency validation allows it, but it will not delete individual sibling blocks from a retained thinking-bearing assistant message. This keeps Anthropic/GitHub Copilot extended-thinking replay byte-for-byte compatible with provider requirements.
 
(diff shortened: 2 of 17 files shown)


flora131:
Addressed in `1d31f7384`:

- Changed compact delegated tool-call summaries to be flattened into text (`Tool calls:\n- ...`) instead of synthetic `toolCall` blocks, so prompt-template responses cannot introduce unpaired tool-use history.
- Moved summary processing after the existing `result.messages` passthrough early return.
- Added a regression assertion that the rebuilt prompt-template response survives `convertToLlm` without tool-call parts.
- Added a clear runtime-error test for malformed dynamic fanout shapes now that provider-hostile schema conditionals are removed.
- Documented that `reconcilePersistedToolDependencyFilters` mutates fresh filters in place and that its bounded loop is only a non-termination backstop.

Validation reran locally and via hooks: targeted Bun/Vitest tests, `bun run typecheck`, `bun run check:file-length`, and pre-push `bun run test:unit`.

PR #1581, opened 2026-07-01
fix(subagents,workflows): add attempt watchdog and request-incompatible fallback classification

### LINKED ISSUE #1580: Subagent model attempt hangs indefinitely (2.5h, 0 turns) when a fallback model request stalls — no per-attempt timeout/idle watchdog
## Summary
A parallel `subagent` run (3× builtin `codebase-analyzer`) stalled for **2h34m** at `0/3 done` and had to be manually aborted. Root cause: the child model-attempt loop in `subagent-runner-step.ts` `await`s `runPiStreaming(...)` with **no execution timeout and no idle/no-progress watchdog**. When a fallback model request hung (zero turns, zero tokens, zero output), the `await` never resolved, blocking the child — and the entire parallel batch — until the user aborted.

## Environment
- `@bastani/atomic` **0.9.3**
- Builtin agent: `codebase-analyzer`; effective (available-filtered) model chain: `zai/glm-5.2:low` → `zai-coding-cn/glm-5.2:low`
- Invocation: `subagent` tool, 3 parallel tasks, concurrency 3

## Evidence (subagent-artifacts `*_meta.json`, all 3 children identical)
```
model:           zai-coding-cn/glm-5.2:low
attemptedModels: [ "zai/glm-5.2:low", "zai-coding-cn/glm-5.2:low" ]
modelAttempts:
  1) zai/glm-5.2:low          success:false exit:1  error:"No API key found for zai."
  2) zai-coding-cn/glm-5.2:low success:false exit:1  error:""          <- hung
durationMs:      9240948            (~2h34m)
usage:           { input:0, output:0, cost:0, turns:0 }
toolCount:       0
exitCode:        1
```
- `*_input.md` written at 14:57; `*_meta.json` written at 17:31 (Δ ≈ 2h34m) — nothing happened between start and manual abort.
- `*_output.md` files are **0 bytes**.

## Mechanism
1. Candidate #1 `zai/glm-5.2:low` fails instantly with `No API key found for zai` → classified retryable (`/api key/i`, `isRetryableModelFailure`) → advance to next candidate. (reasonable)
2. Candidate #2 `zai-coding-cn/glm-5.2:low` is launched via `runPiStreaming(...)` and **hangs** — no turns, no tokens, no output.
3. `subagent-runner-step.ts` `await runPiStreaming(...)` has **no timeout**, so it blocks forever.
4. All three parallel children block identically; the batch stays `0/3 done` for 2h34m until manual abort, at which point exit 1 is recorded and meta is written.

## Root cause (code)
`dist/builtin/subagents/src/runs/background/subagent-runner-step.ts` (~L125):
```ts
const run = await runPiStreaming(args, step.cwd ?? ctx.cwd, ctx.outputFile, env, ...);
// ^ no AbortController, no run-duration timeout, no idle/no-progress watchdog
```
`runPiStreaming` (`subagent-runner-streaming.ts`) only has a **post-exit** stdio-drain guard (`attachPostExitStdioGuard({ idleMs: 2000, hardMs: 8000 })`) and **interrupt-triggered** kill timers via `registerInterrupt`. During a normal (non-interrupted) run there is **no wall-clock timeout and no idle watchdog**, so a model/provider request that never returns (or a stream that never yields) blocks the attempt — and the whole run — indefinitely. The foreground path (`runs/foreground/execution-run-sync.ts`) has the same shape.

## Contributing factors
1. **Unavailable primary provider in the shipped default chain.** `agents/codebase-analyzer.md` is `model: zai/glm-5.2:low`, `fallbackModels: zai-coding-cn/glm-5.2:low, openai-codex/gpt-5.5:low, github-copilot/gpt-5.5:low, ...`. Here `zai` has no API key, so attempt #1 is guaranteed to fail and the whole run rides on the single hanging `zai-coding-cn` fallback. (worker/debugger/code-simplifier/codebase-* all ship the same zai-led chain.)
2. **No idle watchdog / no-progress abort.** `turns:0 / toolCount:0 / usage:0` for 2.5h never triggers cancellation. The control/attention system emits *notifications* (needsAttention / activeLongRunning) but does not *cancel*.
3. **Parallel batch has no per-child or total timeout**, so one/all hung children leave the run at `0/N done` with no upper bound.

## Expected behavior
- Each model attempt has (a) an **idle/no-progress watchdog** (no child events/turns/tokens within N seconds → kill child, mark attempt failed) and (b) an **absolute per-attempt wall-clock cap**.
- A hung/timed-out attempt is treated as **retryable**, so remaining fallbacks (openai-codex, github-copilot, anthropic, google, openrouter, …) are actually tried instead of the run dying on a stall.
- Parallel batches bound total wait and surface a clear timeout error instead of `0/N done` forever.

## Suggested fixes
- Wrap `runPiStreaming` in `subagent-runner-step.ts` (and `execution-run-sync.ts`) with an `AbortController` driven by **idle timeout** (reset on each streamed child event) **+ hard max duration**; on trip, kill the child and push a synthetic retryable `ModelAttempt` so the fallback loop advances.
- Plumb a **request/stream timeout** down to the provider client so a hung HTTP/stream can't block forever.
- (Config) Resolve/reorder builtin `model`/`fallbackModels` to prefer the user's currently-authenticated provider, and drop candidates whose provider has no API key **before** spawning, so a missing-key primary doesn't leave the run dependent on a single fallback.

## Repro sketch
1. Configure a subagent whose primary model provider has no API key, and whose next fallback points at a provider/endpoint that accepts the connection but never responds (blackhole/throttle it).
2. Run `subagent` (single or parallel).
3. Observe the run hang indefinitely with `turns:0`, requiring manual abort; artifact `durationMs` grows unbounded and `*_output.md` stays 0 bytes.

flora131:
## Summary

Fixes the #1580 stall end-to-end and hardens model fallback chains in subagents and workflows.

Issue #1580 reported a parallel subagent run stuck at `0/3 done` for 2h34m: the first candidate failed fast (missing API key, already retryable), but the second candidate's request hung with 0 turns/0 tokens, and `await runPiStreaming(...)` had no idle watchdog or per-attempt timeout — so the attempt (and the whole batch) blocked until manual abort. Separately, request/context-incompatibility failures (HTTP 400/413/422, context-length overflow, unsupported tool/parameter) were classified non-retryable, so a chain that did fail with one of those never reached the current user-selected model as the last resort.

Closes #1580

### Per-attempt idle watchdog + wall-clock cap (the #1580 hang fix)

- New `packages/subagents/src/runs/shared/attempt-watchdog.ts`: every model attempt now runs under an **idle/no-progress watchdog** (default 5 min, reset on every streamed child event/stdout/stderr activity) and an **absolute wall-clock cap** (default 60 min). On trip, the child is killed (SIGTERM → SIGKILL after a 3s grace) and the attempt resolves with a synthetic `"...timed out..."` failure that the existing classifier treats as **retryable**, so the fallback loop advances to the next candidate instead of blocking forever.
- An in-flight tool execution counts as activity, so a slow, quiet tool call (a long build or test run that streams nothing until it finishes) does not falsely trip the idle window — only the wall-clock cap bounds such attempts. Wired into both the foreground and background spawn paths.
- Wired into both the background path (`subagent-runner-step.ts` / `subagent-runner-streaming.ts`) and the foreground path (`execution-run-sync.ts` / `execution-attempt.ts`), including chain/parallel executors.
- Overridable via `ATOMIC_SUBAGENT_ATTEMPT_IDLE_TIMEOUT_MS`, `ATOMIC_SUBAGENT_ATTEMPT_TIMEOUT_MS`, and `ATOMIC_SUBAGENT_ATTEMPT_KILL_GRACE_MS`. Setting the idle or wall-clock override to `0` (or a negative value) now disables that timeout entirely; non-numeric values remain ignored and fall back to the default.

### Pre-spawn candidate filtering

- New `packages/subagents/src/runs/shared/model-candidate-filter.ts`: candidates whose provider is **known** and affirmatively has **no configured API key/auth** are skipped before spawning a child, recorded as skipped `modelAttempts` so the decision is visible in artifacts. Unknown/custom providers are still attempted (they already fail fast), and the current user-selected model appended as last resort is never filtered out.
- Extracted `collectKnownModelProviders()` in `packages/subagents/src/shared/model-info.ts` to replace four duplicated model-registry provider-derivation expressions.
- The background runner (`subagent-runner-step.ts`) now mirrors the foreground's empty-candidates handling: an empty `modelCandidates` array with no pre-spawn skipped attempts means no candidates were ever configured (no primary, no fallbacks, no current model), so it spawns one default-model attempt instead of silently exiting with no attempt and no error. A list that was filtered down to empty (which always carries skipped attempts) is still respected and surfaced as an error.

### `request_incompatible` fallback classification

- Adds a `request_incompatible` `ModelFallbackFailureKind` to both `packages/workflows/src/runs/shared/model-fallback-failures.ts` and `packages/subagents/src/runs/shared/model-fallback.ts`, included in each package's `FALLBACKABLE_FAILURE_KINDS`.
- Classifies HTTP 400/413/422, request-too-large, context-window/context-length overflow, unsupported tool/parameter, and `invalid_request`/`bad_request`/`too_large`-style codes and messages as `request_incompatible` via status codes, error codes, and message patterns.
- Preserves precedence for non-retryable signals — refusals, content-filter/safety blocks, cancellations, and task failures still stop the chain.
- Direct-message classification in both the subagents and workflows classifiers now runs after nested cause/diagnostic traversal (and before outer status/code) so a generic wrapper message (e.g. "400 bad request") can't mask a non-retryable nested signal — the two classifier copies are aligned on this precedence and a new cross-package conformance test (`model-fallback-classifier-conformance.test.ts`) runs a shared failure corpus through both to guard against future silent drift.

## Key files

- `packages/subagents/src/runs/shared/attempt-watchdog.ts` (new)
- `packages/subagents/src/runs/shared/model-candidate-filter.ts` (new)
- `packages/subagents/src/runs/shared/model-fallback.ts`
- `packages/workflows/src/runs/shared/model-fallback-failures.ts`
- `packages/subagents/src/runs/background/{subagent-runner-step,subagent-runner-streaming}.ts`
- `packages/subagents/src/runs/foreground/{execution-run-sync,execution-attempt,chain-execution*}.ts`
- `packages/subagents/src/shared/model-info.ts`

## Tests, docs, changelogs

- `test/unit/subagents-attempt-watchdog.test.ts`: watchdog trips on a stalled child and the fallback loop advances with a synthetic retryable failure; idle timer resets on child activity and on in-flight tool execution (no false kills, foreground and background); wall cap trips independently of tool activity; SIGTERM→SIGKILL escalation; pre-spawn filter skips known keyless providers but never unknown providers or the current-model last resort; background default-attempt regression when no candidates are configured; idle/wall timeout disabled via `0`/negative env override; non-numeric env values ignored.
- `test/unit/model-fallback-request-incompatible.test.ts`, `test/unit/model-fallback-classifier-conformance.test.ts`, `test/unit/subagents-model-fallback.test.ts`, `test/unit/stage-runner-fallback-resume.test.ts`: classifier behavior, subagents/workflows classifier parity, and an end-to-end workflow fallback loop reaching the current user-selected model.
- Docs: `packages/coding-agent/docs/subagents.md` (watchdog/timeouts/tool-activity deferral/pre-spawn filtering/env disable escape hatch) and `packages/coding-agent/docs/workflows.md` (request-incompatible fallback).
- Changelog entries under `[Unreleased]` in `packages/coding-agent`, `packages/subagents`, and `packages/workflows`.

## Validation

- `bun test test/unit/subagents-attempt-watchdog.test.ts test/unit/subagents-model-fallback.test.ts test/unit/model-fallback-request-incompatible.test.ts test/unit/model-fallback-classifier-conformance.test.ts test/unit/stage-runner-fallback-resume.test.ts` — all pass
- `bun run test:unit` — green (run in the pre-push hook)
- `bun run typecheck` / `bun run lint` — pass
- `bun run check:file-length` — pass (watchdog and filter extracted into new ≤100-line modules; no file pushed over the 500-line gate)

## QA E2E video

No QA E2E video applies; this is backend runner/classifier behavior with no browser/UI scenario. The watchdog and fallback-loop regression tests provide executable proof.


flora131:
# Implementation Notes

Task: Investigate and fix GitHub issue https://github.com/bastani-inc/atomic/issues/1580 in the Atomic monorepo. Work in a separate git worktree named ../atomic-issue-1580-model-fallbacks (create it from origin/main if needed). Determine why the rest of the model fallback candidates did not get tried. Implement the required behavior: if none of the configured fallback candidates are applicable/usable for the current request/context, fall back to the currently selected user model rather than stopping without an applicable fallback. Preserve existing fallback behavior when candidates are applicable. Add or update focused tests that reproduce the issue and verify the selected user model fallback path. Update relevant user-facing docs in packages/coding-agent/docs and package changelog(s) under ## [Unreleased] if behavior changes. Use Bun commands only (bun test, bun run typecheck/lint as appropriate). Create a pull request when complete with a clear summary and validation evidence.

## Running Notes

- Record implementation decisions, deviations from the research findings, tradeoffs, blockers, validation notes, and anything else the user should know.

## Implementation Decisions

- Preflight found the worktree was a Bun workspace checkout with missing dependencies; `bun install` was run successfully before implementation.
- Root cause confirmed by subagent analysis: the workflow and subagent fallback loops already append/advance through candidate models, including the current selected user model, but the retryability classifiers marked request/context incompatibility failures as `unknown`/non-retryable. That stopped the chain before later configured candidates or the selected user model.
- Implemented a dedicated `request_incompatible` fallback kind in both workflow and subagent classifiers instead of overloading provider/transport failures. This keeps audit metadata specific and preserves existing behavior for other fallback causes.
- Classification is intentionally narrow: HTTP 400/422 plus curated request/context-incompatibility codes/messages (invalid/bad request, context length/window overflow, request too large, unsupported/unknown/invalid tool/parameter/function). Other 4xx statuses are not made broadly retryable.
- Refusal, content-filter/safety, cancellation, and task-failure precedence remains non-retryable and continues to win over request-incompatible signals.
- The implementation subagent refactored the subagents `kindFromCode` switch into a set/table lookup to satisfy the repository's 500-line source-file gate. This was a small behavior-preserving refactor but is a notable deviation from the minimal research guidance.
- The workflows fallback classifier is now close to the file-length gate (validation reported 499 lines); future additions should consider splitting it.

## Continuation Update — Reviewer Precedence Fix

- Latest research identified an unresolved reviewer finding: the workflows classifier could classify generic wrapper messages such as `invalid request` / `400 bad request` as retryable before inspecting nested diagnostics or causes for non-retryable content-filter, safety, AbortError/cancellation, or task-failure signals.
- Implemented the minimal precedence fix in `packages/workflows/src/runs/shared/model-fallback-failures.ts`: direct-message fallback classification now runs after diagnostics and cause traversal, while direct refusal/cancellation checks remain ahead of fallbackable request-incompatible classification.
- Added focused workflow classifier regression tests proving nested non-retryable causes/diagnostics win over retryable wrapper messages and genuine request-incompatible wrappers still classify as fallbackable.
- No subagent classifier change was needed for this continuation because the reviewer finding was specific to the workflows classifier ordering.
- The workflows classifier is now exactly at the 500-line gate; validation passes, but any future additions should split the file rather than adding more lines.

## Continuation Update — HTTP 413 Payload Too Large

- Latest research identified another unresolved reviewer finding: both classifiers mapped HTTP 400/422 to `request_incompatible` but omitted bare HTTP 413 Payload Too Large.
- Implemented HTTP 413 status classification in both `packages/workflows/src/runs/shared/model-fallback-failures.ts` and `packages/subagents/src/runs/shared/model-fallback.ts`. This covers `{ status: 413 }`, `{ statusCode: 413 }`, `{ httpStatus: 413 }`, `{ code: 413 }`, and `{ code: "413" }` because status extraction and code-to-status classification already route through `kindFromStatus`.
- Added focused 413 tests in workflow and subagent fallback classifier test coverage. Docs and changelog wording now mention HTTP 400/413/422 and payload/request-too-large conditions.
- Scope remains narrow: only 400/413/422 are classified as request-incompatible; other 4xx statuses are not broadly retryable, and refusal/cancellation/content-filter/task-failure precedence is unchanged.
- The workflows classifier remains exactly at the 500-line gate, so future changes should split/refactor before adding more cases.

## Validation Outcomes

- Focused tests passed after the final HTTP 413 update: `bun test test/unit/model-fallback-01.test.ts test/unit/model-fallback-request-incompatible.test.ts test/unit/subagents-model-fallback.test.ts test/unit/stage-runner-fallback-resume.test.ts` (55 pass, 0 fail).
- Type checking passed after the final HTTP 413 update: `bun run typecheck`.
- Lint passed after the final HTTP 413 update: `bun run lint`.
- File-length gate passed after the final HTTP 413 update: `bun run check:file-length`.
- Final validation confirmed modified implementation, test, docs, and changelog files plus a new focused workflow classifier test file.

## Blockers / Deferred

- No functional blockers remain.
- No pull request was created in this stage because the orchestrator role instruction explicitly said to ignore PR submission requests.
- Untracked orchestration artifacts remain in the worktree (`subagent-*.md`, `progress.md`, and existing research artifacts); validation recommended staging only implementation/test/docs/changelog files in a later PR stage.

## QA E2E Video

- No QA E2E video applies: the change is backend/classifier model-fallback behavior with no browser/UI scenario. The executable behavior proof is the workflow fallback loop regression in `test/unit/stage-runner-fallback-resume.test.ts`.

flora131:
Re: the review's scope concern — resolved by expanding this branch to actually fix the hang rather than downgrading to `Refs`. Two follow-up commits (`eb91530e0`, `21f034ecd`) add a per-attempt idle watchdog (default 5 min, reset on any child activity) plus an absolute wall-clock cap (default 60 min) around `runPiStreaming` in both background and foreground paths; a tripped attempt is killed and recorded as a synthetic retryable failure so the fallback loop advances. Candidates whose provider affirmatively has no configured key are now skipped pre-spawn (recorded as skipped attempts; unknown/custom providers and the current-model last resort are never filtered). `Closes #1580` now reflects the actual root-cause fix. The 500-line headroom note was also addressed by extracting the new logic into dedicated modules.

flora131:
Addressed the review feedback in 776c4b067:

**🔴 Foreground/background asymmetry on empty candidates** — fixed in `subagent-runner-step.ts`. The runner now distinguishes the two empty cases using an invariant of `filterSpawnableModelCandidates`: every filtered-out candidate is always recorded as a skipped attempt in `step.modelAttempts`. So an empty `modelCandidates` array **with** skipped attempts means "filtered to empty" (respected; surfaced as the `No spawnable...` error), while an empty array **without** skipped attempts means "no candidates were ever configured" and now spawns one default-model attempt, mirroring the foreground `modelsToTry` distinction. Regression test added (`background runner spawns one default attempt when no candidates were ever configured`).

**🟡 `positiveEnvMs` can't disable the watchdog** — `ATOMIC_SUBAGENT_ATTEMPT_IDLE_TIMEOUT_MS`/`ATOMIC_SUBAGENT_ATTEMPT_TIMEOUT_MS` set to `0` (or negative) now disable the corresponding timeout entirely; non-numeric values remain ignored (default applies) and this is now documented in `subagents.md`. `ATOMIC_SUBAGENT_ATTEMPT_KILL_GRACE_MS` keeps positive-only semantics. Tests cover idle-disabled-via-`0` (silent child survives idle window, bounded only by wall cap) and non-numeric/negative env resolution.

**🟡 Quiet reasoning turns / duplicated classifier tables** — acknowledged; no code change. The idle window stays overridable (and now fully disableable) for providers with long silent generations, and the cross-package conformance test remains the guard for the classifier copies.

`bun run test:unit` (2822 pass), `typecheck`, and `check:file-length` are green.

flora131:
Addressed the second-pass feedback in 723d21b6f:

**1. Cost tradeoff of `request_incompatible` walking the chain** — confirmed as intended per #1580: the explicit goal is that when no configured candidate can serve the request, the run reaches the user's currently selected model as last resort instead of failing outright. The chain is typically short (primary + a few fallbacks + current model), each incompatible attempt fails fast at request time (no long generation), and refusals/cancellations/task failures still stop immediately. The context-overflow-specific idea (truncate/compact rather than retrying the oversized payload on a possibly smaller-window candidate) is a good follow-up but out of scope for this fix — retrying a larger-window fallback is exactly the recovery #1580 asks for; a smaller-window candidate just fails fast again and the chain proceeds.

**2. Breadth of the bare `too large` pattern** — kept as-is (structured status/code and refusal/cancel/task-failure checks already run first, per the current precedence), but the residual fragility is now documented and pinned: added a conformance fixture (`benign message containing too large`) that documents the known false positive in both classifier copies, with a comment requiring lockstep updates if either copy tightens the pattern.

**3. Two classifier copies** — extracting a shared module is blocked by the package split (the copies must not import each other), so the conformance suite header now explicitly states its limitation: it enforces *behavioral* parity over a finite corpus only, notes the copies are structurally different, and instructs that any rule change must update both copies plus add a corpus fixture.

**4. `isToolActive` vs. abnormal tool end** — added the requested callsite comments in both `execution-attempt.ts` and `subagent-runner-streaming.ts` (idle deferred indefinitely if a tool end event is lost; the wall cap is the sole backstop — do not lower it assuming idle fires), plus a new test proving the wall cap terminates a tool-start-without-end stall and fallback advances.

**5. `error ??= message` ordering** — acknowledged as load-bearing; left as-is (pre-existing non-retryable error keeps precedence over the synthetic timeout, which is the desired behavior).

**6. `killGraceMs` not disableable** — now documented in three places: a comment at the resolver (`attempt-watchdog.ts`), `docs/subagents.md`, and both changelog entries (`0`/negative/non-numeric fall back to the default so escalation always stays bounded).

Also split the watchdog test suite into shared helpers + a separate pre-spawn candidate-filtering suite to stay under the repo's 500-line file gate. `bun run test:unit`, `typecheck`, and `check:file-length` are green.

PR #1641, opened 2026-07-06
feat(compaction): restore tiered fallback ladder for context compaction

lavaman131:
## Summary

Restores tiered fallback to Atomic's verbatim context compaction in `packages/coding-agent` as a 4-tier hybrid ladder: the current strict `compression_ratio` target stays tier 1, with graceful degradation through feasible-acceptance, a restored pre-#1399 critical planner pass, and a new deterministic no-model LRU eviction — so overflow recovery always ends in either a committed compaction that fits the effective input budget or a terminal error proving nothing more was safely deletable.

## Changes

- **Tier 1 (unchanged)**: standard planner run with strict target + nudge loop; manual `/compact` keeps strict-only behavior.
- **Tier 2**: on target miss, accept the feasible below-target reduction when projected `tokensAfter` clears the relevant budget (threshold path: the `shouldCompact` trigger boundary `effectiveInputBudget - reserveTokens`; overflow path: `getEffectiveInputBudget`). Partial salvage from a mid-run provider overflow follows the same acceptance rule instead of being discarded. Threshold never escalates past tier 2.
- **Tier 3 (overflow only)**: critical planner pass with protected entries eligible under restored pre-#1399 `critical_overflow` LRU semantics (`packages/coding-agent/src/core/compaction/context-compaction-critical.ts`), implemented as an internal relaxed transcript so every plan still flows through the unchanged fail-closed validation (last-5 recent floor via widened `preserve_recent`, latest thinking-bearing assistant immutability, task-bearing floor, tool-call/result pairing).
- **Tier 4 (overflow only, new)**: deterministic code-level LRU eviction — no model, no auth (`packages/coding-agent/src/core/compaction/context-compaction-eviction.ts`) — oldest-first eviction validated incrementally, with deterministic exchange variants (retain-smallest task anchor, retain-newest thinking) to avoid false exhaustion under non-monotone validation invariants; a bounded brute-force oracle test pins that it succeeds exactly when any validated deletion subset fits.
- **Auth-missing overflow** now routes directly to tier-4 deterministic eviction instead of today's silent no-op; overflow with no compactable transcript now fails loudly (`Context overflow recovery failed: …`) instead of no-oping.
- **Hard iteration caps (50)** on planner provider turns, planner nudges, and eviction passes; hitting a cap surfaces a terminal error carrying achieved stats.
- **No public surface changes**: the compaction barrel is byte-identical (no mode API re-exposed); extension hook shapes unchanged — extension-supplied deletion requests still bypass the ladder after local fail-closed validation (pinned by test).
- **Docs & changelog**: `docs/compaction.md` failure-paths table rewritten for the ladder; `CHANGELOG.md` gains detailed `[Unreleased]` Added/Fixed entries.

## Notes

- Validation: `bun run typecheck` ✅, `bun run lint` ✅, `bun run check:file-length` ✅, `bun run test:unit` ✅ (2834 pass / 0 fail), full CI-equivalent coding-agent vitest suite ✅ (283 files passed / 5 skipped; 2281 tests passed / 36 skipped), `bun run docs:check` ✅.
- Reviewed by a 3-reviewer adversarial loop: all three returned "patch is correct" with full requirements traceability; single remaining P3 cosmetic finding (zero-deletion exchange plan can pollute terminal exhaustion diagnostics — no incorrect commit possible) left as follow-up.
- One surfaced design decision deliberately **not** implemented (contradicts the literal objective wording): subtracting `reserveTokens` from the overflow acceptance/eviction budgets to leave headroom against re-overflow. See the implementation-notes comment for the full disposition.
- No QA E2E video: this is non-UI core-library behavior; proof is via session-level integration tests driving the real auto-compaction → validation → commit path plus the deterministic-eviction oracle.


lavaman131:
# Implementation Notes

Task: Read the file at /Users/tonystark/Documents/projects/atomic/.atomic/context/compaction-tiered-fallback-context.md first — it contains the full investigation, verified git history, current code map, the user-approved tier design, constraints, and validation commands.

Objective: restore tiered fallback to Atomic's verbatim context compaction in packages/coding-agent as a hybrid ladder, keeping the current strict compression_ratio target as tier 1 while adding graceful degradation:

- Tier 1 (unchanged): standard planner run with strict target + nudge loop.
- Tier 2: on target miss, accept the feasible x% reduction when projected tokensAfter clears the relevant budget (overflow path: effective input budget via getEffectiveInputBudget; threshold path: the shouldCompact trigger threshold so compaction does not immediately re-trigger). Partial salvage from a mid-run provider overflow follows the same acceptance rule instead of being discarded.
- Tier 3 (overflow path only): critical planner pass with protected entries eligible under LRU priority-eviction prompt semantics restored from pre-#1399 critical_overflow mode (reference: git show c5bce8ff2~1:packages/coding-agent/src/core/compaction/context-compaction.ts), enforced by validation (recent-entry boundary, latest thinking-bearing assistant fully immutable, task-bearing floor, tool-call/result pairing).
- Tier 4 (overflow path only, new): deterministic code-level LRU eviction — no model, no auth — evicting oldest safely-deletable entries through the existing validation pipeline until the effective input budget fits or nothing more is deletable. On overflow with missing auth, skip model tiers and run tier 4 directly instead of today's silent no-op.

All loops must carry hard iteration caps (max 50) so no tier can spin indefinitely; hitting a cap surfaces a terminal error with achieved stats.

Done criteria:
1. Threshold path: tier 1 → tier 2 acceptance; never escalates to protected eviction.
2. Overflow path: full ladder tier 1→2→3→4; ends with either a committed compaction that fits the effective input budget or a terminal error proving nothing more was safely deletable.
3. Feasible partial results below target are committed when they satisfy the tier-2 acceptance rule (verified by tests), instead of being thrown away.
4. All deletion plans from every tier pass the existing fail-closed validation; no public mode API is re-exposed; extension hook shapes unchanged.
5. Unit tests cover each tier transition, the acceptance rule on both paths, deterministic tier-4 eviction order/termination, auth-missing overflow behavior, and iteration caps.
6. bun run typecheck, bun run lint, bun run check:file-length, and bun run test:unit all pass.
7. packages/coding-agent/docs/compaction.md updated (especially the stale '#### Failure paths' table) to describe the ladder accurately, and packages/coding-agent/CHANGELOG.md gains an [Unreleased] entry per repo changelog rules.

## Running Notes

### What was implemented (summary)

- New `packages/coding-agent/src/core/compaction/context-compaction-critical.ts`: restored pre-#1399 critical-overflow semantics — `CONTEXT_CRITICAL_OVERFLOW_RECENT_ENTRY_COUNT = 5`, `isCriticalOverflowProtectedEntryDeletable` (protected entries deletable only when stale [outside last-5], not error-bearing, and task-bearing), `relaxTranscriptForCriticalEviction`, and the restored `<critical-overflow-mode>` prompt block.
- New `packages/coding-agent/src/core/compaction/context-compaction-eviction.ts`: tier-4 deterministic LRU eviction (`runDeterministicContextEviction`, `CONTEXT_COMPACTION_MAX_EVICTION_PASSES = 50`) — oldest-first entry eviction validated incrementally through the existing `validateContextDeletionRequest`, natural termination on candidate exhaustion, terminal errors carrying achieved stats + budget.
- `context-compaction-runner.ts`: ladder orchestration in `contextCompact` (tier-1 strict run unchanged → tier-2 acceptance → tier-3 critical planner pass → tier-4 deterministic eviction), internal `ContextCompactionLadderOptions { acceptanceTokenBudget?, criticalEvictionTokenBudget? }`, `CONTEXT_COMPACTION_MAX_PLANNER_NUDGES = 50` cap on the nudge loop, overflow-only escalation on planner exceptions (aborts always rethrow).
- `agent-session-compaction.ts`: computes ladder budgets from `options.reason` (overflow → effective input budget for both knobs; threshold → `effectiveBudget - reserveTokens` acceptance only; manual → no ladder), and routes overflow-with-missing-auth directly to tier-4 deterministic eviction instead of the silent no-op.
- `context-transcript-analysis.ts`: exported the existing private `hasAssistantError`/`hasToolResultError`/`hasFailedBashExecution` helpers (module-level export only; barrel untouched).
- Tests: `context-compaction-deletion-tool-06.suite.ts` (ladder transitions, salvage acceptance, threshold non-escalation, nudge cap), `context-compaction-eviction.test.ts` (relaxation eligibility, LRU order/termination, pairing, floor, preserve_recent, progress guarantee, determinism, cap constant), `agent-session-overflow-eviction.test.ts` (auth-missing overflow commits via tier 4; auth-missing threshold stays a silent no-op).
- Docs: `packages/coding-agent/docs/compaction.md` failure-paths table rewritten for the ladder plus overview/trigger/walkthrough/validation/extension-hook sections refreshed; `CHANGELOG.md` gained detailed `[Unreleased]` Added/Fixed entries.

### Decisions not fully specified by the research (and rationale)

1. **Relaxed-transcript approach instead of mode/policy parameter threading.** Tiers 3–4 operate on a derived transcript in which critically-eligible protected entries (per restored `isCriticalOverflowProtectedEntryDeletable`) are marked `protected: false` (entry + content blocks); `protectedEntryIds` is preserved verbatim so persisted results keep truthful protection metadata. This means zero signature changes across `canDeleteTarget`/`validateContextDeletionRequest`/`createContextDeletionTool`, the compaction barrel (`context-compaction.ts`) stays byte-identical (no public mode API), the 499-line `context-deletion-tools.ts` is untouched (500-line gate), and every non-relaxed invariant (preserve_recent recent boundary, latest thinking-bearing assistant immutability, task-bearing floor, tool pairing) remains enforced fail-closed by the *existing* validation code paths.
2. **Manual compaction (`/compact`, `contextCompact` RPC) is unchanged (strict tier-1 only).** The objective/acceptance criteria enumerate only the threshold and overflow paths for tier-2 acceptance; per the literal contract, manual keeps today's strict behavior.
3. **Ladder plumbing is numeric budgets, not a reason/mode enum.** The runner takes internal `{ acceptanceTokenBudget?, criticalEvictionTokenBudget? }`; the session layer maps reason → budgets. This avoids re-introducing anything mode-shaped near the public surface (the type is exported from the runner module for the session-layer import but deliberately NOT from the compaction barrel).
4. **Nudge-cap interpretation.** "Hitting a cap surfaces a terminal error with achieved stats" is implemented as: the cap stops further nudge follow-ups (never a hang); the run then ends and the ladder evaluates the achieved state — if a tier acceptance rule is satisfied the result is committed (graceful degradation is the point of the ladder), otherwise the failure is a terminal error carrying achieved stats (reduction %, deletion count, tokensAfter). Same logic for the tier-4 pass cap.
5. **Overflow-path error escalation.** Non-abort planner exceptions (e.g. "planner called no tools", non-overflow provider errors) on the overflow path are recorded as failed attempts and escalate to the next tier (matching the approved ladder diagram "planner fails / no auth → tier 4"); abort/cancellation always rethrows immediately. On the threshold path tier-1 exceptions propagate exactly as before.
6. **Tier-4 progress guarantee.** Success requires ≥1 validated deletion. If token estimates already "fit" the budget with zero deletions (estimate drift after a real overflow), tier 4 still evicts the single oldest safely-deletable entry rather than committing an empty plan (prevents no-op commits and re-overflow loops).
7. **Tier-4 latest-thinking pre-filter.** The newest thinking-bearing assistant entry is statically excluded from candidates; all dynamic cases (an entry *becoming* the latest retained thinking assistant after other deletions) are left to the authoritative validation, which excludes the offending candidate via the per-candidate validate-and-exclude loop.
8. **Tier-3 acceptance and short-circuit.** The tier-3 planner run accepts strict-target success OR budget-fit (≥1 deletion and projected tokensAfter ≤ effective input budget); its streamFn short-circuit and nudge predicate treat budget-fit as done so a fitting critical pass doesn't keep burning provider turns. Tier-1's short-circuit remains strict-target-only so tier 1 is genuinely unchanged.
9. **Tier-2 acceptance comparisons use the transcript-estimate `stats.tokensAfter`** (as the objective literally specifies: "projected tokensAfter clears the relevant budget"), not usage-based context accounting. Threshold acceptance budget is `getEffectiveInputBudget(model) - settings.reserveTokens`, i.e. exactly the `shouldCompact` trigger boundary (an initial `Math.max(0, …)` clamp was removed after the continuation review; see "Continuation reviewer fixes" below).

### Tradeoffs

- Tier-4 incremental validation is O(candidates × validation cost) — deliberately simple and fail-closed; acceptable for a last-resort recovery path on bounded transcripts. The 50-pass outer cap is structural belt-and-braces; the sweep naturally converges in ≤2 passes.
- The tier-4 pass cap (50) cannot be triggered organically in tests (natural convergence); tests assert the constant, natural termination, and the exhaustion terminal error instead. The nudge cap (50) is exercised behaviorally.
- The tier-1 stop-stream message text changed slightly (now phrased as "acceptable context-reduction result" to serve both strict and budget-accepted stops); no test or docs depended on the old wording.
- `runFailedAttempt` records escalated planner exceptions in the attempt's `providerError` slot, so ladder terminal errors read "provider error: …" even for non-provider planner failures — cosmetic, kept for simplicity.

### Deviations from research guidance

- None of substance. The research suggested threading pre-#1399-style mode options through validation; the relaxed-transcript equivalent was chosen (decision 1) because it achieves identical semantics with a strictly smaller public-surface and file-length footprint. The stale-vestige cleanup opportunity (plural `attempts` scaffolding) noted in the research turned out to be exactly what the ladder needed — it is now genuinely used for multi-attempt failure messages.

### Validation outcomes

- `bun run typecheck` ✅, `bun run lint` ✅, `bun run check:file-length` ✅ (1893 files), root `bun run test:unit` ✅ (2834 pass / 0 fail).
- Targeted compaction suites under Bun: 81 pass / 0 fail / 5 pre-existing skips (live-API-gated e2e tests, skipped without API_KEY).
- `bun run docs:check` in packages/coding-agent ✅ (32 docs pages).
- Full coding-agent vitest-under-bun suite (CI's "coding-agent vitest suite" step, after building natives + package dist exactly as CI does): 2257 pass / 36 skip; the only 2 remaining failures (`footer-data-provider` reftable timing tests) pass on isolated re-run and are unrelated to this change; final combined gate run (`typecheck && lint && check:file-length && test:unit`) exits 0.
- Compaction-focused vitest filter: 13 files / 103 tests pass. Auto-compaction queue suites: 16/16 pass.
- Setup note: this fresh worktree had no `node_modules`; `bun install` was run before any implementation work.

### Regression found and fixed during full-suite validation

- The full coding-agent vitest suite surfaced 6 failures in `agent-session-auto-compaction-queue-0{1,2}.suite.ts`: those suites mock the compaction barrel (`vi.mock("../src/core/compaction/index.js")`), but the initial implementation imported `contextCompact` into `agent-session-compaction.ts` directly from the runner module, bypassing the mock (the barrel's `prepareContextCompaction` fake then fed the real runner). Fix: import `contextCompact` from the barrel again (its re-export accepts the new optional ladder argument), keep `ContextCompactionLadderOptions` as a type-only import from the runner (erased at runtime, so module mocks stay authoritative), and keep `runDeterministicContextEviction` as a direct module import (intentionally not barrel-exported). All 6 tests pass after the one-line fix.
- The other 10 full-suite failures were environmental, not regressions: native grep/find/tree-sitter/PTY tests require `bun run --cwd packages/natives build` (CI runs it before the suite) — green after building natives; `workflow-agent-dir-isolation.test.ts` requires the built `dist/` of `@bastani/atomic` (CI's "Build @bastani/atomic package" step) — green after `bun run build`.

### QA E2E note

- No browser UI is involved (core library/agent-internal behavior), so no playwright QA video applies — recording one would fabricate evidence.
- TUI E2E of the ladder itself is impractical in this checkout: proving tier transitions end-to-end requires a real provider context-overflow (≈200k-token session and provider-side overflow responses). Attempts performed instead: `bun packages/coding-agent/src/cli.ts --version` boots the CLI from the modified tree (exit 0, prints the versionless-main `0.0.0` placeholder); `~/.atomic/agent/auth.json` exists, but a genuine overflow scenario would require an enormous paid provider request and is not a deterministic proof. The executable-scenario verification is covered by session-level integration tests that drive the real `_runAutoCompaction` → `_applyContextVerbatimCompaction` → commit path (backup snapshot, `context_compaction` entry, context rebuild) with faux providers, including the auth-missing overflow tier-4 commit.

## QA E2E Video

No QA E2E video applies: the change is non-UI core-library behavior (compaction ladder inside the coding agent); there is no browser scenario, and a real TUI overflow reproduction requires non-deterministic large paid provider traffic. Proof of behavior: unit + session-level integration tests through the real commit path (see Validation outcomes) and a headless CLI boot smoke test from the modified tree.

### Continuation reviewer fixes (threshold boundary + coverage)

- Removed the `Math.max(0, effectiveBudget - settings.reserveTokens)` clamp from threshold ladder wiring in `_applyContextVerbatimCompaction`; threshold acceptance now uses exactly `effectiveBudget - settings.reserveTokens`. Added an inline comment explaining the equivalence to `!shouldCompact(tokensAfter, effectiveBudget, settings)` and that a negative boundary correctly means no result can be accepted.
- Added runner-level threshold tier-2 acceptance coverage: `accepts a below-target tier-2 result on the threshold path when projected tokens fit the trigger boundary` in `context-compaction-deletion-tool-06.suite.ts`. It passes only `{ acceptanceTokenBudget }`, verifies the below-target deletion is returned, asserts `tokensAfter <= B`, and checks no `<critical-overflow-mode>` prompt appeared.
- Added session-level ladder wiring coverage in `agent-session-auto-compaction-queue-01.suite.ts`: `passes threshold and overflow ladder budgets to context compaction` verifies `contextCompact` call arg index 6 is `{ acceptanceTokenBudget: getEffectiveInputBudget(model) - reserveTokens }` for threshold and `{ acceptanceTokenBudget: effectiveBudget, criticalEvictionTokenBudget: effectiveBudget }` for overflow.
- File length check before validation: queue-01 is 496 lines and context-compaction-deletion-tool-06 is 262 lines, both under the 500-line gate.
- Validation rerun after these fixes: `bun run typecheck` ✅, `bun run lint` ✅, `bun run check:file-length` ✅, `bun run test:unit` ✅ (2834 pass / 0 fail), `bunx vitest --run compaction` ✅ (21 files passed / 1 skipped; 168 passed / 7 skipped), `bunx vitest --run agent-session-auto-compaction-queue` ✅ (2 files, 17 tests passed).
- Re-checked `packages/coding-agent/docs/compaction.md` and `packages/coding-agent/CHANGELOG.md`; existing wording already says `effectiveInputBudget - reserveTokens` and remains accurate after removing the clamp, so no docs edits were needed for this continuation.

### Final continuation verification (closure pass)

- Working tree re-verified in its completed state: 8 modified + 5 new files (204 insertions / 66 deletions), no stray artifacts; all todos closed.
- Fresh combined gate run (single command, exit 0): `bun run typecheck` ✅, `bun run lint` ✅, `bun run check:file-length` ✅ (1893 files), `bun run test:unit` ✅ (2834 pass / 0 fail), followed by `bunx vitest --run compaction agent-session-auto-compaction-queue` ✅ (21 test files passed / 1 skipped; 168 tests passed / 7 skipped — the 7 skips are pre-existing live-API-gated e2e tests).
- No further contract gaps identified; the prior adversarial review's 11/11 items remain PASS. No commit made (later stage handles PR/commit).

### Failure-mode review fixes (protectedEntryIds disjointness)

- Addressed `review/compaction-failure-modes.md` Finding 1: `relaxTranscriptForCriticalEviction` now rebuilds `protectedEntryIds` from entries that remain protected after relaxation, so tier-3/tier-4 validation results copied by `validateContextDeletionRequest` keep `deletedTargets` disjoint from the effective protected id list used in persisted compaction entries, `session_compact` events, and the interactive summary count.
- Updated tests: `context-compaction-eviction.test.ts` now asserts relaxed `protectedEntryIds` excludes relaxed stale task-bearing entries while retaining still-protected recent/error-bearing ids, and both deterministic eviction + tier-3/tier-4 ladder tests assert `protectedEntryIds` contains none of the deleted entry ids.
- Softened the threshold-budget comment in `agent-session-compaction.ts` to clarify the `shouldCompact` equivalence is in transcript-estimate space; the live retry trigger uses provider usage. Did not change overflow budgets: the reviewer's reserve-subtraction suggestion conflicts with the literal objective's effective-input-budget requirement and remains a design decision to surface separately.
- Validation after this fix: `bun run typecheck` ✅, `bun run lint` ✅, `bun run check:file-length` ✅, `bun run test:unit` ✅ (2834 pass / 0 fail), `bunx vitest --run compaction agent-session-auto-compaction-queue` ✅ (21 passed / 1 skipped; 168 passed / 7 skipped).
### Reviewer findings disposition (review/compaction-failure-modes.md)

- **Finding 1 (Medium, FIXED)** — protectedEntryIds/deletedTargets overlap for tier-3/4 results: fixed as described above; disjointness assertions added to eviction and ladder tests.
- **Finding 2 (Medium, SURFACED — deliberately NOT implemented)** — the reviewer suggests subtracting `reserveTokens` from the overflow acceptance/eviction budgets so recovery lands with headroom (real prompts add system prompt + tool schemas beyond the transcript estimate, and tier-4 stops at the first fitting plan, so a "successful" recovery can re-overflow and hit the one-shot retry guard). This is a real failure mode, but the suggested fix **contradicts the literal objective**, which pins the overflow budget to "effective input budget via getEffectiveInputBudget" for tier 2 and "until the effective input budget fits" for tier 4. Per the literal contract, the objective wording wins and the conflict is surfaced here instead of silently implemented. The reviewer itself notes docs/CHANGELOG state this budget deliberately. The actionable sub-item (an over-claiming code comment about threshold re-trigger) WAS fixed (comment now scoped to transcript-estimate space). Recommend the user decide post-merge whether overflow budgets should gain a reserve margin.
- **Finding 3 (Low, ASSESSED — cap retained by design)** — the tier-4 50-pass cap cannot bind because a single sweep plans-or-excludes every candidate (exclusions are monotone; reviewer confirmed fail-closed correctness). The reviewer's option of dropping the cap would violate the contract requirement that "all loops carry hard iteration caps (max 50)", and adding real retry semantics would add behavior beyond the contract. Decision: keep the cap as a defensive structural bound and document it — a clarifying comment was added above the pass loop in `context-compaction-eviction.ts`. CHANGELOG's "capped at 50 passes" remains literally true (the bound exists and is enforced structurally).
- **Informational notes (confirmed deliberate)** — (1) tier-1 ignores the acceptance budget mid-run: intentional, the contract requires tier 1 unchanged (strict target + nudges); acceptance is evaluated post-run (decision 8). (2) tier-3 starts a fresh deletion tool and does not seed tier-1's validated deletions: known cost/latency tradeoff, correctness preserved (tier-4 recomputes deterministically); seeding would be an enhancement beyond the contract.
- Post-fix validation (fresh, exit 0): typecheck ✅, lint ✅, check:file-length ✅, test:unit ✅ (2834/0), compaction+queue vitest ✅ (21 files passed / 1 skipped; 168 tests passed / 7 pre-existing live-API skips).
- Between-pass workflow artifacts `progress.md` and `review/` (created by the research/review stages, not this implementation) were left in place for the workflow harness.

### Review-artifact re-verification (subsequent continuation pass)

- The `review/` files were re-flushed by the workflow harness with newer mtimes but identical content (same Findings 1–3 citing pre-fix code). Verified each premise is stale against the current tree: `relaxTranscriptForCriticalEviction` rebuilds `protectedEntryIds` from still-protected entries (Finding 1 fixed); the eviction test now asserts `relaxed.protectedEntryIds` **not**.toEqual(original) plus an exact still-protected list, and suite 06 uses an `expectProtectedIdsDisjointFromDeleted` helper on tier-3/4 results; the Finding-3 defensive-cap comment is present. Findings 2 (contract conflict, surfaced not implemented) and 3 (cap retained per contract) dispositions unchanged.
- One-off flake observed: a single root `test:unit` run reported 1 fail in the workflows durable-write test area (timing-sensitive; emits "durable write failed" warnings by design; entirely outside this change's surface — the diff touches only packages/coding-agent). Three consecutive re-runs: 2834 pass / 0 fail each. Compaction+queue vitest remained green (21 files passed / 1 skipped; 168 tests passed / 7 pre-existing live-API skips).

### Round-latest research fixes (F1/F2/F3)

- **F1 budget gate fixed:** overflow-path strict-target results are now accepted only when `validatedResult.stats.tokensAfter <= criticalEvictionTokenBudget`; manual and threshold target-met acceptance remain unchanged. The in-run critical-pass predicate now stops only on budget-fit acceptance when an `acceptanceTokenBudget` is supplied, so target-met-over-budget critical results continue to nudge/finish and then escalate instead of committing.
- **F2 total turn cap restored:** `CONTEXT_COMPACTION_MAX_TURNS = 50` is exported, and each planner run counts real provider calls in `streamFn` after the acceptance short-circuit. At the cap it returns a synthetic stop stream (`Reached the context compaction turn cap...`) so achieved validated deletions still flow through the existing attempt/ladder/terminal formatting. Synthetic stops are not counted.
- **F3 prefilter narrowed:** deterministic tier-4 eviction now excludes only the actual latest assistant when that assistant contains thinking/redacted-thinking; older thinking-bearing assistants can be attempted and are still guarded by authoritative per-candidate validation.
- Added `context-compaction-deletion-tool-07.suite.ts` for F1/F2 regressions and extended `context-compaction-eviction.test.ts` for F3. Docs/changelog were updated to mention overflow strict-target budget gating and both planner caps (50 provider turns + 50 nudges).
- Validation after F1/F2/F3 fixes: `bun run typecheck` ✅, `bun run lint` ✅, `bun run check:file-length` ✅ (1893 tracked files), `bun run test:unit` ✅ (2834 pass / 0 fail), `cd packages/coding-agent && bunx vitest --run compaction agent-session-auto-compaction-queue` ✅ (22 files passed / 1 skipped; 175 tests passed / 7 skipped), and `bun run docs:check` ✅ (32 docs pages).

### P2 critical recent-boundary hardening

- Implemented the wider pre-#1399 critical recent-entry boundary: `criticalCompactionParameters()` now widens critical overflow parameters to `max(preserve_recent, 5)`, `relaxTranscriptForCriticalEviction()` writes those parameters onto the relaxed transcript via `getTranscriptCompactionParameters()`, and the tier-3 planner run passes the widened parameters so `runContextDeletionAssistant()` does not overwrite the transcript back to the standard preserve window.
- Tier-4 needed no direct eviction logic change: `runDeterministicContextEviction()` already filters and validates against the relaxed transcript, so existing `canDeleteTarget()` / `validateContextDeletionRequest()` recent guards now reject all last-5 entries, including otherwise-unprotected assistant/tool entries.
- Fixture rework: deterministic-eviction tests and suite-07 overflow-budget fixtures were padded with five tail entries so the intended stale deletion targets remain outside the new critical last-5 floor. Assertions still check the same LRU order, pairing repair, task-bearing floor, latest-thinking guard, strict budget gating, and deterministic behavior; only the transcript shape changed to keep those targets eligible under the restored invariant.
- New coverage: deterministic tier-4 test `never evicts entries inside the critical last-5 floor`; relaxation/validation assertion that effective `preserve_recent` is widened to 5 and rejects an unprotected last-5 target; tier-3 planner test `critical planner rejects unprotected entries inside the last-5 floor but can delete older equivalents`.
- Docs/changelog tightened to state tiers 3/4 enforce an effective recent guard of `max(preserve_recent, 5)` across all entries. Validation: typecheck ✅, lint ✅, check:file-length ✅, coding-agent compaction+queue vitest ✅ (22 passed / 1 skipped; 177 passed / 7 skipped), docs:check ✅. First `bun run test:unit` hit an unrelated root watchdog flake (`subagents-attempt-watchdog` skipped the stalled model log); immediate rerun passed ✅ (2834 pass / 0 fail).

### Final-round fixes (F1 fixture, F2 extension bypass, F3 greedy replacement)

- **F1 fixture repaired:** `agent-session-overflow-eviction.test.ts` now keeps the 4 large stale assistant payloads unchanged and appends a 5-entry tiny recent continuity tail, so deterministic no-auth overflow eviction can delete enough stale entries while respecting the restored critical last-5 floor. Focused test now passes: 2/2.
- **F2 extension-bypass carve-out pinned:** no production behavior change. Added an overflow auto-compaction extension test proving an extension-provided deletion request commits even when `stats.tokensAfter` remains above the effective input budget, `ModelRegistry.getApiKeyAndHeaders` is not called, a `context_compaction` entry is persisted, `compaction_end` carries a result, and `session_compact.fromExtension === true`. Docs/changelog now state the overflow budget-fit guarantee applies to Atomic's internal ladder results; extension deletion requests bypass the ladder and its overflow budget gate after local fail-closed validation.
- **F3 deterministic eviction remedy:** `runDeterministicContextEviction()` now handles the latest-retained-thinking non-monotone case. If `planned + candidate` fails validation and prior planned targets include thinking-bearing assistant entry deletions, it retries after dropping those prior thinking targets and adding the current candidate. On success it adopts the validated/reconciled target set and leaves dropped thinking ids unexcluded for a later pass; on retry failure the current candidate is excluded as before. This preserves strict oldest-first behavior in non-conflict cases, is deterministic, and terminates because candidates still move through planned/dropped/excluded states under the existing 50-pass cap.
- Added regression coverage: `context-compaction-eviction.test.ts` now asserts the replacement case deletes newer assistant A/B, does not delete the older thinking assistant, and fits the budget; existing thinking-deletable and latest-thinking-skip tests remain green.
- Validation: focused changed-file vitest (`agent-session-overflow-eviction`, `context-compaction-eviction`, `compaction-extensions`) ✅ (23 passed / 1 pre-existing live-API skip); `bun run typecheck` ✅; `bun run lint` ✅; `bun run check:file-length` ✅; `bun run test:unit` ✅ (2834 pass / 0 fail); full coding-agent CI-equivalent `cd packages/coding-agent && bun run --bun test` had only the known `footer-data-provider` reftable flake (2273 passed / 36 skipped / 1 failed), and `bunx vitest --run test/footer-data-provider.test.ts` immediately passed ✅ (10/10); `bun run docs:check` ✅.
- F4 git hygiene: no action taken on `packages/workflows/builtin/goal.ts`; research identified it as branch lag/out of scope. Overflow headroom remains quarantined/design-level per prior notes.

### Final-round fixes (final-f1..f4 research batch)

- **F1 (broken fixture, FIXED)** — the P2 last-5 floor made `agent-session-overflow-eviction.test.ts`'s 7-entry fixture unsatisfiable (3 of 4 large stale assistants fell inside the critical floor; eviction could only reach 1239 tokens vs the 200 budget). Verified RED before fixing (`bunx vitest --run agent-session-overflow-eviction` → 1 failed) — the failure had been invisible to `compaction`-filtered runs, which do not match that filename. Fixture now keeps the 4 large stale assistants and appends a 5-entry tiny continuity tail; test green. Process lesson recorded: always finish with the FULL coding-agent suite (`bun run --bun test`), not name-filtered subsets.
- **F2 (extension bypass, PINNED — Option C)** — contract-mandated adjudication (context constraints: extension deletionRequests "bypass the ladder as today"; hook shapes unchanged; no breaking changes): behavior unchanged; added pinning tests asserting an overflow extension deletionRequest commits even when validated tokensAfter exceeds the effective input budget, planner auth is never resolved, and `session_compact.fromExtension === true`; docs now state explicitly that the overflow budget-fit guarantee applies to Atomic's internal ladder results and extension-supplied requests bypass the ladder and its budget gate after local fail-closed validation.
- **F3 (tier-4 greedy incompleteness, FIXED)** — narrow deterministic replacement-on-conflict: when `validate(planned + candidate)` fails and `planned` contains thinking-bearing assistant entry targets, retry with those targets dropped; on success adopt the new plan and let the next outer pass re-attempt the dropped entries (they then re-validate or get permanently excluded). Keeps strict oldest-first order in all non-conflict cases; termination: entries move monotonically through planned/dropped/excluded with drops bounded by retry successes (each planning a distinct new candidate), under the 50-pass cap which now has genuine multi-pass work (stale "cannot bind" comment updated). Regression test per research: {older thinking (small), newer A, newer B} with a budget only {A,B} fits — eviction now deletes A+B, keeps the older thinking entry, fits budget (previously a false "nothing more was safely deletable" terminal error).
- **Orchestrator-caught bug in the worker's F3 change (FIXED)** — storing reconciled targets in `planned` (needed so the thinking-filter sees reconciliation-added thinking deletions) broke the request mapper: the local mapper dropped `blockIndex` from content_block targets, so mixed-content pairing scenarios would throw "Invalid content block index" on re-validation and wrongly exclude candidates. Replaced the local mapper with the existing `deletionRequestFromTargets` helper (preserves blockIndex).
- **F4 dispositions** — `packages/workflows/builtin/goal.ts` differences are branch lag vs origin/main (worktree detached at a05fc5fe4), not a revert: no action, do not rebase mid-stage. Overflow headroom finding remains quarantined as design-level (conflicts with the literal effective-input-budget objective wording).
- **Validation (final, all green)**: `bun run typecheck` ✅, `bun run lint` ✅, `bun run check:file-length` ✅, `bun run test:unit` ✅ (2834/0), and the FULL CI-equivalent coding-agent suite `bun run --bun test` ✅ exit 0 — 283 test files passed / 5 skipped; 2274 tests passed / 36 skipped (includes the repaired overflow-eviction test and the previously flaky footer reftable tests). `bun run docs:check` ✅.

### Last confirmed gap fix — no preparable transcript on overflow

- Fixed `_applyContextVerbatimCompaction()` so `prepareContextCompaction(...) === undefined` remains a silent no-op for threshold/manual, but throws on overflow with `Context compaction found no compactable transcript entries; nothing more was safely deletable`. `_runAutoCompaction` catches this and emits the intended loud `compaction_end.errorMessage` prefixed with `Context overflow recovery failed:`.
- Added `agent-session-overflow-eviction.test.ts` coverage for the no-preparable-transcript branch: overflow emits a terminal compaction_end error and persists no `context_compaction` entry; threshold on the same one-entry fixture remains a silent no-op and also persists no `context_compaction` entry. Focused test: 4/4 pass.
- Updated `packages/coding-agent/docs/compaction.md` failure-path table with threshold no-preparation silent no-op and overflow no-preparation terminal error rows, and extended the `[Unreleased]` fixed changelog bullet to cover the no-compactable-transcript overflow case.
- Validation: `bun run typecheck` ✅, `bun run lint` ✅, `bun run check:file-length` ✅ (1893 files), `bun run test:unit` ✅ (2834 pass / 0 fail), `cd packages/coding-agent && bun run --bun test` ✅ on rerun (283 files passed / 5 skipped; 2276 tests passed / 36 skipped), and `bun run docs:check` ✅. First full coding-agent run hit unrelated async bash truncation flake in `test/tools-04-01.suite.ts`; isolated rerun passed 21/21, then full rerun passed.

### Latest research round: unpreparable-overflow terminal error + stale bidirectional finding

- **Finding 1 (FIXED)** — overflow recovery with no preparable transcript previously hit `if (!preparation) return undefined` in `_applyContextVerbatimCompaction` and silently no-opped (same class as the fixed auth-missing no-op, and neither of Done 2's allowed outcomes). Now overflow throws `Context compaction found no compactable transcript entries; nothing more was safely deletable` (surfaced by `_runAutoCompaction` as "Context overflow recovery failed: ..."); threshold keeps its silent no-op and manual keeps its existing "Nothing to compact" wrappers. New tests: `surfaces a terminal overflow error when no compactable transcript can be prepared` and `keeps threshold auto-compaction with no preparable transcript as a silent no-op` (agent-session-overflow-eviction.test.ts). Docs failure-paths row + CHANGELOG Fixed bullet extended.
- **Finding 2 (STALE — not implemented, evidence)** — the requested "bidirectional backtracking" regression for the `[older-thinking, newer-a, newer-b]` shape already exists from the prior pass (`drops an older thinking deletion when it blocks a newer fitting deletion plan`, context-compaction-eviction.test.ts ~189) and passes. Reachability argument: under oldest-first candidate order, a thinking-bearing candidate can only be blocked by deletions of NEWER assistants, which are attempted later — so the reverse conflict direction the finding asks to handle cannot arise at attempt time; the implemented forward-direction replacement covers the reachable class. Per the literal contract, no behavior added beyond requirements.
- Research "do not reopen" list confirmed consistent with prior dispositions: zero-headroom contradicts the objective (stays surfaced), extension bypass pinned, goal.ts branch lag.
- Validation (final, independent): typecheck ✅, lint ✅, check:file-length ✅, test:unit ✅ (2834/0), FULL coding-agent suite `bun run --bun test` ✅ exit 0 (283 files passed / 5 skipped; 2276 tests passed / 36 skipped), docs:check ✅ (worker). Worker's first full run hit an unrelated tools-04-01 async-bash flake; isolated + full reruns green.

### Delegated worker update: generalized tier-4 exchange for task-bearing floor

- Implemented reviewer-b fix in `context-compaction-eviction.ts`: deterministic eviction now recognizes both thinking-bearing assistant targets and task-bearing entry targets as exchange categories. On candidate validation failure it tries, in order, planned-minus-thinking, planned-minus-task-bearing, and planned-minus-both plus the current candidate. A variant is adopted only if validation succeeds and the validated plan strictly lowers `stats.tokensAfter` compared with the current best result; otherwise the candidate is excluded. This preserves LRU candidate order as the primary sweep while preventing false exhaustion for non-monotone validation invariants and preventing worse exchanges.
- Updated the pass-loop termination comment: exchanges strictly decrease non-negative `tokensAfter`, exclusions grow monotonically, and the 50-pass cap remains the hard safety bound.
- New tests in `context-compaction-eviction.test.ts`: `drops an older task-bearing deletion when it blocks a newer fitting deletion plan` and `does not replace a better task-bearing plan with a worse validating exchange`. Existing thinking-conflict, pairing, floor, preserve_recent, latest-thinking, last-5, and determinism tests stayed green unchanged.
- No docs/CHANGELOG edits were required for this sub-change: current docs/changelog describe deterministic LRU eviction and validation-guarded safe deletion generally and do not mention a thinking-only exchange.
- Validation: `bun run typecheck` ✅, `bun run lint` ✅, `bun run check:file-length` ✅ (1893 tracked files), `bun run test:unit` ✅ (2834 pass / 0 fail), `cd packages/coding-agent && bun run --bun test` ✅ (283 files passed / 5 skipped; 2278 tests passed / 36 skipped), and `cd packages/coding-agent && bun run docs:check` ✅.

### Reviewer-b fix: generalized tier-4 exchange (task-bearing floor) with strict-improvement guard

- The task-bearing floor is non-monotone like latest-retained-thinking; the previous exchange dropped only thinking targets, so a small old task-bearing deletion accepted first could permanently block a large newer task-bearing deletion (floor keeps one), producing a false "nothing more was safely deletable" exhaustion claim.
- `validatedExchange()` now tries deterministic variants in order — drop planned thinking targets, drop planned task-bearing targets, drop both — each revalidated fail-closed, and adopts a variant only when it validates AND strictly lowers `stats.tokensAfter` versus the current plan. The guard both prevents regressions (a smaller newer candidate can never displace a larger planned deletion — new test pins this) and guarantees termination (adoptions strictly decrease a non-negative integer; exclusions grow monotonically; 50-pass cap remains the belt). LRU oldest-first stays the primary order; `deletionRequestFromTargets` keeps reconciled `blockIndex` intact.
- New tests: `drops an older task-bearing deletion when it blocks a newer fitting deletion plan` and `does not replace a better task-bearing plan with a worse validating exchange`; all existing eviction/thinking/pairing/floor/last-5/determinism tests unchanged and green.
- Validation (independent): typecheck ✅, lint ✅, check:file-length ✅, test:unit ✅ (2834/0); all compaction-related vitest files ✅ (23 files, 185 pass / 7 pre-existing live-API skips). Full coding-agent suite: worker's run green (2278 pass / 36 skip); my two re-runs each hit exactly one of the two documented pre-existing flakes in areas untouched by this change (tools-04-01 async-bash truncation timing, footer-data-provider reftable timing) — each passes in isolation (21/21 and 10/10 respectively), and an earlier full run this session was fully green (exit 0, 283 files). docs:check ✅.

### K-task-anchor exchange hardening (retain-smallest + oracle)

- Extended `validatedExchange()` beyond all-or-nothing task drops: exchange variants now include retain-smallest task-anchor plans (smallest by relaxed entry tokenEstimate, transcript order tie-breaker), plus the same variant with thinking targets removed. All variants (existing thinking/task/both drops plus retain-smallest variants) are validated fail-closed, deduplicated, and the BEST strictly improving result (lowest `tokensAfter`; construction order breaks ties) is adopted. This closes the k>=3 task-anchor case where the best floor-respecting deletion plan must keep the smallest task anchor and delete the larger anchors.
- Kept the previous safety properties intact: oldest-first candidate order remains the base order; `deletionRequestFromTargets` preserves `blockIndex`; dropped ids are retried on later passes; adoptions strictly decrease `tokensAfter`, exclusions grow monotonically, and the 50-pass cap remains the terminal guard. Avoided `Array.prototype.toSorted` after package-root declaration tests showed emitted-package consumers target an older lib; used `[...targets].sort(...)` instead.
- Added tests in `context-compaction-eviction.test.ts`: three-anchor regression deletes `{task-medium, task-large}` while retaining `task-small`, and a bounded brute-force oracle over four <=8-entry transcripts (task-anchor, thinking-conflict, tool-pair closure, exhaustion) checks that deterministic eviction succeeds exactly when any validated entry-deletion subset fits. Existing expectations did not need weakening.
- Validation (worker final): focused eviction test ✅ (15/15), typecheck ✅, lint ✅, check:file-length ✅ (files remain under 500 lines: eviction 183, test 353), test:unit ✅ (2834/0; one durable workflow flake on first run, rerun green), package-root/context-window + footer focused tests ✅ (26/26 after fixing `toSorted`), full coding-agent suite ✅ (283 files passed / 5 skipped; 2280 tests passed / 36 skipped), docs:check ✅.

### k-task-anchor exchange completion + branch-lag disposition

- **k-anchor fix (implemented)** — with ≥3 relaxed task anchors, the all-or-nothing task drop could converge to retaining a non-smallest anchor (3-anchor trace: {t1,t3} instead of optimal {t2,t3}) and falsely claim exhaustion. `exchangePlans()` now also constructs retain-smallest task variants (retain the min-token anchor among planned+candidate, transcript-order tiebreak; plus a minus-thinking flavor), deduplicated by target signature, and `validatedExchange()` adopts the BEST valid variant (lowest tokensAfter, earlier variant wins ties) under the unchanged strict-improvement guard. Subtle case verified: when the candidate itself is the smallest anchor, the adopted plan may exclude the candidate; the next pass re-attempts it and permanently excludes it (guard rejects the non-improving re-exchange) — termination preserved (strict monotone decrease + monotone exclusions + 50-pass cap). Implementation avoids `toSorted` for older-lib declaration compatibility.
- **New tests** — 3-anchor retain-smallest regression (fails pre-fix) and a bounded brute-force oracle property test: for small crafted transcripts (task anchors, thinking conflicts, tool pairs, exhaustion case; ≤2^8 subsets), the algorithm returns a fitting result EXACTLY when any validated subset fits, else throws the terminal exhaustion error — pinning the truthfulness of "nothing more was safely deletable" broadly. Eviction module now 183 lines; eviction test file 353 lines.
- **Branch-lag disposition (research Finding 2)** — origin/main is 4 commits ahead of this detached worktree, but `git diff HEAD origin/main` over ALL compaction-related paths (compaction src dir, session files, tests, docs/compaction.md, CHANGELOG.md) is EMPTY — zero conflict surface. Refining the research's "rebase before final review": rebasing a dirty detached worktree mid-stage adds risk for no benefit; integration happens at the commit/PR stage. Evidence recorded here.
- **Validation (independent)** — typecheck ✅, lint ✅, check:file-length ✅, test:unit ✅ (2834/0); eviction suite 15/15; full coding-agent suite: worker's run fully green (2280 pass / 36 skip); my re-run hit only the documented footer reftable flake (isolated 10/10 — same pre-existing timing flake as before, untouched area). docs:check ✅.

### Retain-newest-thinking exchange completion

- Completed the tier-4 exchange variant space for the latest-retained-thinking invariant: `exchangePlans()` now constructs the deterministic cross-product of thinking handling (`keep-all`, `drop-all`, `retain-newest`) and task handling (`keep-all`, `drop-all`, `retain-smallest`), minus the no-op, with signature dedup. Retain-newest-thinking removes the newest thinking-bearing target from the deletion plan (keeps it in context) while deleting older thinking targets; best-of adoption and the strict `tokensAfter` improvement guard remain unchanged.
- Added the verified counterexample regression (`thinking-large`, `thinking-small-newest`, `latest-non-thinking`, user recent tail): budget fits only `{thinking-large, latest-non-thinking}`, so the newest thinking assistant must be retained. Expanded the bounded oracle inventory with multi-thinking, mixed task/thinking, and exact-boundary / one-below-boundary budgets (subsets remain <= 2^8).
- File-size status after this pass: `context-compaction-eviction.ts` 222 lines; `context-compaction-eviction.test.ts` 400 lines, both under the 500-line gate. Validation: typecheck ✅, lint ✅, check:file-length ✅, test:unit ✅ (2834/0), full coding-agent suite `bun run --bun test` ✅ (283 passed / 5 skipped; 2281 passed / 36 skipped), docs:check ✅.

### Retain-newest-thinking variants: exchange cross-product completed

- The latest-retained-thinking invariant is position-aware: retaining the NEWEST planned thinking target in context shields all older thinking deletions. Verified counterexample pre-fix: [T1 thinking 40, T2 thinking 5, N latest non-thinking 35], budget only {T1,N} fits — drop-all-thinking {N} was guard-rejected (deletes 35 < 45), N excluded, false exhaustion with {T1,T2}; retain-newest {T1,N} (retain T2) is valid and fits.
- `exchangePlans()` now builds the full deterministic 3×3 cross-product — thinking ∈ {keep-all, drop-all, retain-newest} × task ∈ {keep-all, drop-all, retain-smallest}, minus the no-op — 8 variants, signature-deduplicated, fixed construction order; best-of adoption and the strict tokensAfter-improvement guard unchanged (termination argument unchanged: strict monotone decrease + monotone exclusions + 50-pass cap). Eviction module 222 lines; eviction test 400 lines.
- Tests: retain-newest-thinking regression (`retains the newest planned thinking assistant when that enables an older thinking plus newer assistant deletion`, fails pre-fix) + oracle domain widened per research (multi-thinking, mixed task/thinking cross-class coupling, exact-boundary budget, one-below-boundary budget → terminal exhaustion). Eviction suite now 16/16.
- Extension-bypass re-raise in review/subagent-code-analysis.md: disposition UNCHANGED — the bypass is contract-mandated ("extension-provided deletionRequests bypass the ladder as today"), documented in docs/CHANGELOG, and pinned by test; the review itself cites the pinning test.
- Validation (independent, this pass): typecheck ✅, lint ✅, check:file-length ✅, test:unit ✅ (2834/0), FULL coding-agent suite ✅ exit 0 (283 files passed / 5 skipped; 2281 tests passed / 36 skipped — no flakes this run), docs:check ✅ (worker).

PR #1677, opened 2026-07-08
fix(tui): suppress Working loader for blocking custom UI overlays

### LINKED ISSUE #1670: Workflow input/connect UI incorrectly shows Working spinner
## Summary

On the latest Atomic, the global `Working...` spinner appears in workflow UI states where no agent/workflow work is actually running:

- Running `/workflow <name>` opens the workflow input form before any workflow run starts, but the global spinner is shown while the user is just filling in inputs.
- Running `/workflow connect` opens the run picker, but the global spinner is shown while the user is just choosing a run.
- `/workflow status` and `/workflow resume` did not reproduce in initial testing.

Screenshots from the repro:

- `/tmp/atomic-clipboard-f16ebdd8-e051-4f02-b35a-84b11178c48d.png`
- `/tmp/atomic-clipboard-59c58eb4-562a-4903-98b0-c677316a4336.png`

## Likely root cause

Interactive input handling pre-mounts the working spinner for every submitted prompt before slash-command dispatch:

- `packages/coding-agent/src/modes/interactive/interactive-input-handling.ts`

Workflow UI commands then await interactive user input without hiding that pre-mounted spinner:

- Workflow inline form: `packages/workflows/src/tui/inline-form-overlay.ts`
- Workflow session picker: `packages/workflows/src/tui/session-overlays.ts`

There is already older precedent for the desired behavior in:

- `packages/workflows/src/tui/inputs-overlay.ts`

That picker explicitly hides/restores the host working row via `setWorkingVisible` while the picker is mounted.

## Acceptance criteria

- The global `Working...` spinner is hidden while the workflow input form is mounted before a run starts.
- The global `Working...` spinner is hidden while the `/workflow connect` session picker is mounted.
- The spinner is restored when actual workflow/agent work starts, or when the UI settles/closes.
- Regression tests cover submit, cancel, and error/unsupported paths for the affected workflow UI surfaces.

flora131:
## Summary

Fixes #1670: the host's global `Working...` spinner kept rendering underneath workflow slash-command overlays that were waiting on user input (e.g. `/workflow <name>` input forms, the session connect/resume pickers, the kill-confirm dialog), making the TUI look busy when it was actually idle and blocked on a keypress.

Rather than having each overlay manually hide/restore the spinner, the fix centralizes suppression in the interactive host: any non-overlay `ctx.ui.custom()` mount (the default, or explicit `{ overlay: false }`) is now treated as blocking user input and automatically suppresses `Working...` for as long as it's mounted, restoring it when the component settles or is dismissed. Floating `{ overlay: true }` custom UIs are left alone since they can be passive views over still-active work (e.g. the kill-confirm modal). The workflows package's overlays are updated to rely on this host behavior instead of duplicating hide/restore logic, and `inline-form-overlay.ts` is migrated from the legacy `setEditorComponent`/`getEditorComponent` editor-swap to mounting through `ctx.ui.custom()` directly so it participates in the same lifecycle.

## Changes

- **`interactive-extension-runtime.ts`** / **`interactive-extension-context.ts`** / **`interactive-agent-events.ts`** / **`interactive-mode-surface.ts`**: Added `isWorkingLoaderAllowed()` (true only when `workingVisible` and `blockingInlineCustomUiDepth === 0`) and `refreshWorkingLoaderVisibility()`, and wired `beginHostInlineCustomUi`/`setWorkingVisible`/the agent-event loader restart path through them so the spinner is suppressed for the full lifetime of any blocking inline custom UI, not just while `workingVisible` is explicitly toggled off.
- **`inline-form-overlay.ts`**: `openInlineInputsForm` now mounts the inline form via `ctx.ui.custom(factory, { overlay: false })` instead of swapping in a custom editor with `setEditorComponent`/`getEditorComponent`, removing the bespoke "was our factory still installed" staleness checks needed by the old swap-and-restore approach (including the `/new`/`/resume`/`/fork`/`/reload` stale-context case). Working-loader suppression is now handled entirely by the host for the duration of the mount.
- **`inputs-overlay.ts`**: Removed `openInputsPicker`'s own `hideWorking`/`restoreWorking` closures and the `setWorkingVisible` surface requirement — the host now suppresses the spinner for the picker's mounted lifetime.
- **`session-overlays.ts`**: Removed manual Working-visibility handling from `openSessionPicker`/`openKillConfirm`. Hardened mount-failure handling so both a synchronous throw and an async rejection from `custom(...)` resolve to a safe default (`{ kind: "close" }` / `false`) instead of leaving the promise unsettled, and added `ConfirmUiSurface` to the surface types.
- **`workflow-resume-selector.ts`**: Applied the same defensive synchronous/async mount-failure handling to `openWorkflowResumeSelector`, settling to `{ kind: "close" }` on either failure path.
- **`workflow-command-registration.ts`**: Aligned `canOpenPicker`'s capability gate to only require `ctx.ui.custom` (dropped the now-dead `typeof ctx.ui?.setEditorComponent === "function"` fallback check), and always route through `openInlineInputsForm` rather than gating that call on `setEditorComponent` support, so custom-only UI hosts correctly reach the inline input form instead of skipping straight to the picker fallback.
- **`inline-form-editor.ts`**: Updated doc comments to describe the component as mounted via non-overlay `ctx.ui.custom()` rather than swapped in via `setEditorComponent`.
- Docs: `packages/coding-agent/docs/extensions.md` and `docs/tui.md` document that non-overlay `ctx.ui.custom()` suppresses the global `Working...` loader while mounted, and that `{ overlay: true }` opts out.
- Changelog entries added under `[Unreleased] / Fixed` in both `packages/coding-agent/CHANGELOG.md` and `packages/workflows/CHANGELOG.md`.

## Tests

- `packages/coding-agent/test/interactive-mode-status-custom-ui.suite.ts`: new coverage asserting the Working loader is stopped for the duration of a non-overlay `showExtensionCustom` mount and restarted after it settles, and that overlay (`{ overlay: true }`) mounts do not suppress it.
- `test/unit/overlay-mount-placement.test.ts`: new assertions that `openSessionPicker`/`openKillConfirm`/`openWorkflowResumeSelector` never call `setWorkingVisible` themselves (connect, close, dispose, and mount-failure paths), confirming visibility is left entirely to the host.
- `test/unit/inline-form-04.test.ts`, `test/unit/inline-form-helpers.ts`, `test/unit/inputs-picker-01.ts`, `test/unit/inputs-picker-helpers.ts`: updated to mount `openInlineInputsForm`/`openInputsPicker` through the `ctx.ui.custom()` surface instead of the retired `setEditorComponent`/`getEditorComponent` mocks.
- `test/integration/mock-extension-api-workflow-actions.test.ts` / `test/integration/overlay-entrypoints-helpers.ts`: new coverage confirming custom-only UI hosts (no `setEditorComponent`) reach the inline workflow input form instead of skipping to the fallback picker.

## Validation

- `AGENT=1 bun test test/unit/inline-form-04.test.ts test/unit/inputs-picker-01.ts test/unit/overlay-mount-placement.test.ts packages/coding-agent/test/interactive-mode-status-custom-ui.suite.ts`
- `AGENT=1 bun run typecheck`
- `AGENT=1 bun run check:file-length`
- `AGENT=1 bun run test:unit`


flora131:
Implementation direction I want to switch toward:

Instead of having each workflow overlay manually hide/restore the global `Working...` indicator, centralize this in the host/runtime layer.

The core issue is a UI-state mismatch: after the eager spinner change, `Working...` can appear immediately after command submit, but workflow commands may then open a blocking form/picker and wait for human input. While a blocking custom UI overlay is active, the app is not really “working”; it is waiting on the user.

Preferred approach:

- Treat blocking custom UI overlays as a host-level “waiting for user input” state.
- Suppress the global `Working...` indicator automatically while such an overlay is mounted.
- Restore normal spinner behavior when the overlay settles/disposes.
- Remove per-overlay `setWorkingVisible(false/true)` calls and the `working-visibility-guard.ts` helper.
- Keep focused regression tests for the original cases (`/workflow <name>` input form and `/workflow connect` picker) plus submit/cancel/dispose restoration.

This keeps ownership in the right layer: overlays should not need to know about global spinner state. The current helper-based solution is safe and pragmatic, but it spreads global UI-state responsibility into individual overlay implementations. Central suppression should be cleaner and less error-prone for future overlays.

PR #1710, opened 2026-07-10
fix(web-access): retry failed lazy initialization

### LINKED ISSUE #1704: Lazy-loaded tool calls fail on first invocation instead of awaiting resource initialization
## Observed behavior

A user report confirms that, in fresh subagent sessions, the first invocation of an initialization-dependent search/fetch-like tool can fail while its resources continue loading. Retrying the identical call then succeeds. The likely concrete tool names are `web_search` and `fetch_content`; this report does **not** establish that the local built-in `search` tool is affected.

This makes a cold first call unreliable and exposes initialization timing to the model/user as a manual-retry requirement.

## Expected behavior

The original tool call should await resource/provider initialization and continue once initialization succeeds, without requiring an identical manual retry.

### Fresh subagent session (confirmed surface)

1. Start from a cold process/cache and launch a fresh subagent such as `codebase-online-researcher`, `debugger`, or `worker`.
2. Immediately make the first tool call to `web_search` or `fetch_content`, before initialization-dependent resources/providers have finished loading.
3. If useful for determinism, introduce artificial slowness in resource/provider initialization.
4. Observe whether the first call fails while loading continues.
5. Retry the identical call and compare the result; the reported behavior is that the retry succeeds.
6. Also issue concurrent cold first calls and check whether they share one initialization operation or race/duplicate loading.

### Fresh parent interactive session (suspected; not yet independently confirmed)

1. Start a fresh parent interactive session from a cold process/cache.
2. Submit a prompt that immediately invokes `web_search` or `fetch_content` during deferred startup.
3. Optionally make initialization deliberately slow, then compare the original call with an identical retry.
4. Exercise concurrent cold first calls to expose any missing single-flight boundary.

The parent-session reproduction is suspected based on the shared lazy/deferred-startup architecture, but has not yet been independently confirmed.

### Confirmed observations

- Fresh subagent sessions.
- Initialization-dependent search/fetch-like tools, most likely `web_search` and `fetch_content`.

### Possibly affected / unverified

- Parent interactive deferred startup.
- `code_search` and `get_search_content`.
- MCP proxy/direct tools and generated `get_<resource>` tools.
- Other lazy extensions whose callable wrappers become visible before all downstream resources are ready.

This issue does not claim that every listed surface is affected; these are investigation targets.

## Relevant code paths

- `packages/web-access/index.ts`
  - `loadHeavy`, `executeHeavyTool`, and `heavyPromise` implement lazy loading for heavy web-access functionality.
  - Lifecycle replay and wrapper registrations are relevant to when the callable tools become visible versus when the loaded extension has received lifecycle state.
- `packages/coding-agent/src/modes/interactive/interactive-input-handling.ts`
- `packages/coding-agent/src/modes/interactive/interactive-deferred-startup.ts`
  - The first-prompt gate/shared startup promise is the parent interactive readiness boundary to verify.
- `packages/coding-agent/src/core/resource-loader-reload.ts`
  - Resource/extension reload and readiness behavior may affect registration and execution ordering.
- `packages/mcp/index.ts`
  - `initPromise` is an existing MCP initialization single-flight boundary.
- `packages/mcp/direct-tools.ts`
  - Direct and resource-backed MCP tool exposure/execution should be checked.
- `packages/mcp/server-manager.ts`
  - `connectPromises` provides per-server connection single-flight behavior.
- `packages/subagents/src/runs/foreground/execution-attempt.ts`
- `packages/subagents/src/runs/shared/pi-args.ts`
  - These child-process launch/argument paths are relevant to reproducing and comparing subagent startup behavior.

## Existing tests and coverage gap

Relevant coverage includes:

- `packages/coding-agent/test/interactive-deferred-startup-first-prompt.test.ts`
- `packages/coding-agent/test/interactive-deferred-startup-input.test.ts`
- `packages/coding-agent/test/interactive-mode-startup-latency.test.ts`
- `packages/coding-agent/test/suite/regressions/1223-startup-lazy-builtins.test.ts`
- `test/unit/workflow-stage-bundled-resources.test.ts`
- `test/unit/mcp-lazy-startup.test.ts`

These tests cover important startup gating, lazy registration, and bundled-resource/tool presence behavior, but do not appear to exercise the full end-to-end first execution path under concurrent calls, deliberately slow loading, and initialization failure followed by retry for both search and fetch/resource-backed tools.

## Technical hypothesis (unverified)

**Hypothesis:** a callable wrapper/tool may become visible before downstream lifecycle replay, provider setup, or resource readiness is complete, or an execution path may bypass an intended readiness boundary. The first execution could therefore observe partial initialization while a later execution sees the completed state.

The current tree already contains shared-promise gates at the obvious boundaries (`heavyPromise`, the interactive deferred-startup promise, MCP `initPromise`, and server `connectPromises`), so the exact race and root cause remain unverified. In addition, a rejected web `heavyPromise` currently appears to be retained rather than cleared. That is a retryability concern, but by itself it would not explain the reported behavior where an identical second call succeeds; the failing boundary may be elsewhere or the observed failure may occur after the heavy module promise resolves.

## Version / environment

The exact released Atomic version, provider, tool configuration, operating environment, and relevant cache/process state were not captured in the original report. These should be collected during triage rather than inferred or invented.

## Acceptance criteria

- [ ] On a cold start, the first invocation of an initialization-dependent tool awaits one shared initialization promise.
- [ ] Concurrent first calls use single-flight loading and do not race or duplicate initialization.
- [ ] After successful initialization, the original tool call resumes and completes without a manual retry.
- [ ] Failed initialization surfaces the real, actionable underlying error and permits a later invocation to retry initialization.
- [ ] Readiness and retry behavior are consistent between parent interactive sessions and subagent sessions.
- [ ] Regression tests cover first-call execution, concurrent first calls, deliberately slow loading, initialization failure/retry, search tools, and fetch/resource-backed tools.
- [ ] Preserve lazy loading; do not eagerly initialize every resource unless investigation proves eager loading is required for correctness.

## Duplicate search

An exhaustive search of the existing open and closed issues found no equivalent report. #1178 concerns a distinct startup/subagent behavior, while #1223 covers lazy built-in startup/registration and supplies relevant regression context but does not track this first-execution initialization race and retry behavior.

w3lld1:
## Summary

- Reset the web-access lazy `heavyPromise` when heavy provider initialization rejects.
- Clear `loadedHeavy` on the same failure path so a later `web_search`/`fetch_content` invocation starts from a clean initialization attempt.
- Add a regression test that simulates a failed first heavy initialization and verifies the next tool call retries and succeeds.
- Add an Unreleased changelog entry for the web-access fix.

Fixes #1704

## Verification

- `bun run --cwd packages/coding-agent vitest --run test/suite/regressions/1223-startup-lazy-builtins.test.ts`
- `bun run typecheck`
- `bun run check:file-length`
- `git diff --check`
- pre-push hook: `bun run lint`, `bun run check:file-length`, `bun run test:unit`

### DIFF
--- a/packages/coding-agent/test/suite/regressions/1223-startup-lazy-builtins.test.ts
+++ b/packages/coding-agent/test/suite/regressions/1223-startup-lazy-builtins.test.ts
@@ -55,6 +55,56 @@ await new Promise((resolve) => setTimeout(resolve, 250));
 	}
 }
 
+function assertWebAccessRetriesFailedHeavyInitialization(): void {
+	const tempDir = mkdtempSync(join(tmpdir(), "atomic-web-access-retry-"));
+	try {
+		writeFileSync(join(tempDir, "package.json"), JSON.stringify({ type: "module" }));
+		writeFileSync(join(tempDir, "index.ts"), readRepoFile("packages/web-access/index.ts"));
+		writeFileSync(join(tempDir, "result-renderers.ts"), "export function renderWebAccessToolResult() { return undefined; }\n");
+		writeFileSync(join(tempDir, "index-heavy.ts"), `
+let attempts = 0;
+export default async function webAccessHeavy(pi) {
+	attempts += 1;
+	if (attempts === 1) throw new Error("simulated heavy startup failure");
+	pi.registerTool({ name: "web_search", execute: async () => ({ ok: true, attempts }) });
+}
+`);
+		const extensionUrl = pathToFileURL(join(tempDir, "index.ts")).href;
+		const script = `
+const { default: webAccess } = await import(${JSON.stringify(extensionUrl)});
+const tools = [];
+const pi = {
+  registerTool(tool) { tools.push(tool); },
+  registerCommand() {},
+  registerShortcut() {},
+  registerMessageRenderer() {},
+  on() {},
+};
+webAccess(pi);
+const webSearch = tools.find((tool) => tool.name === "web_search");
+if (!webSearch) throw new Error("web_search was not registered");
+try {
+  await webSearch.execute({ query: "first" });
+  throw new Error("first execution unexpectedly succeeded");
+} catch (error) {
+  if (!String(error?.message ?? error).includes("simulated heavy startup failure")) throw error;
+}
+const result = await webSearch.execute({ query: "retry" });
+console.log(JSON.stringify(result));
+`;
+		const result = spawnSync("bun", ["--eval", script], {
+			cwd: repoRoot,
+			env: process.env,
+			encoding: "utf-8",
+		});
+		expect(result.status, result.stderr || result.stdout).toBe(0);
+		expect(JSON.parse(result.stdout.trim().split(/\r?\n/).at(-1) ?? "{}"))
+			.toEqual({ ok: true, attempts: 2 });
+	} finally {
+		rmSync(tempDir, { recursive: true, force: true });
+	}
+}
+
 function assertMcpColdStartupTools(options: { withCache: boolean; expectedTools: string[] }): void {
 	const tempDir = mkdtempSync(join(tmpdir(), "atomic-mcp-startup-"));
 	const agentDir = join(tempDir, "agent");
@@ -198,6 +248,10 @@ describe("regression #1223 lazy built-in startup imports", () => {
 
 		assertColdRegistrationDoesNotImportHeavy("packages/web-access/index.ts");
 		assertColdRegistrationDoesNotImportHeavy("packages/intercom/index.ts");
+	});
+
+	it("allows web-access lazy heavy imports to retry after initialization failure", () => {
+		assertWebAccessRetriesFailedHeavyInitialization();
 	});
 
 	it("keeps the MCP proxy fallback until cached direct tools are available", () => {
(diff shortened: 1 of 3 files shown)


flora131:
Thanks for this contribution. The narrow fix to clear the rejected lazy-initializer cache looks good, and the focused regression test passes. We’re going to build broader lifecycle and error-state hardening on top of your commit so your authorship remains in the branch ancestry and you receive credit. We’re treating that follow-up as additional coverage rather than assuming this patch alone fully resolves #1704.

flora131:
Thank you again for the focused lazy-initialization retry fix and regression coverage. A revised integration direction arrived after this PR had already been merged, so GitHub now correctly records #1710 as merged rather than closed-unmerged.

The intended changes from this contribution were incorporated into and extended by #1721: contributor commit `579570134b7d2dfe38d6520f461d690ea014f226` is preserved unchanged in #1721’s branch ancestry, and #1721 adds the broader web-access, Intercom, and MCP lifecycle hardening. This preserves the original authorship and credit while making clear that #1721 contains the final integrated result. We appreciate the contribution and the solid narrow regression case that the broader fix builds on.

PR #1725, opened 2026-07-10
feat(#1636): surface transitive session spend in status bar

### LINKED ISSUE #1636: No insight into transitive cost
The dollar amount in the status bar only measures the tokens used by the top-level agent.  Tokens used by subagents in workflows don't seem to be included in this amount, giving an unrealistic picture of the token spend.  I would love to see the full transitive cost in the status bar, including all subagents and subsubagents.

gebner:
Report full transitive session spend in the interactive footer: the current session plus subagents, nested subagents, and workflow-stage descendants. The footer dollar amount and token/cache badges now use transitive totals, while the context-window percentage remains current-session only. Remove the always-on `~` cost marker because provider accounting is inherently slightly delayed.

Add a transitive usage aggregator that accepts live descendant rollups, walks persisted descendant session trees for reconciliation, aliases live and durable reports by session file, excludes inherited fork transcript usage, and uses revision snapshots so stale persisted-session walks cannot delete or overwrite newer live reports.

Wire delegated spend through all active execution layers:
- subagents emit foreground, async, and live foreground-update usage rollups;
- workflow stages persist usage snapshots and emit interim/final stage rollups;
- workflow stages forward nested descendant updates from subagents/workflows;
- extension runtimes expose root session identity and usage rollup ports;
- /cost reconciles descendant session files and shows self/descendant/total breakdowns.

Update docs, changelogs, and tests for subagent/workflow transitive usage, live workflow reporting, stale-reconciliation protection, and footer rendering.

Validation before squash/rebase: typecheck, lint, check:file-length, focused
https://github.com/bastani-inc/atomic/issues/1636 tests pass. Full unit suite currently has two OAuth lifecycle failures that also reproduce on clean origin/main.

Fixes #1636


lavaman131:
Hey @gebner, thanks for the contribution. Will take a look later tonight and provide feedback or merge it if things look good. Currently focusing on the p0 errors you reported with GitHub Copilot API and Claude, so a stable pre-release and release can get shipped. Then, we can include this in the next pre-release.

gebner:
FWIW, this change here is a lot tricker than I'd anticipated.  I just found another issue.  Will mark ready for review once it's got a bit more testing.

gebner:
Apparently the GPT-5.6 change on main broke cost reporting.  This is all fixed now and ready for review!

lavaman131:
Thank you for the substantial work on transitive spend reporting here. I completed an independent full-branch review against current `main` and confirmed four P1 correctness gaps in the current PR revision:

1. incomplete/stale reconciliation can replace a broader known live rollup and reduce reported descendant usage;
2. malformed or unavailable inherited-parent session data can be treated as exact (and can overcount fork-inherited usage), so incomplete totals are not always safe lower bounds;
3. Windows session-file aliases/subtree coverage are case- and separator-sensitive, allowing duplicate accounting;
4. async workflow/subagent usage can be attributed to the session active at completion rather than the root session that launched it, with slash-command and resume paths still missing launch-root propagation.

I opened #1750 as a replacement/follow-up based directly on this PR. It preserves the original implementation commit and authorship, is rebased onto current `main`, adds the confirmed corrections and regressions, and documents the validation performed (including 3,257 passing unit tests). No merge action has been taken.

flora131:
Closing this PR as #1750 is now the follow up where changes are being tracked

PR #1726, opened 2026-07-10
fix(workflows): surface resumable runs on session restore

### LINKED ISSUE #1722: Resuming a session via --session does not re-attach/surface its running background workflow
## Summary

Resuming a chat session with `atomic --session <id>` restores the conversation transcript but silently drops the live/durable **background workflow** (e.g. a running `ralph` loop). The workflow does not re-attach or re-connect, there is no banner or hint, and the user is left thinking the background worker "disappeared."

## Environment

- Reopening a session in a different terminal/space manager (in my case [Herdr]; also applies to tmux/new tab) via `atomic --session <uuid>`.
- Default file-backed durable workflow backend (`~/.atomic/workflow-durable`).

## Repro

1. In session A, start a background `ralph` workflow; let it run.
2. Close session A's tab/terminal (e.g. to move it into tmux/Herdr).
3. In a new terminal, reopen the same chat: `atomic --session <session-A-uuid>`.
4. The transcript comes back, but the ralph run does **not** appear or reconnect in the new session.

## Expected

On `--session` resume, if the session references a resumable/live durable workflow, Atomic should surface it — auto-connect the overlay (as `/workflow resume` does), or at minimum print a banner like: *"This session has a resumable workflow `<run-id>` (ralph). Run `/workflow resume <run-id>` to reconnect."*

## Actual

Nothing. No banner, no auto-connect, no pointer to `/workflow resume`. The running workflow is effectively invisible in the resumed session.

## Evidence / root cause

The chat session only stores a **discovery index** pointer (`customType: "workflow.durable.checkpoint"`), while the run itself lives in the separate durable store with its own session lineage:

- Durable store: `~/.atomic/workflow-durable/workflow-<uuid>.json` → `handle.status: "running"`, `handle.name: "ralph"`, with its own stage `sessionId`/`sessionFile` values distinct from the parent chat session.
- The parent chat session references the workflow `runId`, but `--session` resume neither re-dispatches nor connects it.
- `status:"running"` in the durable file goes stale the moment the original tab closes (last `updatedAt` stops advancing), confirming the live process ends with the tab while the run remains only *durably resumable*.

## Workaround

Inside the reopened session:

```
/workflow status                  # find the run-id
/workflow resume <run-id>         # replays checkpoints and auto-connects the overlay
```

(`/workflow connect <run-id>` if the run is still live.)

## Suggested fix

Make `--session` / `--continue` / `-r` resume detect referenced resumable durable workflows for that session and either (a) auto-connect the overlay like a successful `/workflow resume`, or (b) show a one-line actionable banner with the exact `/workflow resume <run-id>` command.

w3lld1:
## Summary

- detect resumable durable workflow references when an existing Atomic session starts or is resumed
- show each workflow name, short ID, and exact `/workflow resume <id>` command
- ignore completed, cancelled, and explicitly non-resumable workflows, using the latest cached state for each workflow

## Validation

- `/tmp/bun-runtime/bun-linux-x64/bun test test/unit/resumable-workflow-notices.test.ts` (2 passed)
- `git diff --check`

The full workspace typecheck and test suite were not run locally because this runner did not have the repository dependencies installed and had insufficient disk space for a full `bun install`.

Fixes #1722

### DIFF
--- a/packages/workflows/src/extension/extension-lifecycle.ts
+++ b/packages/workflows/src/extension/extension-lifecycle.ts
@@ -4,6 +4,7 @@ import { cancellationRegistry } from "../runs/background/cancellation-registry.j
 import { stageControlRegistry } from "../runs/foreground/stage-control-registry.js";
 import { store } from "../shared/store.js";
 import { restoreOnSessionStart } from "../shared/persistence-restore.js";
+import { findResumableWorkflowNotices } from "../shared/resumable-workflow-notices.js";
 import { installCompactionHook } from "../shared/persistence-compaction-policy.js";
 import { clearForms } from "../tui/inline-form-store.js";
 import { installStoreWidget } from "../tui/store-widget-installer.js";
@@ -57,7 +58,7 @@ export function registerWorkflowLifecycleHandlers(
     return { cancel: true };
   });
 
-  pi.on("session_start", async (_event, ctx) => {
+  pi.on("session_start", async (event, ctx) => {
     runtimeState.resetWorkflowDiscoveryForSession();
     deAdvertiseAskUserQuestionWhenHeadless(pi, ctx?.hasUI);
     await runtimeState.ensureWorkflowConfigLoaded();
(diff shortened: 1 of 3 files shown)


lavaman131:
Thank you for adding the restored-session resume notice. During follow-up review, I found one eligibility gap: failed workflows with absent `resumable` metadata were omitted, even though durable workflow semantics treat failed (and blocked) runs as resumable unless they are explicitly marked `resumable: false`. A follow-up based on this work centralizes the eligibility rule and adds regression coverage: #1748. It supersedes/corrects this PR without changing its core intent.

w3lld1:
Thanks for the follow-up and for preserving the core intent in #1748. Closing this superseded PR to keep the review queue clear.

PR #1859, opened 2026-07-18
feat(workflows): show model + thinking level on graph node cards

sina85:
## What this does

Shows the effective **model + thinking level (and Codex fast tier)** for each workflow stage on the `/workflow connect` graph node cards, and persists that identity across `/workflow resume`. The main-session footer has always shown `<model> <thinking>`, but background/orchestrated runs were opaque about which model (and reasoning level) each stage used. This closes that gap in the place you inspect a running workflow — the graph overlay.

> **Scope note:** an earlier revision of this PR also added the same display to the above-editor `BACKGROUND` widget. Per review that has been **removed** — the widget summarizes each run on a single line, so a parallel fan-out on several models has no clean way to show them. Model/thinking/fast identity lives only on the graph node cards, which have one card per stage and room to show each. This PR now has **zero net diff to `widget.ts`**.

## 1. Graph node cards (`/workflow connect`)

Node cards showed title / duration / status / deps but deliberately hid the model. They now carry a dedicated compact model row beneath the status line:

```
╭───── specialist ─────╮  ╭─────── scout ────────╮  ╭───── codex-fix ──────╮
│         12s          │  │          8s          │  │          8s          │
│      ● running       │  │      ● running       │  │      ● running       │
│claude-opus-4.8 · high│  │        gpt-5         │  │  gpt-5.1-codex fast  │
│         root         │  │         root         │  │         root         │
╰──────────────────────╯  ╰──────────────────────╯  ╰──────────────────────╯
```

- Provider prefix dropped and thinking omitted when `off` to fit the ~22-cell card; `—` shows when no model is resolved yet.
- **Codex fast tier** is appended on the model row, mirroring the footer's `formatCodexFastModeModelLabel` (`<model> <thinking> fast`). When the card would overflow, the thinking level is dropped so the `fast` marker is never truncated away.
- **Live model fallbacks are reflected:** when a stage falls back to another model the row updates to the model actually running (it renders from the live `StageSnapshot.model`, which `applyModelFallbackMeta` rewrites on each fallback).
- Cards grow from 5 → 6 rows (`NODE_H`); graph layout and edge routing derive from `NODE_H` and adapt automatically.

## 2. Persistence across `/workflow resume`

`thinkingLevel` now rides alongside `model` at every serialize/restore site of the mandatory DBOS durable path — the `DurableStageCheckpoint` type, stage/task checkpoint metadata, the DBOS envelope encode/validate/decode, the completed-run catalog merge, checkpoint→snapshot hydration, and the replay `__modelFallbackMeta` (plus `WorkflowTaskResult`). A resumed run therefore restores the same model + thinking identity, not just the model.

## Plumbing (shared live-snapshot path)

- `StageModelFallbackMeta` gains `thinkingLevel`, populated by `StageSessionController.currentModelFallbackMeta()` from the live session (or a pending level set before the session exists).
- `executor-stage-factory`'s `applyModelFallbackMeta` writes it to `StageSnapshot.thinkingLevel`, next to the existing `model` field.
- The node card renders the compact `<short-model> · <thinking> [fast]` row via a small local `codex-fast-label.ts` — a mirror of the footer helper, kept local because importing the `@bastani/atomic` barrel into the TUI module graph breaks the `pi-tui`-mocked overlay test subprocesses.

## Contract change

The node cards previously **hid** the model by design; this reverses that. The three tests pinning the old behavior are updated to assert the model row is shown, and the dependency-row assertions move down one row.

## Tests

- `node-card.test.ts`: compact model row; thinking append/omit; fast marker on the model row (not the deps row); width-guard (drop the level to preserve `fast`); `—` when absent; mid-run fallback swap; geometry (`NODE_H` = 6).
- `durable-dbos-topology.test.ts`: model + thinking-level round-trip through the DBOS envelope (encode → decode).
- `durable-stage-frontier-fixes.test.ts`: cached-replay hydration restores `thinkingLevel` (checkpoint → snapshot).
- `stage-runner-reasoning-suffix.test.ts`: a 429 fallback surfaces the fallback model and thinking level on `meta`.
- `bun run typecheck` / `lint` / `check:file-length` clean; the `overlay-graph-*` and `durable-*` suites pass.

## Notes

- Rebased onto the latest `main`.
- The only `test:unit` failures on this branch (MCP lazy-startup ×3 and the `Pi v0.81.1 declarations and publish artifacts` sync check) reproduce identically with these changes stashed on a clean tree — pre-existing and unrelated.

<h2><a href="https://app.greptile.com/api/retrigger?id=45278373"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a>Confidence Score: 5/5</h2>

Safe to merge with low risk.

<h2>Findings</h2>

1. P1&nbsp;**Widget path missing** <a href="https://github.com/bastani-inc/atomic/pull/1859#discussion_r3626857311">▶</a>

<details><summary>Fix with agent prompt</summary>

`````markdown
### Issue 1
packages/workflows/src/shared/store-types.ts:215-219
`StageSnapshot.thinkingLevel` is documented for background-run surfaces here, but `packages/workflows/src/tui/widget.ts` still builds its meta line only from `mode`, progress, and elapsed time (`metaLine`/`themedRunLines`), so the `BACKGROUND` widget never displays the model or thinking level promised by this PR. This leaves direct background runs opaque even when `stageSnapshot.model` and `thinkingLevel` are populated.

---

`````

</details>

<h3>Summary</h3>

- Adds a compact model and thinking-level row to `/workflow connect` graph cards.
- Preserves canonical `-fast` model suffixes and omits `off` thinking in narrow card rows.
- Threads `thinkingLevel` through live fallback metadata, durable checkpoints, DBOS envelopes, completed catalogs, and replay paths.
- Updates docs and tests for the new task result field, resume behavior, fallback updates, and taller node-card geometry.

<details open><summary><h3>Diagram</h3></summary>

```mermaid
sequenceDiagram
participant Runner as Stage runner
participant Snapshot as StageSnapshot
participant Card as Graph node card
participant Durable as Durable checkpoint/DBOS
participant Replay as Resume replay

Runner->>Snapshot: apply model + thinkingLevel fallback metadata
Snapshot->>Card: render compact model row
Snapshot->>Durable: encode model + thinkingLevel metadata
Durable->>Replay: decode/hydrate checkpoint
Replay->>Snapshot: restore replayed model + thinkingLevel
Snapshot->>Card: render restored identity after resume
```
</details>

<sub>Reviews (20) · Last reviewed commit: ["merge: refresh PR #1859 with latest main"](https://github.com/bastani-inc/atomic/commit/b86d867edfad2c59c4907bae8e098d93d3e4d484)</sub>


lavaman131 (changes requested):
Hey @sina85, thanks for the contribution. This looks like a very cool idea! For the background widget change can you make sure that parallel workflow stages with different models are intuitive to see? I think the graph node cards are cleaner since they can each show the model, but the background widget gets messy. Also, can you make sure you properly handle model fallbacks and change the model in the node card when that happens? Another edge case is to make sure you display (fast) for the model if using the fast tier for OpenAI providers. Finally, you need to make sure that when you run `/workflow resume` that this data is saved. If you give evidence for each of these will review again and we can look into getting this checked in for the next release.

sina85:
Hi @lavaman131, thanks for the feedback, I will double check the edge cases and get back to you.

sina85:
@lavaman131 thanks for the detailed review.

**Update (edited):** after the first pass I reconsidered your parallel point and **removed the model/thinking display from the `BACKGROUND` widget entirely** — you were right that a single per-run line gets messy for a parallel fan-out, and the widget isn't the place for it. That identity now lives **only on the `/workflow connect` graph node cards** (one card per stage, so each shows its own model). The PR now has **zero net diff to `widget.ts`**. Evidence for the node-card work + persistence below; all four of your points are addressed (point 1 by removal).

### 1. Parallel stages → handled by the graph node cards, not the widget
Per your observation, the widget was the wrong surface. Each stage gets its own node card, so parallel stages on different models are each shown cleanly, no cramming:

```
╭───── specialist ─────╮  ╭─────── scout ────────╮  ╭───── codex-fix ──────╮
│         12s          │  │          8s          │  │          8s          │
│      ● running       │  │      ● running       │  │      ● running       │
│claude-opus-4.8 · high│  │        gpt-5         │  │  gpt-5.1-codex fast  │
│         root         │  │         root         │  │         root         │
╰──────────────────────╯  ╰──────────────────────╯  ╰──────────────────────╯
```

### 2. Model fallbacks update the node card
The card renders from the live `StageSnapshot.model`, which `applyModelFallbackMeta` rewrites on every fallback, so the displayed model swaps to the model actually running. Proven on both halves of the chain:
- **controller → meta:** `stage-runner-reasoning-suffix.test.ts` now asserts that after a 429 fallback `meta.model === "openai/fallback"` (the fallback, not the failed primary).
- **render:** `node-card.test.ts` → *"model row reflects a fallback swapping the stage's model mid-run"* renders the same snapshot before/after the swap.

### 3. Fast tier shown with the model
Mirrors the footer's `formatCodexFastModeModelLabel` (`<model> <thinking> fast`). On the node card the fast marker sits on the model row (off the deps row); when the ~22‑cell card would overflow, the thinking level is dropped so the marker is never truncated:

```
fast tier on model row            width-guard (·high dropped, fast kept)
╭───── specialist ─────╮          ╭───── specialist ─────╮
│         12s          │          │         12s          │
│      ● running       │          │      ● running       │
│  gpt-5.1-codex fast  │          │  gpt-5.1-codex fast  │
│         root         │          │         root         │
╰──────────────────────╯          ╰──────────────────────╯
```

Tests: node‑card *"shows the fast tier on the model row, not the deps row"*, *"…keeps both the thinking level and the fast marker when they fit"*, *"…drops the thinking level to preserve the fast marker when the card would overflow"*.

### 4. `/workflow resume` persists the data
Promoted the deferred follow‑up into this PR. `thinkingLevel` now rides alongside `model` at every serialize/restore site of the mandatory DBOS durable path — the `DurableStageCheckpoint` type, stage/task checkpoint metadata, the DBOS envelope encode/validate/decode, the completed‑run catalog merge, the checkpoint→snapshot hydration, and the replay `__modelFallbackMeta` (plus `WorkflowTaskResult`). Tests:
- `durable-dbos-topology.test.ts` → *"round-trips stage model + thinking-level metadata through the DBOS envelope"* (encode → decode).
- `durable-stage-frontier-fixes.test.ts` → the cached‑replay hydration test now asserts `stage.thinkingLevel` survives checkpoint → snapshot.

### Verification
- `bun run typecheck` / `bun run lint` clean; `bun run check:file-length` passes. `widget.ts` and its test are reverted to upstream (zero widget diff); the shared `codex-fast-label.ts` stays for the node card, mirroring `formatCodexFastModeModelLabel` locally so the `@bastani/atomic` barrel never enters the TUI module graph (importing it breaks the `pi-tui`‑mocked overlay test subprocesses).
- Full `test:unit` is green apart from 4 pre‑existing failures (MCP lazy‑startup ×3 and the `Pi v0.81.1 declarations and publish artifacts` sync check), which reproduce **identically with these changes stashed on a clean tree** — unrelated to this PR.
- Rebased onto the latest `main`.

lavaman131 (dismissed):
Hey @sina85, can you address the review bot (Greptile) feedback and rebase with the latest origin/main. Otherwise, once you do that, looks great and we can check it in pending CI is green.

flora131:
Sorry for the delay in responding. I just had a chance to check this after the latest commits. I really appreciate the contribution.

I verified the exact PR head (`23eb0c23eb5881c102b78585387c20979d702d62`) using the local Atomic CLI in tmux and found three blockers:

1. **The new replay test suite does not load under the repository’s test runner.**

   `test/unit/executor-replay-model-identity.test.ts:14` imports `describe` from `bun:test`, but root unit tests run under Node/Vitest.

   Repro:

   ```sh
   npx vitest --run --project unit \
     test/unit/node-card.test.ts \
     test/unit/executor-replay-model-identity.test.ts \
     test/unit/durable-dbos-topology.test.ts \
     test/unit/durable-stage-frontier-fixes.test.ts \
     test/unit/stage-runner-reasoning-suffix.test.ts \
     test/unit/overlay-graph-perf.test.ts
   ```

   Result:

   ```text
   Cannot find package 'bun:test'
   ```

   Please use the repository’s Vitest imports so the two new replay tests execute.

2. **A long configured Codex model name truncates the required `fast` marker.**

   Rendering `openai-codex/gpt-5.3-codex-spark` in fast mode produces:

   ```text
   │gpt-5.3-codex-spark f…│
   ```

   The model comes from the configured catalog. The card should reserve enough room for the full `fast` marker, truncating the model name first if needed. Please add a regression test for this long-name case.

3. **The changed files fail the repository’s Biome gate.**

   `npm run check` reports five formatting/import errors in the new or modified files, including indentation, import order, and comment spacing. Please format the branch and rerun the check.

The core behavior otherwise worked in manual testing: model and thinking rows appeared in `/workflow connect`, `off` was omitted, unresolved identity showed `—`, the six-row layout remained readable, and same-process `/workflow resume` retained the model and thinking level.

Please rerun the focused tests and `npm run check` after addressing these items.

flora131:
By the way, you can also ask Atomic to verify your PR directly. For example:

> Can you run a workflow to verify this PR? Check out the exact PR head in an isolated worktree, run that checkout’s local Atomic CLI inside a tmux session, and manually test the behavior covered by the PR. Please run the relevant automated checks as well. If you find any bugs, document each one with reproduction steps, expected and actual behavior, and return the findings inline when the workflow completes.
>
> PR: https://github.com/bastani-inc/atomic/pull/1859


You can ask to attach evidence to this PR or even change the prompt to automatically apply the fixes.

sina85:
> By the way, you can also ask Atomic to verify your PR directly. For example:
> 
> > Can you run a workflow to verify this PR? Check out the exact PR head in an isolated worktree, run that checkout’s local Atomic CLI inside a tmux session, and manually test the behavior covered by the PR. Please run the relevant automated checks as well. If you find any bugs, document each one with reproduction steps, expected and actual behavior, and return the findings inline when the workflow completes.
> > PR: #1859

Thanks for replying, I will run those changes and update the PR.

sina85:
@flora131 thank you — all three were spot on, and the `bun:test` one in particular was a real gap. Fixed in `15ac7c14`.

### 1. Replay test now loads under Vitest
`test/unit/executor-replay-model-identity.test.ts:14` imported `describe` from `bun:test`; switched it to `vitest`. It runs and passes under the repo runner now.

### 2. Long model name truncates first, `fast` marker is preserved
`modelText` now drops the thinking level first, then truncates the **model name**, reserving the whole ` fast` marker. Actual render of `openai-codex/gpt-5.3-codex-spark` in fast mode:

```
before:  │gpt-5.3-codex-spark f…│
after:   │gpt-5.3-codex-sp… fast│
```

Short names are unchanged (`│  gpt-5.1-codex fast  │`). Added a regression test: *"keeps the full fast marker and truncates a long model name instead of the marker."*

### 3. Biome gate is clean
Ran `biome check --write` (tabs in `codex-fast-label.ts`, import order in `node-card.ts` and the test). `npm run check` now passes.

### Verification (the exact commands you gave)
```
$ npm run check
Checked 2340 files … No fixes applied.     # biome
tsc --noEmit                                # clean
npm-shrinkwrap.json is up to date.          # shrinkwrap
→ exit 0

$ npx vitest --run --project unit \
    test/unit/node-card.test.ts \
    test/unit/executor-replay-model-identity.test.ts \
    test/unit/durable-dbos-topology.test.ts \
    test/unit/durable-stage-frontier-fixes.test.ts \
    test/unit/stage-runner-reasoning-suffix.test.ts \
    test/unit/overlay-graph-perf.test.ts
 Test Files  6 passed (6)
      Tests  68 passed (68)
```

Also appreciate the tip on driving Atomic to verify a PR in an isolated worktree/tmux — I'll use that flow next time so the runner/gate mismatch doesn't slip through again. Ready for another look when you have a chance.

flora131:
Recorded a hands-on run of this PR on a local build of the head commit (15ac7c14c3441ad399b9d08f17ba961f2f531edd), driving the Atomic TUI in a tmux session and inspecting the /workflow connect graph node cards. Attaching the screen recording of that session for reference.


https://github.com/user-attachments/assets/8ec28226-5ede-4702-81ff-62519cfb5278

flora131:
Thanks for the fix here. While testing this PR's exact head (`15ac7c14c3441ad399b9d08f17ba961f2f531edd`) in the real Atomic TUI, I found one issue with the new model/thinking-level row:

**Thinking levels can be truncated or omitted for normal model IDs.**

Repro:
1. Run workflow stages with models such as `anthropic/claude-haiku-4-5:high` and `anthropic/claude-sonnet-4-5-20250929:medium`.
2. Open the workflow graph with F2.
3. Inspect the model row on each stage card.

Expected: the effective model and full thinking level remain identifiable.

Actual:
- `claude-haiku-4-5 · high` renders as `claude-haiku-4-5 · hi…`.
- The longer Sonnet model renders as `claude-sonnet-4-5-202…`, with `medium` omitted entirely.

It looks like the 22-cell card interior applies the model-row width budget only in fast mode; otherwise the generic card truncation can consume the thinking level. The attached test video shows this around 1:24–1:40.

Are you able to reproduce this as well? If so, could the row reserve space for the thinking level before truncating the model name?

sina85:
@flora131 reproduced it exactly — thanks, and good catch. My earlier fix only reserved space in **fast** mode, so for ordinary models the generic 22-cell truncation ate the thinking level. Fixed in `6cb25927`.

`modelText` now treats the thinking level and the ` fast` marker as one **load-bearing suffix** and always truncates the **model name** first, in every mode (not just fast). Actual renders on the exact head:

```
before                          after
claude-haiku-4-5 · hi…          claude-haiku-4… · high
claude-sonnet-4-5-202…          claude-sonne… · medium
```

And the combined + short cases stay correct:

```
gpt-5.1-c… · high fast     (fast + level overflow: both kept)
gpt-5.3-codex-sp… fast     (fast, long name: marker kept)
claude-opus-4.8 · high     (fits: unchanged)
gpt-5.1-codex fast         (fits: unchanged)
```

Added two regression tests: the non-fast long-name case (`· high` / `· medium` survive) and the combined level+fast overflow.

Verified with the real gates:
```
$ npm run check      → exit 0  (biome clean · tsc clean · shrinkwrap up to date)
$ npx vitest --run --project unit  <the six focused files>
 Test Files  6 passed (6)
      Tests  70 passed (70)
```

Ready for another look. (Also — appreciate you recording the TUI run; that's a much better signal than my earlier headless checks, and it caught exactly what I missed.)

flora131:
Thanks, Sina — we’ll take care of rebasing this onto current `main` and then run a formal merge-readiness code review, including the updated checks and the model-row persistence paths. We’ll follow up here with the review result and any remaining changes.

flora131:
Thanks, Sina — this is all good from our formal review now.

We rebased the branch onto the latest `main`, preserved your commit authorship, added the missing `packages/workflows` changelog entry, and added the PR link and your `@sina85` attribution to the coding-agent changelog. We then pushed the reviewed head back to your branch.

The rebased head passes `npm run check`, all 94 focused workflow tests, the full unit suite (6,762 passed, 2 skipped), documentation link validation, and the full pre-push hook set, including the integration and CI-contract suites. GitHub CI will rerun on the updated head.

The 0.9.14 prerelease train has already been cut and is next for production, so we’ll prepare this change for the following 0.9.15 release. Thank you for the contribution and for working through the earlier review feedback.

flora131:
Sorry for the delay getting back to you. We missed carrying this through after our formal review, and the 0.9.15 release shipped without it. Current `main` still does not show or persist a stage's thinking level, so the feature remains absent.

We'll take ownership of the next step rather than ask you to rebase again. We'll reapply the reviewed change onto current `main`, rerun the relevant checks, and merge it if it remains clean. If intervening changes require your input, we'll call that out here.

Thanks again for the thorough implementation and for your patience.

flora131:
Maintainer update: we’ve rebased and pushed the reviewed changes at `1b425a90663a8001e81aee440a0b410899abde3a`.

What we did:
- Reconciled the overlapping implementation, documentation, and tests with main while preserving all 14 contributor commits and their author/coauthor metadata.
- Adapted the graph-card display to the current `-fast` model selection rather than restoring the removed fast-mode toggle. The feature remains confined to graph cards, not the BACKGROUND widget.
- Fixed a rebase regression where resumed task results lost `thinkingLevel`, even though checkpoints and graph cards retained it. Added regression coverage for replay and optional-field behavior.
- Restored the `thinkingLevel` field in the public task-result documentation example and added a documentation regression test.
- Reconciled a test-fixture compatibility conflict without weakening the existing guard.

Validation: 258 focused/related tests and `npm run check` passed. We also captured 24 real Atomic CLI/tmux scenarios covering model/thinking display, fallback, narrow layouts, navigation, and restart/resume persistence. Those runtime scenarios used a deterministic fixture provider, not live vendor model calls.

Separate findings—initial test database isolation, cold-resume initialization, and a macOS fixture portability issue—are documented in our local `ISSUES-1859.md` handoff rather than folded into this feature. The initial isolation incident is retained in the record; subsequent accepted runtime verification used an explicitly private database.

CodeQL passed at the latest check. We are handling the known unrelated unit/integration CI failures separately; this is not a claim that all CI is green. The PR has not been merged.

PR #2048, opened 2026-07-28
fix(compaction): bound range planner recovery

elefthei:
## Summary

- cap first-pass Verbatim Compaction range-planner reasoning at `low`
- retry exactly once at `minimal` reasoning when a length-stopped response has no recoverable usable ranges
- retain whole-region planning, transport retries, partial-range recovery, diagnostics, cache-retention isolation, and fresh request session IDs
- add focused planner, diagnostic, post-tool preflight, and continuation replay regressions
- update the compaction contract and changelog

## Validation

On the `0.9.11-alpha.5` hotfix line before integration into current `main`:

- `bun run vitest --run test/verbatim-compaction-planning.test.ts test/truncated-range-recovery.test.ts test/compaction-diagnostics.test.ts test/post-tool-compaction-preflight.test.ts` — 100 passed, 2 skipped
- `bun test test/unit/executor-continuation-structured-task.test.ts` — 1 passed
- `bun run typecheck` — passed
- coding-agent documentation link validation — passed

After integration onto current `main`, dependency hydration was blocked by an Azure Artifacts HTTP 401 for the updated Pi 0.82.1 packages, so the main-line typecheck and harness rerun could not be completed locally.

---
<a href="https://app.blacksmith.sh/bastani-inc/codesmith/atomic/pr/2048"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787795911&installation_model_id=10562&pr_number=2048&repository=bastani-inc%2Fatomic&return_to=https%3A%2F%2Fgithub.com%2Fbastani-inc%2Fatomic%2Fpull%2F2048&signature=0c3559ec249fa39311247215c406808b669523b7e680023e71b2acc72d0b6554"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
[...]

PR #2050, opened 2026-07-28
fix(compaction): port codex's fallback rungs so compaction cannot kill a turn

flora131:
## Summary

Compaction could hard-fail — and during a post-tool preflight, kill the active turn — when the range planner returned no usable ranges. The reported trigger was an inherited `high` reasoning level letting `gpt-5.6-sol` spend the planner's entire ~13K output budget on reasoning. A rate limit outlasting the retry budget did the same thing.

The root cause was inherited from pi and is two lines: an output cap of `0.8 × reserveTokens` (13,107) that reuses an **input-side** reserve as an output cap, plus forwarding the live session reasoning level into it. On every current reasoning model — `gpt-5.6-sol` and every adaptive-thinking Claude alike — reasoning is drawn from that same cap.

This ports `openai/codex`'s compaction ladder, substituting verbatim line planning for summary generation. Three rungs:

1. **`planned`** on the session model — verbatim line planning
2. **`planned` on a borrowed fallback model** — reuses `settings.fallbackModels` for one planner request, without switching the session's model
3. **`fresh`** — `startNewContextWindow`, a total model-free rung porting `compact_token_budget`

Rungs 1–2 preserve quality and run at any urgency. Only rung 3 destroys context, so only it requires `load_bearing` urgency — `/compact` still fails honestly rather than silently clearing your context.

Design doc: [`specs/2026-07-27-compaction-fallback-rungs.md`](specs/2026-07-27-compaction-fallback-rungs.md). §4.3 traces every element to its codex mechanism; §9.3 lists the five deliberate divergences.

## Why not just cap the reasoning level

That was the first attempt (#2048). It treats the symptom: `low` is a constant rather than a bound, it retries on any length stop, it leaves the invented cap in place, and it does nothing for rate limits. Codex sets **no** output cap at all — `max_output_tokens` does not exist in its `ResponsesApiRequest`. Removing the cap is what makes inheriting the reasoning level safe, which is why codex can inherit it too (`compact.rs:699`).

Notably there is **no** starvation retry here. Codex never modifies reasoning effort across compaction attempts, and a reasoning-only response is not even an error there — it flows through to its `"(no summary available)"` placeholder. An earlier draft of this design invented an `off`-reasoning retry; it was removed for fidelity.

## What this does not do

The strict zero-usable-ranges rejection is **unchanged**. pi has this same defect but hides it: it never validates its summary and appends a deterministic file manifest, so a totally failed compaction produces a plausible non-empty artifact and destroys context silently. That validation is why this bug was reportable at all, and softening it was never on the table.

## Breaking change

`runVerbatimCompaction` is a published export (`src/index.ts`). Positional args collapse into one request object, and pre-resolved credentials become a `resolveAuth(model)` callback, because a borrowed fallback model needs its own credentials rather than the session model's.

```ts
// before
await runVerbatimCompaction(prep, model, apiKey, headers, signal, thinkingLevel, options);

// after
await runVerbatimCompaction(prep, model, {
  resolveAuth: options.resolvePlannerAuth,
  signal,
  urgency: "recoverable",   // "load_bearing" for overflow recovery / post-tool preflight
  thinkingLevel,
  ...options,
});
```

Everything else is preserved: settings and defaults, the persisted entry shape (`details.rung` widens additively), resume behavior, extension events, and `settings.retry`.

`settings.fallbackModels` has its scope widened — documented as main-chat-scoped, it is now also borrowed for compaction. **A configured fallback model may therefore receive the compaction transcript.** Called out in the docs and changelog.

## Validation

| Gate | Result |
|---|---|
| `bun run test:unit` | 4343 pass, 1 skip, 0 fail (572 files) |
| `bun run test:integration` | 287 pass, 0 fail (33 files) |
| `bun run typecheck` | clean |
| `bun run check:file-length` | passed (2490 tracked files) |

Plus real end-to-end evidence from a live tmux-driven session against `anthropic/claude-opus-5` at thinking level `high` — **posted as a comment below**. The captures are deliberately kept in the PR comment rather than committed to the tree, matching how `.gitignore` already excludes dated `research/` artifacts.

The load-bearing test is borrowing purity: after a compaction rescued by a fallback model, `session.model`, `session.thinkingLevel`, the session entry list and `_fallbackAttemptedKeys` are all identical to their pre-compaction values, with no `model_changed` / `model_select` / `model_fallback_start` event and no `agent.continue()` call. A user's session model must not silently change because compaction got throttled.

## Reviewer notes

- **§9.3 divergence 3 (`CompactionUrgency`) is the one to push on.** It is the only element that adds a type and a parameter rather than removing something. The argument: codex's fresh-window rung is either always-on under a feature flag or absent — it is never reached as a *failure fallback* — so without urgency, Atomic's manual `/compact` would clear context on failure, which codex never does.
- **Pre-existing accounting quirk, not introduced here.** A compaction on an already-small region can report negative `percentReduction` (observed: `tokensBefore=3126`, `tokensAfter=9869`). `tokensBefore` is the whole-context estimate while `tokensAfter` adds the kept-tail estimate, so the two are not like-for-like. The arithmetic in `withWholeContextStats` predates this change; flagged for a separate fix rather than folded in here.
- The spec and research commits are on this branch and appear in the diff. They are the design record, not implementation under review.

<details>
<summary>Rung ladder</summary>

```
request to model M
  ├─ transient error (429 / overloaded / 5xx)
  │     → backoff retry, SAME model (existing retryAssistantCall)
  ├─ context overflow
  │     → drop oldest region lines, retry SAME model   (codex history.remove_first_item)
  └─ bad plan (starved / malformed / no usable ranges)
        → M is done, no second attempt                 (codex has no branch here)

M done → borrow next fallbackModels entry → repeat
       → list exhausted → fresh window (load_bearing) | RangePlanError (recoverable)
```

</details>

---
<a href="https://app.blacksmith.sh/bastani-inc/codesmith/atomic/pr/2050"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787840922&installation_model_id=10562&pr_number=2050&repository=bastani-inc%2Fatomic&return_to=https%3A%2F%2Fgithub.com%2Fbastani-inc%2Fatomic%2Fpull%2F2050&signature=65babbb5838c6172a03570c18f7eab764453a6bb245401cb8fb04c83ce406865"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


flora131:
## E2E evidence — live tmux session against a real provider

This comment **is** the evidence of record. The full capture file was 871 lines / 62 blocks across four sessions (the live run was repeated after each review-repair iteration rather than assumed to still hold); it is deliberately not committed to the tree, since it is a snapshot of one machine's terminal scrollback with absolute local paths — the same category `.gitignore` already excludes for dated `research/` artifacts. The decisive excerpts are reproduced below verbatim.

Every block there is verbatim `tmux capture-pane` output or verbatim session JSONL. Nothing reconstructed or hand-written.

| Fact | Value |
| --- | --- |
| Command under test | `bun packages/coding-agent/src/cli.ts` (real CLI from this worktree, real TUI) |
| Provider / model | `anthropic` / `claude-opus-5`, thinking level `high` |
| tmux | 3.7b, 200x50 |
| Scope | **success paths only** — no faked 429 or reasoning-only truncation |

### Real work crosses the threshold and compacts mid-turn

```text
 Compacted from 47,029 tokens (ctrl+o to expand)

                                        ↑14 • ↓353 • R189k • W31k • CH60.3% • $0.296 (sub) • 4.7%/1.0M (auto)
────────────────────────────────────────────────────────────────────────────────────────────────────────────
❯
────────────────────────────────────────────────────────────────────────────────────────────────────────────
(anthropic) claude-opus-5 high • /private/tmp/atomic-e2e-r8
```

### The durable entries

```json
{"rung": "planned", "strategy": "verbatim-lines", "plannerModel_present": false, "linesBefore": 31, "linesKept": 15}
{"rung": "planned", "strategy": "verbatim-lines", "plannerModel_present": false, "linesBefore": 25, "linesKept": 13}
{"rung": "planned", "strategy": "verbatim-lines", "plannerModel_present": false, "linesBefore": 23, "linesKept": 12}
```

```text
entry kinds: {"session": 1, "model_change": 1, "thinking_level_change": 1, "message": 16, "compaction": 3}
```

Exactly one session-start `model_change` / `thinking_level_change` pair across three compactions — the session model was never mutated. The footer reads `(anthropic) claude-opus-5 high` throughout. No planner was borrowed, and no diagnostic or recovery sidecar was written, which is correct: nothing failed.

### The part that actually matters

A synthetic marker (`MERIDIAN-COBALT-3307`) was established **before** any compaction, then recalled correctly **after** three boundaries.

This is the only step that distinguishes a working compaction from a broken one. Steps 1–4 prove compaction *happened*; a compaction that silently destroyed context would pass all of them. It is precisely how the equivalent defect stayed invisible upstream — see the PR description.

### Honesty notes from the run

**A pre-existing defect was found and is not fixed here.** A mid-turn compaction followed by *parallel* tool calls produces an Anthropic 400 (`each tool_use must have a single result`). The run verified this against the unmodified baseline — with the entire change stashed via `git stash -u`, the same scenario reproduces the identical failure. Reported rather than silently omitted; out of scope for this PR.

**Nothing was skipped.** §10 records that no required success-path step failed to run. The rate-limit, quota, starvation, overflow-trimming and fresh-window rungs are deliberately not exercised live — those are the §8 unit and integration matrix.

**Secret scan clean.** §11 records the scan over all captures; no matches. The `MERIDIAN-COBALT-3307` / `APRICOT-LEDGER-7731` / `PELICAN-QUARRY-4402` strings are synthetic recall markers this run created, not credentials.

flora131:
Filed the two pre-existing defects found while capturing the E2E evidence, so they are tracked independently of this PR:

- #2051 — mid-turn compaction with parallel tool calls emits duplicate `tool_result` blocks → unrecoverable 400. Reproduces on base with the branch stashed. Recurring regression class upstream (earendil-works/pi#1764 → #4570 → #6832).
- #2052 — compaction stats mix provider-authoritative and heuristic token estimates, producing negative `percentReduction`. Display-only; `stats.tokensAfter` is never read for control flow.

Neither is caused by or fixed in this PR.

PR #2138, opened 2026-08-02
fix(workflows): keep overlay graph TUI responsive for large stage graphs

### LINKED ISSUE #2100: Atomic gets super slow after several stages
### What happened?

This is the same in Windows and WSL. As the number of stages increases, the responsiveness of the TUI decreases. 

### Steps to reproduce

Run a big workflow

### Expected behavior

_No response_

### Version

_No response_

Shreyasd10:
## Summary

When a workflow grows past a few dozen stages — especially wide fan-outs — the overlay graph TUI stopped feeling interactive. Every paint and every animation tick walked the full topology: cards, edges, hit targets, and layout rebuilds scaled with stage count rather than with what was on screen. Reported in [#2100](https://github.com/bastani-inc/atomic/issues/2100) as the same slowdown on Windows and WSL once a big workflow is running.

This change makes large graphs paint and tick against the visible viewport, keep topology/layout identity across status-only store updates, and skip animation frames when nothing is animating.

Closes #2100

### Viewport-bounded paint

`GraphCanvas` now takes optional `GraphCanvasBounds`. `setCell`, `hline`, and `vline` refuse work outside those bounds, and `toLines()` only materializes the clipped window. The graph renderer constructs the edge canvas with the current vertical/horizontal scroll window and skips edges whose AABB misses it entirely.

Card placement walks layout **bands** via `_firstVisibleLayoutBand(viewportTop)` and stops once a band is past `viewportBottom`. Off-screen columns are skipped; partially visible cards are column-sliced. Hit-target recording uses the same band walk and clips rects to the visible frame, so mouse activation cost tracks the viewport rather than the graph.

A 400-stage chain focused at the bottom paints the focused card and composes ≤16 body rows without mentioning the off-screen root. A 100-pair and a 1_000-pair fan-out both compose the same 13 rows and plot ≤4 edges — work stays flat as width grows.

### Topology retention across status-only updates

Status ticks used to rebuild the expanded graph and recompute layout even when only stage statuses changed. Two cooperating pieces now share one rule:

- `sameExpandedWorkflowTopology(left, right)` — structure equality over run identity, parent/root links, stage/tool ids, execution order, parent edges, and the failed/skipped terminal-boundary bit that can change nested expansion.
- `refreshExpandedWorkflowGraph(graph, snapshot)` — reuses cached virtual ids, parent ids, and targets while refreshing live stage/tool fields from the new snapshot. Returns `undefined` if anything structural is missing so the caller falls back to a full `expandWorkflowGraph`.

`GraphView._rebuildLayout` / `_graphStages` keep the same `cachedLayout` array identity, the same layout node objects, and the same `cachedRenderGeometry` when topology matches; only stage payloads and `hasAnimatingStages` update. Focused index and expanded targets stay put. Plain `failed` / `skipped` real-store updates take the same retention path.

### Constant-time animation gate

The overlay animation interval still fires every `ANIMATION_TICK_MS` (100 ms), but each tick now short-circuits through `_needsAnimationTick()`:

```ts
promptState !== null || toastManager.hasActive() || hasAnimatingStages
```

`hasAnimatingStages` is maintained at rebuild time (`running` / `awaiting_input` only). `ToastManager.hasActive()` is an O(1) length check so the gate never materializes the toast array. Idle pending-only graphs no longer burn ~10 full paints/sec; completing the last running stage stops timer-driven `requestRender` calls.

### Supporting micro-fixes in the same hot path

- `computeLayout` builds a `rowIndex` map once instead of `group.indexOf(s.id)` per stage (O(n) → O(1) per node on wide columns).
- `ANIMATION_TICK_MS` docs note the idle-skip contract so the next reader does not reintroduce unconditional ticks.

## Before / after

Tall chain (~400 stages), focused on the leaf; wide fan-out (~1_000 parallel pairs), focused near the origin.

**Before**

- Every frame composed every row, plotted every edge, and rebuilt layout on each status tick.
- Idle completed/pending graphs still requested ~10 paints/sec.
- Responsiveness degraded as stage count grew — same on Windows and WSL (#2100).

**After**

- Paint cost tracks the viewport: ≤16 composed rows for a 24-row overlay body; fan-out edge/card work stays flat from 100 → 1_000 pairs.
- Status-only updates retain layout array identity, render geometry, and expanded targets; the focused card updates in place.
- Idle graphs request zero animation frames; a running → completed transition stops ticks without disposing the overlay.

## Tests

Extended the existing animation suite and added a dedicated perf suite. Assertions measure bounded work (compose/edge/card counts, layout identity) rather than pasted glyphs.

`test/unit/overlay-graph-perf.test.ts` (new)

- composes and paints only the visible rows and cards in a tall graph
- clips off-screen fan-out edges and keeps expensive work viewport bounded
- bounds long edge segments to the viewport width and height
- retains layout and expanded target identity across status-only store updates
- retains expanded topology for plain failed and skipped real-store updates
- renders and records hit targets without iterating the complete layout
- checks idle animation eligibility without iterating graph nodes

`test/unit/overlay-graph-animation.test.ts`

- fires `requestRender` on a steady cadence while stages are animating
- skips animation ticks when no stage needs wall-clock paint
- stops timer-driven render requests after a real-store running stage completes
- does not start the timer in widget mode
- does not crash when `requestRender` is omitted in overlay mode
- stops firing renders after dispose
- running-stage border pulse advances with wall-clock time

## Verification

| Command | Outcome |
| --- | --- |
| `npm ci --ignore-scripts` | exit 0 (needed after fast-forwarding 45 commits behind `main`) |
| `npm run check` | exit 0 — biome clean (2347 files), `tsc --noEmit` clean, shrinkwrap current |
| `npx vitest --run --project unit test/unit/overlay-graph-perf.test.ts test/unit/overlay-graph-animation.test.ts` | exit 0 — 2 files, 14 passed |

## Notes

- `packages/workflows` still ships raw TypeScript — no `dist/`, no build step, no bundling.
- No dependency changes; no lockfile touched.
- Changelog entry added under `## [Unreleased]` → `### Fixed` in `packages/workflows/CHANGELOG.md`.

<h3>Greptile Summary</h3>

This change improves workflow-overlay responsiveness for large graphs by limiting drawing and hit testing to the visible viewport, retaining graph topology and layout across status-only updates, and avoiding animation-driven repaints for idle graphs.

Two potential regressions were disproved by executed checks: vertical and horizontal clipping preserved visible cards, edges, and click targets; and nested workflow boundaries correctly changed from flattened child content to the boundary card and back when their status moved from running to failed to completed. The focused runtime checks passed, along with 41 related graph navigation and performance tests.

**What T-Rex did**
- Completed the general contract validation, including viewport scrolling checks against the GraphView runtime, verification of nested workflow boundary updates, and running the focused validation plus graph navigation and performance regression suite with all tests passing.
- Updated the focused runtime test by correcting the click coordinate, then validated vertical and horizontal scrolling to confirm viewport clipping works as intended.
- Validated that the nested terminal-boundary transitions render inner → boundary → inner during the running → failed → completed sequence.
- Compared results against HEAD execution and concluded that baseline comparison was not useful due to the parent revision lacking the changed implementation, with HEAD providing the final validation evidence.

<a href="https://app.greptile.com/trex/runs/16953719/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(workflows): keep overlay graph TUI r..."](https://github.com/bastani-inc/atomic/commit/9505c1ba19a015d7097ab4638bf5d19a3e880df8) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49458502)</sub>


flora131:
Thanks for the contribution @Shreyasd10. Will take a look.

flora131:
Thanks for this @Shreyasd10 — the diagnosis in #2100 is right and the shape of the fix is the right one.

Two mechanical things meant I couldn't push to your branch conveniently, so I've re-pushed **your commit, unmodified**, to a branch in this repository and opened #2143.

**Your authorship is intact.** `git log` on that branch shows `shreyasdevadiga <shreyasdevadiga99@gmail.com>` as the commit author, GitHub attributes the contribution to you, and #2143 credits you in the first line of the description. The only difference is that the commit is rebased onto current `main` — this branch was 3 commits behind — so review fixes can land as separate commits on top without rewriting your work.

I'm reviewing the implementation now and will post findings. Follow-up commits will go on #2143; I'll close this one once that lands so we're not reviewing the same diff twice.

Appreciate the thorough PR description and the perf test suite — the bounded-work assertions are the right way to test this.

ashkanvg:
Hi @Shreyasd10 — while testing this PR with payload-heavy nested workflows, I found a separate bottleneck before the viewport renderer: every store mutation still built a full JSON clone of workflow inputs/results/child outputs/tool bodies.

I opened a companion PR against your branch: https://github.com/Shreyasd10/atomic/pull/1

On the exact #2138 head with 480 nodes + a 100 MiB workflow input, median mutation time was 147.71 ms; the companion reduces it to 2.85 ms while keeping your viewport render around 4 ms. It preserves your topology/layout/animation implementation and adds a cached immutable payload-free observation channel.

Validation: `npm run check`, 5,719 unit tests, 484 integration tests, 38 CI-contract tests, plus combined payload + viewport performance regressions. Happy for you to merge the companion PR or cherry-pick `cd07863c49c51640ad1e298667d486156a2f6d1d`.

flora131:
Update, @Shreyasd10 — @ashkanvg opened [Shreyasd10/atomic#1](https://github.com/Shreyasd10/atomic/pull/1) against your branch, stacking a companion fix on top of your commit. It's a good one: your change made painting topology-bounded, and he found that every store mutation still built a full `JSON.stringify`/`JSON.parse` snapshot, so a small status update walked complete workflow inputs and tool bodies before your optimized renderer ever ran.

The two belong together, so both are now on [#2143](https://github.com/bastani-inc/atomic/pull/2143) as two commits, each keeping its own author:

```
8a45311 fix(workflows): avoid cloning payloads for graph updates   ← @ashkanvg
608309f fix(workflows): keep overlay graph TUI responsive …        ← @Shreyasd10
```

You are both credited in the #2143 description, and GitHub attributes each commit to whoever wrote it.

The review of your commit has finished. Headline: five independent reviewers tried to falsify it and four failed. The clipped renderer produced **zero character differences** and byte-identical hit rectangles against `origin/main` across roughly 2,700 executed render comparisons plus an independent 150-graph fuzz. The topology gate survived 120,000 randomized snapshot pairs with zero divergences. The animation gate covers every clock-driven pixel any real store state can produce. That is a genuinely solid change.

One thing must be fixed, and it isn't in the rendering code: the changelog entry landed inside the already-tagged `## [0.9.11-alpha.9]` section while `## [Unreleased]` sits empty just above it. `AGENTS.md` forbids editing released sections. The entry text itself is accurate; it just needs to move up. I'm applying that now so you don't have to round-trip.

The rest is test sensitivity rather than defects, and I'll lay it out in the full review shortly. The short version: both tall-graph fixtures scroll to the bottom, so the band-walk `break` — the actual heart of the fix — can be turned into `continue` with all 216 overlay-graph tests still passing.

I'll post the detailed review here, then close this PR in favour of #2143.

flora131 (commented):
> Posted from a maintainer branch. The work under review is @Shreyasd10's commit `608309f`, now carried on [#2143](https://github.com/bastani-inc/atomic/pull/2143) with authorship intact.

# Review: `fix(workflows): keep overlay graph TUI responsive for large stage graphs`

## Verdict

The change is sound. Five reviewers tried to break the rendering, the topology gate, and the animation gate, and four of them came back with nothing. The clipped renderer produced **zero character differences** and byte-identical hit rectangles against `origin/main` across ~2,700 executed render comparisons plus an independent 150-graph fuzz; `sameExpandedWorkflowTopology` survived 120,000 randomized snapshot pairs and a 23-step end-to-end nested lifecycle with zero divergences; the animation gate covers every clock-driven pixel any real store state can produce. The measured win is real: a 2000-stage chain went from 2000 card renders and 256 ms/frame to 2 card renders and 0.5 ms/frame. Removing `// @ts-nocheck` from `overlay-graph-animation.test.ts` is a bonus worth keeping.

One repository-rule violation blocked merge, and it is not in the rendering code. **It has been fixed in this branch**, with a regression test. Everything else below is test sensitivity: I re-applied four mutations by hand and confirmed that the band-walk `break`, the `toLines` column origin, and the layout `parentIds` check can each be broken with all 216 overlay-graph tests still green. The code is right; the tests would not tell you if it stopped being right.

---

## Major — changelog entry landed in a released section (fixed in this branch)

`packages/workflows/CHANGELOG.md`, line 23; section header at line 9.

AGENTS.md is explicit: new entries always go under `## [Unreleased]`, and "each version section is immutable once released". `## [Unreleased]` at line 7 was empty, and the new entry landed inside `## [0.9.11-alpha.9] - 2026-08-01` — a section that is already tagged. The published notes for alpha.9 advertised a fix that shipped after it.

```
$ grep -n "^## \[" packages/workflows/CHANGELOG.md | head -3
7:## [Unreleased]
9:## [0.9.11-alpha.9] - 2026-08-01
$ grep -n 2100 packages/workflows/CHANGELOG.md
23:- Fixed workflow graph rendering for large, wide fan-outs by ...
$ git tag --list "*0.9.11-alpha.9*"
0.9.11-alpha.9
```

**What was done.** The entry moved verbatim under `[Unreleased]` beneath a new `### Fixed` heading, and the alpha.9 block was restored to its tagged content. `grep -n '^## \['` now prints `7:## [Unreleased]` and `13:## [0.9.11-alpha.9]`; the entry sits at line 11. A section-by-section comparison against `git show 0.9.11-alpha.9:packages/workflows/CHANGELOG.md` shows the alpha.9 block and the whole remainder of the file are byte-identical to the tag. `git diff origin/main -- packages/workflows/CHANGELOG.md` is now a single four-line addition under `[Unreleased]`.

The entry text itself is accurate and specific — all three of its claims were verified by probe — so it was kept as written.

**A regression test was added** in `test/unit/changelog.test.ts` (`released changelog sections`). It parametrizes over all seven `packages/*/CHANGELOG.md`, anchors on the newest version that resolves to a local tag, and deep-compares the whole released tail against the file at that tag. The three pre-existing `changelog parsing` tests are untouched.

It is not vacuous. Reverting only `packages/workflows/CHANGELOG.md` to its defective state while keeping the test:

```
$ npx vitest --run --project unit test/unit/changelog.test.ts
 1 failed | 10 passed
AssertionError: packages/workflows/CHANGELOG.md: released section [0.9.11-alpha.9] differs
from its content at tag 0.9.11-alpha.9. Released sections are immutable — new entries
belong under ## [Unreleased].
```

All seven per-package tests execute (none skip), a missing tag fails loudly via `assert.ok(anchor, ...)` rather than passing silently, and both checkout steps in `.github/workflows/test.yml` use `fetch-depth: 0`, so tags are present on Linux and Windows. `npm run check` and `npm run test:unit` both exit 0 (5801 passed, 2 pre-existing skips).

---

## Minor — the band-walk `break` survives mutation, because every tall fixture scrolls to the bottom

`packages/workflows/src/tui/graph-view-graph-render.ts:74` and `:134`.

`if (band.top >= viewportBottom) break;` is the line that makes card placement and hit-rect recording O(visible bands) instead of O(all bands). It is the point of the PR. Both tall-graph tests focus the last stage (`initialFocusedStageId: "s399"` / `"s999"`), so the viewport sits at the bottom of the scroll range with no bands below it — `break` and `continue` are indistinguishable there.

I applied the mutation myself:

```
$ perl -0pi -e 's/if \(band\.top >= viewportBottom\) break;/... continue;/g'   # both sites
$ npx vitest --run --project unit test/unit/overlay-graph
 Test Files  11 passed (11)
      Tests  216 passed (216)
```

Reverted; `git status --porcelain` empty. With a counting Proxy over `cachedRenderGeometry.bands`, band index reads for a 16-row viewport over 400 bands go **24 → 421** under that mutation. A future edit could restore the O(N)-per-frame walk this PR exists to remove and nothing would notice.

**Fix:** add a tall-graph case focused at the top (`initialFocusedStageId: "s0"`) and bound the walk alongside the existing `composeCalls`/`cards` counters — `assert.ok(view.bandVisits <= 40)`.

---

## Minor — `GraphCanvas.toLines()` content is never asserted

`packages/workflows/src/tui/graph-canvas.ts:159` (`let cursorCol = left;`), bounds at `:86-92`.

The one canvas test asserts `mergedCells` counts and `toLines().length`. It never looks at a line. Changing the emitted origin from `bounds.left` to `0` adds `left` leading spaces to every edge row whenever the graph is horizontally scrolled — routine for the wide fan-outs this PR targets — so edges detach from the cards.

```
$ perl -0pi -e 's/let cursorCol = left;/let cursorCol = 0;/'
$ npx vitest --run --project unit test/unit/overlay-graph
 Test Files  11 passed (11)   Tests  216 passed (216)
```

The exclusive-edge mutation at `setCell` (`row >= bottom` → `row > bottom`) also passes 14/14.

**Fix:** assert content. Build a canvas with `{top:0,bottom:1,left:100,right:110}`, set cells at columns 100 and 109, assert `toLines()[0] === "X        Y"`. Add a case writing at `row === bottom` / `col === right` and assert the lines stay empty.

---

## Minor — `_rebuildLayout`'s `sameTopology` parent comparison can be deleted undetected

`packages/workflows/src/tui/graph-view-state.ts:205-213`.

This check decides whether the cached `LayoutNode[]` **and** `cachedRenderGeometry` are reused. Replace the two `parentIds` lines with `return true` and the stage objects still refresh, but the geometry goes stale: after a reparent, cards keep old coordinates and edges keep pointing at the old parent. Suite stays green.

```
$ npx vitest --run --project unit test/unit/overlay-graph
 Test Files  11 passed (11)   Tests  216 passed (216)
```

The observable difference is precise: baseline `edgeGeometry = [[28,0,14,8]]`, mutated `[[0,0,14,8]]`. The retention tests miss it because they assert the mirrored stage fields, which the branch does reassign, not the derived geometry.

**Fix:** in the retention test, re-declare the run with a stage reparented and assert on `renderGeometryForTest().edges` coordinates instead of `node.stage.parentIds`.

---

## Minor — two of three animation-gate clauses, and `awaiting_input`, are unasserted

`packages/workflows/src/tui/graph-view-state.ts:173`, `:220-222`, `:232-234`.

`_needsAnimationTick()` is `promptState !== null || toastManager.hasActive() || hasAnimatingStages`. The tests pin only the third term. Each of these passes 14/14: drop the toast clause (M27), drop the prompt clause (M28), make the brand-new `toast.ts:62` `hasActive()` return `false` (M30), drop `|| stage.status === "awaiting_input"` at both recompute sites (M33). Probes P5/P6/P7/Q4/Q5 fail under them and pass at baseline, so these are real gaps, not equivalent mutants.

M30 means the method this diff adds has zero coverage. M33 contradicts the comment this diff adds at `graph-view-constants.ts:40-42`, which promises `awaiting_input` keeps ticking.

(These numbers come from the corrected mutation harness. The first pass used `--reporter=basic`, which does not exist in vitest 4, so every run aborted before collecting a test and the non-zero exit read as a kill. All 32 results were discarded and re-run under `--reporter=json` with an executed-test-count assertion. Good catch by that reviewer, and worth knowing if you write mutation harnesses here.)

**Fix:** one test that opens the gate through each condition independently — a toast via the manager, a pending prompt via the store, an `awaiting_input` stage — plus a direct `createToastManager()` test that `hasActive()` flips on `add` and back on `dismiss`.

---

## Minor — `toastManager.hasActive()` is now load-bearing, but nothing expires a toast

`packages/workflows/src/tui/toast.ts:49-63`, consumed at `graph-view-state.ts:173`.

`toasts` shrinks only in `tick(now)`, and `tick()` has no caller anywhere in the package — on this branch or on `origin/main`. The render path calls `active()`.

```
$ grep -rn "toastManager\.\(add\|tick\)\|\.tick(" packages/workflows --include=*.ts
(no output)
```

So a toast whose `dismissAfterMs` has elapsed stays on screen and pins the gate open at the full tick rate forever — exactly the failure the gate exists to prevent. Injecting a 1 s-TTL toast into an idle graph and advancing fake timers 60 s:

```
60s after a 1s-TTL toast: ticks=600 toastsLeft=1 gate=true
after explicit tick():    hasActive=false
```

The manager is fine; the caller is missing. It is unreachable today because `add()` has no caller either — the whole toast stack is dead code — but this diff makes a broken expiry contract load-bearing.

**Fix:** make `hasActive(now = Date.now())` time-aware and call `this.toastManager.tick(Date.now())` at the top of `_renderToastOverlay`. Or, if the toast stack is genuinely dead, delete it and drop the term from the gate rather than leaving a booby trap.

---

## Minor — the `topologyStatus` clause is unreachable dead code

`packages/workflows/src/shared/expanded-workflow-graph.ts:266-268` and `:313`.

```ts
(xChildRunId !== undefined && topologyStatus(x.status) !== topologyStatus(y.status)) ||
```

It sits after `xChildRunId !== yChildRunId` in the `||` chain, so it runs only when the ids are equal, and it requires them defined. `authoritativeWorkflowChildRunId` returns `undefined` for `failed` and `skipped`, so both stages must be neither, both `topologyStatus` calls return `"other"`, and the comparison is always false. The failed/skipped boundary flip it looks like it guards is caught one line earlier, where the id collapses to `undefined`.

Three reviewers converged, by three methods:

```
exhaustive stage-pair space: 5184 pairs
pairs that reach the clause with a defined child run id: 722
pairs on which it actually fires: 0

combinations=1600  where the status clause changed the answer=0
```

Collapsing `topologyStatus` to a constant and deleting the clause are both equivalent mutants no probe could kill.

**Fix:** delete `topologyStatus` and the clause, and note on the `xChildRunId !== yChildRunId` line that it is what catches the boundary flip. As written it reads like live safety logic and invites the next reviewer to assume the case is handled here.

---

## Minor — the two "bounded work" Proxy guards trap one iteration protocol, and one can never fire

`test/unit/overlay-graph-perf.test.ts:67-89`, asserted at `:278` and `:289`.

`guardLayoutIterationForTest` throws only on `Symbol.iterator`. The test named "renders and records hit targets without iterating the complete layout" therefore proves only that no `for…of` runs over `cachedLayout`. An indexed loop, `.map`, `.forEach` or `.some` over the full layout passes silently. Proved by subclassing `GraphView`, adding a deliberate `for (let i = 0; i < this.cachedLayout.length; i++)` scan to `_renderGraph`, and applying the identical guard copied from line 67:

```
full-layout elements touched during one frame: 1000
 ✓ 1 test passed
```

`guardAnimationLayoutScanForTest` is weaker still: `_needsAnimationTick()` reads three scalar fields and never touches `cachedLayout`, so it cannot fire for any implementation, and `assert.equal(view.animationEligible(), false)` is satisfied by an O(N) implementation too.

**Fix:** count reads instead of trapping one protocol. Wrap `cachedLayout`, `cachedDisplayStages` and `cachedRenderGeometry.edges` in Proxies that increment on numeric-key and `length` access, and assert an absolute per-frame budget derived from `bodyRows / (NODE_H + rowGap)`.

---

## Nit — the `cards` counter cannot see tool-node cards

`test/unit/overlay-graph-perf.test.ts:130-133, 154-157, 167, 178`.

Both card bounds count calls to the `getStageQueuedMessageCount` host hook. `_stageQueuedMessageCount` returns 0 without calling the hook whenever `_stageChatTarget` is undefined, which is every `nodeKind === "tool"` node. One stage plus 40 tool nodes at 24 rows and width 96:

```
tool cards painted=3   getStageQueuedMessageCount calls=1
```

A clipping regression confined to tool projections leaves both bounds satisfied. Count at the `renderNodeCard` call site through an instrumented seam, or add tool nodes to the fan-out fixture and assert on painted text.

---

## Nit — `assert.equal(result.composed, 13)` never exercises row clipping

`test/unit/overlay-graph-perf.test.ts:173-183`.

`parallelPairs` is two bands, so `totalRows = 1*(NODE_H + rowGap) + NODE_H = 13`, and the body is `24 - 2 - 3 - 3 = 16` rows. The graph fits. The composed-row count is the whole graph height, not evidence of clipping — row clipping is covered only by the tall-chain test. The 16/8/4/2 bounds are correct but written as literals rather than derived from `NODE_H`, `NODE_W`, `colGap` and the viewport, so a geometry-constant change turns them into confusing failures. Derive them, and give the fan-out fixture more than two depth levels so it crosses the body-row boundary.

---

## Nit — the failed/skipped retention test never renders, unlike its sibling

`test/unit/overlay-graph-perf.test.ts:232-265`.

The completed-status test ends with `assert.match(visibleText(view.render(80)), /✓ complete/)`. The failed/skipped test asserts object identity plus `layoutForTest()[0]?.stage.status === status` — it re-reads the field the refresh just wrote. A structurally correct but visually wrong card for a failed stage passes. Add the matching render assertion per status (`/✗ failed/`, `/skipped/`).

---

## Nit — `sameExpandedWorkflowTopology` ignores `run.name`

`packages/workflows/src/shared/expanded-workflow-graph.ts:289-327` (run field list), `:409` (`targets: graph.targets`).

The run-level comparison checks `id`, `parentRunId`, `parentStageId`, `rootRunId`, `stages.length`, `toolNodes.length` — not `name`. `refreshExpandedWorkflowGraph` then returns the cached `targets` map by reference, so a rename pins every nested stage target, and the label derived from it, to the old name:

```
sameExpandedWorkflowTopology(before, after) = true
full rebuild  target.runName = new-name
retention     target.runName = old-name
```

The function is internally inconsistent about this — `ExpandedWorkflowTool.runName` *is* re-read from `sources.run.name`, so after a rename the tool list and the stage targets disagree inside one graph. It was also the sole divergence in the 120,000-pair fuzz: `includeRunName=true pairs=20000 mismatches=3`, `runName` scrubbed → `mismatches=0`.

Unreachable today: no store method mutates `run.name` (`RunSnapshot.name` is `readonly`), and the only consumer, `expandedStageLabel`, is fed fresh graphs at `extension/workflow-targets.ts:145,182`. Either compare `a.name !== b.name`, or rebuild the targets map from `sources.run.name` the way `tools[].runName` already is.

---

## Nit — the emitted ANSI is not byte-identical, and that is an improvement

`packages/workflows/src/tui/graph-view-graph-render.ts:104-114`, `graph-canvas.ts:146-186`.

On-screen characters and hit geometry are identical. The escape sequences are not. The old path sliced the already-composed row, dropping SGR codes emitted left of the cut, so glyphs after the cut inherited stale style and the terminal default background leaked behind edge glyphs and trailing blanks. The new path slices each card line individually, so the style reaching a glyph is the one the card asked for.

Decoding SGR into `(char, fg, bg, bold)` cells over 500 random DAGs × random widths, rows and offsets:

```
characterMismatches=0
   9057  BG ch=SPACE  old=-  new=30,30,46
   2059  BG ch=─      old=-  new=30,30,46
    ...
      1  FG ch=│      old=88,91,112  new=166,227,161
```

Every divergence is one-directional — `30,30,46` is `graphTheme.bg`, and the reverse appears zero times. The ~140 real-glyph colour and bold diffs all sit at a left-clipped edge, and the dumps show the new value is the correct one (a completed card's border matching its own green `✓`; `borderDim` on edge runs). A second reviewer hit the same background divergence independently at `iter 1 row 4 col 63`.

**No code change is warranted.** Only the PR's "byte-identical" wording is wrong — say "identical characters and hit geometry". A cell-level golden test over a left-clipped card would pin the improved styling if you want a guard.

---

## Nit — `sameExpandedWorkflowTopology` has no direct unit test

`packages/workflows/src/shared/expanded-workflow-graph.ts:258-264`, `:289-327`.

The function is exported and pure, yet nothing calls it directly. Deleting the `nodeKind` comparison, deleting the `parentIds` comparison, or weakening `sameIds` to compare only definedness all leave the whole overlay-graph suite green.

One correction worth recording: the mutation reviewer rated this **blocking** on the strength of `if (left.runs.length === right.runs.length) return true;` passing the two diff test files. I applied that mutation and ran the wider suite:

```
$ npx vitest --run --project unit test/unit/overlay-graph
 2 failed | 214 passed
 × auto-focuses a newly awaiting stage prompt node so Enter attaches to the HIL UI
   (overlay-graph.test.ts, overlay-graph-navigation-02.test.ts)
```

The user-visible failure mode — a new stage never entering the graph — *is* covered, if only incidentally by two navigation tests. So the blocking rating does not hold and I downgraded it. The narrower gap is real: add a direct test asserting `true` for an unchanged snapshot and `false` for a reparent, a `nodeKind` flip, an added stage, and a same-length parent-id swap.

---

## What was probed and came back clean

So you can see the coverage behind the verdict:

- **Render parity at every scroll extreme.** 1,416 configurations — chain-of-40, 12×6 cross-linked fan, 5-node diamond, single stage; widths 8–96 including the 40-column floor; viewport rows 7 (one body row) through 24; vertical offsets 0–200 and horizontal 0–400, covering top, mid, bottom and past-the-end clamping. `charDiff=0 countDiff=0`.
- **Randomized differential.** 500 random DAGs with a `recordStageEnd` → `invalidate()` → re-render on each. `mismatches=0`. A second reviewer's independent 150-graph fuzz agreed on glyphs, per-column background state and hit rects.
- **Growing topology.** 216 append-a-rank steps forcing `_buildRenderGeometry` to rebuild while the graph crossed the viewport boundary. `mismatches=0`.
- **Keyboard-driven scrolling.** 12 navigation keys through `handleInput` so `_scrollFocusedIntoView` ran naturally rather than being poked. `mismatches=0` — moving vertical scroll-into-view from `_visibleGraphLines` into `_renderGraph` does not shift a frame.
- **Click targeting.** 360 viewport configurations with JSON-identical hit rects, then 252 real SGR click sequences checked against the text actually painted under the cursor. Never a neighbour.
- **Canvas boundaries.** `hline`/`vline` honour half-open `[left,right)` / `[top,bottom)` exactly; reversed endpoints, zero-width and inverted bounds paint nothing; every mutation path funnels through `setCell`, and firing every other entry point out of bounds wrote nothing. 2,000 random op sequences against an unbounded canvas: identical to `origin/main`, string for string.
- **Topology retention.** 26-case structural enumeration (25 detected, the 26th is the `run.name` nit); 120,000 randomized pairs with `mismatches=0` across the 31,787 declared identical; a field-by-field comparison over every mutable stage and tool field with zero diffs; a 23-step nested lifecycle driven through the real `Store` where `graphRetained=15` of 23 and divergences were 0.
- **Animation gate.** Every clock read in the render path enumerated by stack trace — exactly three (`elapsedStageMs`, `pulsePhase`, prompt caret). All eight statuses checked for clock sensitivity against the gate verdict; every real-store state is safe. `blocked` was hunted specifically and is fine — the card swaps the duration line for the blocked badge. Restart latency after pause/resume/awaiting-input reopens the gate in one interval, and the double-notify from `overlay-adapter.ts` is absorbed.
- **Conventions.** No `dist`/`tsconfig.build`/`outDir`; every added import ends in `.js`; no `any` or `unknown` added; no explicit per-test timeouts added or restated; slowest new test 404 ms against the 30 s budget; nothing in `packages/coding-agent/docs` documents repaint cadence or clipping internals, so no doc update is owed.
- **Adjacent suites.** 9 files, 202 tests pass, including the 29 layout tests covering the `group.indexOf(s.id)` → precomputed `rowIndex` change (behaviour-preserving: every id is in `colMap` before `colGroups` is built, so `indexOf` never returned `-1`).

Four claims were raised and dropped as not defects: the residual O(N + E) per-frame cost (real, but the change never claimed O(1) frames and it removed an O(N²) walk); five synthetic frozen stage shapes no shipped writer can produce; the empty-graph early return no longer resetting `graphScrollOffset` (the next render clamps it); and four equivalent mutants correctly left uncaught.

---

## Recommended action

The blocking changelog defect is fixed in this branch and covered by a new test that fails on the original mistake. What remains is test hardening, none of it in the shipped code.

Before merge, add the three assertions that cover the mutations I confirmed survive — the top-anchored band-walk bound, the `toLines` content assertion, and the reparent geometry assertion. Each is a few lines and each guards a line this PR exists to add.

After that, in this PR or a follow-up: assert the two unpinned animation-gate clauses and `awaiting_input`; add a direct unit test for `sameExpandedWorkflowTopology`; and decide on the toast stack — make `hasActive()` time-aware and call `tick()` in the render path, or delete the stack and drop the term from the gate. Delete the dead `topologyStatus` clause. Reword "byte-identical" to "identical characters and hit geometry" in the PR description.

Nice work. The hard part is right; the tests just need to catch up to it.

flora131:
Closing in favour of [#2143](https://github.com/bastani-inc/atomic/pull/2143), which carries your commit `608309f` with you as author, plus a maintainer commit fixing the changelog placement flagged in the review above.

@ashkanvg's companion work is stacked on top as [#2144](https://github.com/bastani-inc/atomic/pull/2144), so the two land as separate reviewable PRs with separate credit rather than one combined diff.

Nothing about this reflects on the work — the review speaks for itself. This is only about where the commits live so maintainers can push fixes. Thanks again @Shreyasd10.

flora131:
Shipped. @Shreyasd10 — your commit is on `main` as [`532c4d0`](https://github.com/bastani-inc/atomic/commit/532c4d0d0), with you as the author:

```
532c4d0d0 shreyasdevadiga <shreyasdevadiga99@gmail.com>
  fix(workflows): keep overlay graph TUI responsive for large stage graphs
```

It merged with a merge commit rather than a squash specifically so that stays true — `git log --author` finds it and `git blame` attributes to you.

For the record of what your change was put through: five independent adversarial reviewers tried to falsify it and four came back with nothing. Zero character differences and byte-identical hit rectangles against `main` across ~2,700 executed render comparisons, 120,000 randomized topology snapshot pairs with no divergence, and a 2000-stage chain going from 256 ms/frame to 0.5 ms/frame. The implementation was right. What needed work was the tests around it, and that is a much better problem to have.

@ashkanvg's companion fix landed alongside it as [`0f11d37`](https://github.com/bastani-inc/atomic/commit/0f11d3799), also authored to them. Four follow-up PRs went on top: a memory leak on the durable path that his work surfaced, test hardening, a CI gap, and removal of a dead toast stack.

Thanks for the fix and for the unusually good PR description — the before/after framing made the whole thing easy to review.

Shreyasd10:
Thanks @flora131 — appreciate the review and for getting this landed with authorship intact.

PR #2143, opened 2026-08-02
fix(workflows): keep overlay graph TUI responsive for large stage graphs

### LINKED ISSUE #2100: Atomic gets super slow after several stages

flora131:
Carries #2138 by [@Shreyasd10](https://github.com/Shreyasd10). Closes #2100.

**This is [@Shreyasd10](https://github.com/Shreyasd10)'s work.** `608309f` is their commit, unmodified apart from a rebase onto current `main`; `git log` shows `shreyasdevadiga <shreyasdevadiga99@gmail.com>` as author and GitHub attributes it to them. It lives on a branch in this repository so maintainers can push review fixes, which #2138 on their fork did not allow conveniently.

**Bottom of stack #2145.** [#2144](https://github.com/bastani-inc/atomic/pull/2144) by [@ashkanvg](https://github.com/ashkanvg) stacks on top and fixes the complementary half of the same slowdown.

| Commit | Author |
| --- | --- |
| `608309f` — keep overlay graph TUI responsive for large stage graphs | [@Shreyasd10](https://github.com/Shreyasd10) |
| `72d6fa4` — move the #2100 changelog entry out of a released section | maintainer (review fix) |

## Summary

Large workflow graphs, especially wide fan-outs, made the overlay graph TUI stop feeling interactive. Every paint and every animation tick walked the full topology: cards, edges, hit targets, and layout rebuilds scaled with stage count rather than with what was on screen. Reported in #2100 as the same slowdown on Windows and WSL.

This makes large graphs paint and tick against the visible viewport, keep topology and layout identity across status-only store updates, and skip animation frames when nothing is animating.

### Viewport-bounded paint

`GraphCanvas` takes optional `GraphCanvasBounds`. `setCell`, `hline`, and `vline` refuse work outside those bounds, and `toLines()` materializes only the clipped window. The renderer builds the edge canvas with the current scroll window and skips edges whose AABB misses it.

Card placement walks layout bands via `_firstVisibleLayoutBand(viewportTop)` and stops once a band is past `viewportBottom`. Off-screen columns are skipped, partially visible cards are column-sliced, and hit-target recording uses the same band walk, so mouse cost tracks the viewport rather than the graph.

### Topology retention across status-only updates

- `sameExpandedWorkflowTopology(left, right)` — structure equality over run identity, parent/root links, stage and tool ids, execution order, parent edges, and the failed/skipped terminal-boundary bit that can change nested expansion.
- `refreshExpandedWorkflowGraph(graph, snapshot)` — reuses cached virtual ids, parent ids, and targets while refreshing live stage and tool fields. Returns `undefined` when anything structural is missing, so the caller falls back to a full `expandWorkflowGraph`.

`_rebuildLayout` keeps the same `cachedLayout` array identity, the same layout node objects, and the same `cachedRenderGeometry` when topology matches.

### Constant-time animation gate

Each 100 ms tick short-circuits through `_needsAnimationTick()`: `promptState !== null || toastManager.hasActive() || hasAnimatingStages`. Idle pending-only graphs no longer burn ~10 full paints/sec.

## Review

A five-verifier adversarial review ran against this branch; the [full report is on #2138](https://github.com/bastani-inc/atomic/pull/2138#pullrequestreview-3849185339). Headline:

- **Zero character differences** and byte-identical hit rectangles against `origin/main` across ~2,700 executed render comparisons, plus an independent 150-graph fuzz.
- `sameExpandedWorkflowTopology` survived **120,000 randomized snapshot pairs** and a 23-step nested lifecycle with zero divergences.
- Measured win: a 2000-stage chain went from 2000 card renders and 256 ms/frame to 2 card renders and 0.5 ms/frame.

One correction to the original description: output is **not** byte-identical. Characters and hit geometry are; the ANSI differs, and every divergence is the new renderer being more correct — the old path sliced composed rows and dropped SGR codes left of the cut, so glyphs inherited stale style.

One defect was found and is fixed here in `72d6fa4`: the changelog entry had landed inside the already-tagged `## [0.9.11-alpha.9]` section while `## [Unreleased]` sat empty, which `AGENTS.md` forbids. A `released changelog sections` test now compares every package changelog's released tail against its content at that tag, so it cannot recur silently.

Remaining review items are test sensitivity, not defects in shipped code. The most worthwhile: the band-walk `break` — the heart of the fix — can be changed to `continue` with all 216 overlay-graph tests passing, because both tall-graph fixtures scroll to the bottom.

## Verification

| Command | Outcome |
| --- | --- |
| `npm run check` | exit 0 |
| `npm run test:unit` | exit 0 |

## Notes

- `packages/workflows` still ships raw TypeScript — no `dist/`, no build step, no bundling.
- No dependency changes; no lockfile touched.

<h3>Greptile Summary</h3>

This change improves workflow-overlay responsiveness for large graphs by clipping rendering and hit testing to the visible viewport, preserving graph layout during status-only updates, and avoiding idle animation redraws.

Focused runtime checks disproved three possible failures: status-only updates retained the existing layout while updating the displayed status; a nested workflow boundary change rebuilt the displayed topology as child nodes became available; and a 40-node graph remained constrained to a 24×44 viewport with mouse targets limited to visible cards.

The PR is safe to merge; no blocking failure remains.

The exercised overlay paths correctly retained layout for status changes, replaced topology for nested workflow expansion, and constrained wide and tall graph rendering and hit targets to the visible viewport.

**What T-Rex did**
- T-Rex ran npx vitest --run --project unit test/unit/pr-2143-workflow-overlay-runtime.test.ts against the real GraphView and store paths.
- The unit tests passed after exercising status refresh without layout replacement, nested-boundary topology replacement, viewport clipping, and a visible SGR mouse click.
- T-Rex then ran the before/after runtime probe and observed retained status layout identity, child-node topology replacement, and a graph viewport constrained to 24 rows by 44 columns.
- The second validation step documented that the status-only relayout failure, the nested workflow boundary topology replacement failure, and the wide/tall viewport clipping/visible hit-target failure were disproved.

<a href="https://app.greptile.com/trex/runs/17016283/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

<sub>Reviews (5): Last reviewed commit: ["fix(test): make the released-changelog c..."](https://github.com/bastani-inc/atomic/commit/bafe240c6d5f9c89ac82b45e4264a37e231ca47f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49514510)</sub>


flora131:
## Test hardening applied — ready for review

The review's remaining items are closed. Each new assertion was scored by applying the exact regression it targets and requiring a test to fail; the pass ran under an automated kill-matrix rather than self-report.

**Now pinned, with the regression each one kills:**

| Contract | Regression that now fails |
| --- | --- |
| Band-walk early exit | `break` → `continue` (top-anchored 400-stage fixture; band visits 24 → 421) |
| Per-frame read budgets | per-card layout slice, duplicate edge walk, per-card display-stage clone |
| `toLines()` column origin | `cursorCol = left` → `0` |
| Canvas half-open bounds | `>=` → `>` at either edge |
| Reparent geometry | deleting the `parentIds` comparison in `sameTopology` |
| Animation gate, per clause | dropping the prompt clause, the toast clause, or `awaiting_input` at either recompute site |
| Toast expiry ordering | moving `tick()` after the `active()` read |
| `sameExpandedWorkflowTopology` | reparent, `nodeKind` flip, added stage, same-length parent-id swap |
| Painted edges and header counts | removing all edge plotting; emptying the header's stages |

That last row is worth calling out. The read budgets bound how much work a frame does and say **nothing** about what it paints — removing every `_plotEdge` call, or passing `stages: []` to the header, left all 231 overlay-graph tests green. Those two now have output assertions.

Both former `cards <= 2` upper bounds became equalities: a direct probe confirmed both 16-row tall fixtures paint exactly 2 cards. The unexplained `+ 8` slack in the layout budget is gone — it was wide enough to hide a per-card layout slice.

**Deliberately O(N) budgets.** The edge budget is `2E+1` and the display-stage budget `2N+2`, because the render examines every edge for a cheap viewport-bounds test before plotting only the visible ones, and the header scans stages once. They catch a duplicate walk or a per-card clone; they are commented so nobody later reads them as viewport bounds.

**One source deletion:** the unreachable `topologyStatus` clause in `sameExpandedWorkflowTopology`. It sits behind an equal-child-run-id check and needs both ids defined, but `authoritativeWorkflowChildRunId` returns `undefined` for `failed` and `skipped`, so it never fired across an exhaustive 5184-pair cross-product. The boundary flip it appeared to guard is caught one line earlier.

**Left as a follow-up, deliberately.** The toast stack has no production producer anywhere in the repository — the expiry fix is real and stayed, and the gate's toast clause is covered by tests, but wiring or deleting the stack is not this PR's job. An earlier automated pass tried to justify keeping it by adding a public `GraphOverlayPort.showToast` API with no consumer; I reverted that. Growing the public API of a perf PR to make dead code look live is the wrong trade.

`npm run check` and `npm run test:unit` green (5833 passed). Production diff is unchanged from the reviewed implementation apart from that one dead-clause deletion.

flora131:
## CI fixed — two failures, both mine, both outside the gates I was running

**Windows (all four PRs): the released-changelog check compared line endings, not content.**

The check I added reads the working-tree file and compares it against `git show <tag>:<path>`. On Windows `core.autocrlf` gives the checkout CRLF, while `git show` writes the blob verbatim with LF — so splitting both on `"\n"` left a trailing `\r` on every working-tree line and reported **all seven** package changelogs as modified. Now splits on `/\r?\n/`.

It could not fail locally: macOS and Linux checkouts are LF. Worth noting the earlier verification of that test checked that CI fetches tags (`fetch-depth: 0`) and stopped there — tag availability was the risk I thought about, line endings were not.

**Linux (this PR only): this layer broke an integration fixture and did not fix it.**

The animation gate means a run with `stages: []` leaves `hasAnimatingStages` false, so no tick fires and `overlay-entrypoints-animation.test.ts` fails — along with `overlay-entrypoints.test.ts`, which imports it. The repair existed, but one layer **above** this one, which left this layer red on its own. A lower layer in a stack has to be independently green, so the repair moved down here, to the change that broke it. It is [@ashkanvg](https://github.com/ashkanvg)'s work, reproduced unchanged and co-authored on the commit.

**The process gap.** Both failures live outside `npm run check` + `npm run test:unit`, which is all the pre-push hook runs and all I was running. `test:integration` and `test:ci-contracts` were never gated at any point in this work.

All four suites now pass locally on every layer:

| Suite | Result |
| --- | --- |
| `npm run check` | exit 0 |
| `npm run test:unit` | 622 files, 5836 passed |
| `npm run test:integration` | 35 files, 484 passed |
| `npm run test:ci-contracts` | 5 files, 38 passed |

Separately worth considering, though not part of this stack: the pre-push hook runs only `check` and `test:unit`, so this class of failure reaches CI by default.

PR #2187, opened 2026-08-04
feat(install): add standalone release archive installers

### LINKED ISSUE #2159: feat(install): self-contained curl/irm installers with no npm/bun dependency
## What do you want to change?

Add one-line installers at the repository root so Atomic can be installed without a JavaScript package manager:

```bash
curl -fsSL https://raw.githubusercontent.com/bastani-inc/atomic/main/install.sh | sh
```

```powershell
irm https://raw.githubusercontent.com/bastani-inc/atomic/main/install.ps1 | iex
```

Model the core logic on oh-my-pi's installers — [`scripts/install.sh`](https://github.com/can1357/oh-my-pi/blob/main/scripts/install.sh) and [`scripts/install.ps1`](https://github.com/can1357/oh-my-pi/blob/main/scripts/install.ps1) — with one deliberate divergence: **the Atomic installer must never call npm, pnpm, yarn, or bun, and must never bootstrap bun.** It installs the prebuilt GitHub Release archive and nothing else.

## Why?

- Today `README.md` offers only `npm install -g @bastani/atomic`, `pnpm add -g`, and `bun add -g`. Every path goes through `registry.npmjs.org`. Users behind a corporate proxy, an artifact-mirror allowlist, or an outright npm-registry block cannot install Atomic at all, even though every release already ships self-contained binaries on GitHub Releases.
- No Node/npm requirement also removes the `Node.js 22.19 or newer` prerequisite for binary users, which matters for CI images, devcontainers, and locked-down workstations.
- The repo used to have a root `install.sh` / `install.ps1`; they were dropped in the monorepo restructure (#936) and #947 has been 404ing since. This restores that entry point on the current release layout.

### Files

- `install.sh` — repo root, POSIX `sh` (`#!/bin/sh`, no bashisms; it must run under Alpine's busybox `ash`).
- `install.ps1` — repo root, Windows PowerShell 5.1 compatible.
- Raw URLs: `https://raw.githubusercontent.com/bastani-inc/atomic/main/install.{sh,ps1}`.

### Carry over from oh-my-pi

- Flag parsing (`--ref`/`--ref=`/`-r`, `-Ref`), `--help`.
- `host_arch()`, including the `sysctl -in hw.optional.arm64` check so a Rosetta shell on Apple Silicon still gets the `arm64` build rather than the translated `x86_64` one.
- musl detection: `/etc/alpine-release` or `ldd --version 2>&1 | grep -qi musl`.
- Release-tag resolution, install-dir handling, and the "not on PATH" hint at the end.
- `install.ps1` appending the install dir to the **User** PATH and telling the user to restart the terminal.

### Drop from oh-my-pi

- `--source` / `-Source` mode, `install_bun`, `require_bun_version`, `bun install -g`, `bun_arch_matches_host`, and the whole default-to-bun branch. There is one mode: download the release archive.
- The only external tools allowed are `curl` **or** `wget`, `tar`, `mkdir`, `mv`, `chmod`, and one of `sha256sum` / `shasum` / `openssl`. On Windows: `Invoke-WebRequest`, `Expand-Archive`, `Get-FileHash`. No `jq`, no `git`.

### Atomic-specific: the release assets are directory archives, not bare binaries

oh-my-pi downloads a single `omp-<platform>-<arch>` file. Atomic does not. `scripts/build-binaries.sh` and `.github/workflows/publish.yml` publish these nine assets:

| Host | Asset | Archive layout |
| --- | --- | --- |
| macOS arm64 / x64 | `atomic-darwin-arm64.tar.gz`, `atomic-darwin-x64.tar.gz` | top-level `atomic/` directory |
| Linux glibc | `atomic-linux-x64.tar.gz`, `atomic-linux-arm64.tar.gz` | top-level `atomic/` directory |
| Linux musl | `atomic-linux-x64-musl.tar.gz`, `atomic-linux-arm64-musl.tar.gz` | top-level `atomic/` directory |
| Windows | `atomic-windows-x64.zip`, `atomic-windows-arm64.zip` | **flat** — `atomic.exe` at the archive root |
| all | `SHA256SUMS` | `sha256sum` format over the eight archives |

So the installer must:

1. Map platform + arch + libc to the asset name above; fail with a clear message on anything unmapped.
2. Download to a temp dir and `trap`/`try-finally` clean it up on any failure.
3. Fetch `SHA256SUMS` from the same release and verify the archive before extracting. Refuse to install on mismatch.
4. Extract the full payload — the binary needs its sibling `app.js`, `builtin/`, and `node_modules/` — into a versioned root, e.g. `${ATOMIC_INSTALL_DIR:-$HOME/.local/share/atomic}/versions/<version>`, then swap the current pointer and place a symlink (or a tiny shim on Windows) at `${ATOMIC_BIN_DIR:-$HOME/.local/bin}/atomic`. Note the tarball nests under `atomic/` while the zip does not.
5. Be idempotent: reinstalling the same version, or upgrading over an existing install, must both succeed without leaving a half-extracted tree.

### Version resolution without the API where possible

Prefer following the `https://github.com/bastani-inc/atomic/releases/latest` redirect and reading the tag from the resolved `Location`, falling back to `api.github.com/repos/bastani-inc/atomic/releases/{latest,tags/<ref>}`. Unauthenticated `api.github.com` is 60 requests/hour per IP, which is a real failure mode on shared CI egress. Honor `GITHUB_TOKEN`/`GH_TOKEN` as an `Authorization` header when present, but never require it. `releases/latest` also excludes prereleases, which is the behavior we want given `*-alpha.*` tags are published continuously; `--ref 0.9.11-alpha.12` stays the opt-in.

### Knobs

`ATOMIC_INSTALL_DIR`, `ATOMIC_BIN_DIR`, `ATOMIC_VERSION` (equivalent to `--ref`), `GITHUB_TOKEN`/`GH_TOKEN`. Windows default install dir `$env:LOCALAPPDATA\atomic`.

### Drift protection

Add a contract test alongside the existing ones in `test/ci/` asserting that the asset names in `install.sh` and `install.ps1` are exactly the set built by `scripts/build-binaries.sh` and uploaded by the `stage-github-release` job in `.github/workflows/publish.yml`. Renaming an archive without updating the installers should fail CI, not fail in users' terminals.

### Docs

- `README.md` — add the two one-liners as the first install option, ahead of the package managers, and note that the binary path has no Node.js requirement.
- `packages/coding-agent/docs/quickstart.md` and `docs/windows.md` — same, plus the PowerShell arg form (`& ([scriptblock]::Create((irm ...))) -Ref 0.9.11`).
- `docs/containerization.md` / `docs/termux.md` if the binary path changes the guidance there.

### Known follow-up (out of scope unless a contributor wants it)

`detectInstallMethodForRuntime()` in `packages/coding-agent/src/config-self-update.ts` already returns `bun-binary` for standalone builds, and `getSelfUpdateCommandForMethod()` returns `undefined` for it — so `atomic update` has no path for archive installs. Once the installers exist, that case can print or re-run the `curl`/`irm` one-liner. Happy to take that as a separate PR.

## Acceptance criteria

- [ ] `install.sh` and `install.ps1` exist at the repo root and install a working `atomic` with **zero** npm/pnpm/yarn/bun/Node.js involvement.
- [ ] `sh install.sh` succeeds on Alpine (busybox, musl) and on Debian slim with only `curl`/`wget` + `tar` present.
- [ ] Correct asset selected for darwin-arm64 (including under Rosetta), darwin-x64, linux-x64, linux-arm64, linux-x64-musl, linux-arm64-musl, windows-x64, windows-arm64.
- [ ] SHA256 verified against the release `SHA256SUMS`; a corrupted download aborts without touching an existing install.
- [ ] `--ref <tag>` / `-Ref <tag>` pins an exact release; the default resolves the latest stable and skips prereleases.
- [ ] `atomic --version` works after install; PATH guidance is printed when the bin dir is not on PATH.
- [ ] Rerunning the installer upgrades in place and is idempotent.
- [ ] Asset-name contract test added under `test/ci/`.
- [ ] README and `packages/coding-agent/docs` updated.

## Notes for contributors

Open to contributors — comment with your approach before starting and a maintainer will assign it (see [CONTRIBUTING.md](https://github.com/bastani-inc/atomic/blob/main/CONTRIBUTING.md)). The shell and PowerShell scripts can land as one PR; the `atomic update` follow-up above should be a separate one. No `packages/*/CHANGELOG.md` entry is needed for the scripts themselves — they are distribution tooling, not shipped package behavior — but the README/docs changes belong in the same PR.

flora131:
## Summary

Adds standalone POSIX shell and Windows PowerShell installers that install Atomic from verified GitHub Release archives without npm, pnpm, Yarn, Bun, or Node.js.

## Changes

- add root `install.sh` and `install.ps1` entry points with stable/latest and pinned-release support
- select the correct Darwin, Linux glibc/musl, and Windows release asset
- verify `SHA256SUMS` before changing an existing install
- install complete payloads into versioned roots with current pointers and bin launchers
- preserve prior installs and clean temporary state on failures and repeated installs
- harden Windows PATHEXT, pointer, rollback, temporary-directory cleanup, and transaction-owned bin-path rejection
- keep cleanup warnings from replacing the primary installer error under caller-controlled PowerShell preferences
- add asset-drift, installer safety, platform fixture, container, and documentation contracts
- document the curl/irm path across README, quickstart, Windows, container, and Termux guidance
- merge current main; keep both the musl archive matrix and main's model-data hydration / publish-release docs

## Notes

- No package changelog entry was added, per issue scope.
- Archive-based `atomic update` remains a separate follow-up.

Closes #2159

## Validation

- Candidate SHA: `4a90d32e665fb6d193a7fa728fb63eb71ee039b6`
- Parent workflow gates passed: `npm run check`, `npm run test:ci-contracts`, focused installer unit tests

<h3>Greptile Summary</h3>

Adds standalone, checksum-verified release-archive installers for POSIX and Windows, alongside musl archive runtime bundling, CI smoke coverage, and installer documentation.

Executed checks disproved the POSIX installer failure paths examined: Alpine without `ldd` selected the musl asset, Debian selected the glibc asset, a forged checksum stopped before any installation or launcher was created, and valid local release fixtures installed and launched successfully. Musl runtime staging also completed for x64 and arm64 with verified `libgcc`/`libstdc++` libraries and a payload-local `$ORIGIN/lib` search path.

### T-Rex validation blocked
Native Windows installer execution was blocked by a missing tool: neither `pwsh` nor `powershell` is installed. The repository’s PowerShell installer suite was also blocked by a missing package, `@bastani/pi-ai/compat`, from the unbuilt AI workspace.

The PR is safe to merge; no blocking failure remains.

The exercised POSIX installation, checksum-rejection, platform-selection, launcher, and musl runtime paths behaved as intended, and no final defect was found.

**What T-Rex did**
- Validated POSIX installer behavior by running the POSIX installer validation script, capturing the forged-checksum rejection path and the subsequent valid installation and launcher flow.
- Validated musl runtime staging for linux-x64-musl and linux-arm64-musl, confirming that required libraries are identified before runtime and that RUNPATH is set to $ORIGIN/lib after staging.
- Prepared and uploaded artifacts for both the POSIX installer and musl runtime validations, including the validation scripts and the before/after capture logs.
- T-Rex ran the requested verification but reported that local artifact references were not uploaded.

<a href="https://app.greptile.com/trex/runs/20322261/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

<sub>Reviews (7): Last reviewed commit: ["test(installer): run Windows junction ov..."](https://github.com/bastani-inc/atomic/commit/fffb488986f0f52f13914cf090cabdc68546047e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50277073)</sub>


flora131:
## Implementation notes (verbatim artifact)

Source: `artifact-f1c6083f-a24a-48dd-b7f0-2bf7a22862a5/implementation-notes.md` from workflow run `b435364a-aeed-4346-87a4-f4cd8d33e485`.

---

# Implementation Notes

Task: Implement https://github.com/bastani-inc/atomic/issues/2159 in the Atomic repository. Start from the latest fetched origin/main in the requested isolated worktree. Add root POSIX-sh install.sh and Windows PowerShell 5.1 install.ps1 that install only prebuilt GitHub Release archives, never invoke or bootstrap npm, pnpm, yarn, bun, or Node.js, resolve stable/latest and pinned refs, select all supported platform/arch/libc assets, verify SHA256SUMS before touching an existing install, atomically install the full archive payload into versioned roots with a current pointer plus bin symlink/shim, clean up failures, support the issue's environment knobs, remain idempotent, and give PATH guidance. Add robust automated tests, including the exact asset-name drift contract against scripts/build-binaries.sh and the stage-github-release upload list in .github/workflows/publish.yml. Test shell behavior for BusyBox/Alpine-compatible POSIX sh and relevant download/checksum/error/idempotency flows without relying on forbidden package tools; test PowerShell where feasible and use contract/static checks where the host lacks Windows PowerShell. Update README.md, packages/coding-agent/docs/quickstart.md, docs/windows.md, and any containerization/Termux guidance affected by the binary path. Do not add a package changelog entry. Follow CONTRIBUTING.md and repository rules, run targeted suites and npm run check, inspect the full diff, and commit/push a focused feature branch. The implementation must satisfy every acceptance criterion and constraint in issue #2159.

## Running Notes

- Record implementation decisions, deviations from research, tradeoffs, blockers, validation outcomes, and user-relevant facts. Keep entries concise and readable.
- Before reporting progress, audit each claim against a tool result from this session. Report only work you can point to evidence for; say so explicitly when something is unverified.

## Iteration 8: findings artifact was empty, findings recovered elsewhere

The consolidated unresolved-findings artifact for this iteration,
`research/2026-08-04-implement-https-github-com-bastani-inc-atomic-issues-2159-in-the-atomic-reposito.md`,
is **0 bytes** (`ls -la research/`, `wc -c`). That is a research-stage write failure, not an absence of
findings: `TODO-a86668c0` ("Research iteration 8 installer findings", status completed) records three
blockers, and `artifact-59a742c4/review-reviewer-b.json` contains the same three with reproductions and
code locations. I recovered the batch from those two sources and repaired it rather than treating the
empty file as an approval.

Cycle-8 review split: reviewer-a returned `patch is correct` with `stop_review_loop: true`; reviewer-b
returned `patch is incorrect` with the three findings below, all marked `required_by_objective`.

### Frozen batch and repairs

**1. [P2] Exact-ref identity was not verified (both installers).** The `--ref`/`-Ref` flow used the
requested tag only to pick the API URL, then trusted whatever `tag_name` came back. Reviewer-b's fixture
asked for `1.0.0`, received `{"tag_name":"2.0.0"}`, and got 2.0.0 installed with `current` pointing at
`versions/2.0.0`. That breaks the issue's "`--ref <tag>` pins an exact release" criterion.

Repair: both installers now compare the resolved tag with the requested ref and fail closed before the
download base is built — `install.sh` with `[ "$RELEASE_TAG" != "$REQUESTED_REF" ]`, `install.ps1` with
`$releaseTag -cne $requestedRef`. Latest-release resolution is untouched, since it has no requested identity.

**2. [P2] A semicolon-containing Windows `ATOMIC_BIN_DIR` corrupted PATH.** The raw value was appended to
the semicolon-delimited PATH, so `Test-AtomicPathContains` read it as two entries: `atomic` was not found
in a new terminal and every rerun appended the broken value again.

Repair: `$binDirHasPathSeparator` now guards both the User and process PATH mutations, and the success
output prints the PATH limitation plus a direct-run command instead of the restart-your-terminal line.
This mirrors the POSIX colon policy already in `install.sh`.

**3. [P3] Trailing newlines were stripped from custom POSIX paths.** `$(...)` removes all trailing
newlines, so an `ATOMIC_BIN_DIR` ending in a newline installed into the trimmed directory while the
requested path stayed absent — a silent rewrite of a stated environment knob.

Repair: `normalize_absolute_path` and `nearest_existing_directory` no longer emit a trailing newline, and
every capture site uses the `&& printf '_'` sentinel with `%_` stripping. `pwd` output keeps its own single
newline stripped through the new `$NEWLINE` constant, both at `START_WORKING_DIR` and inside
`canonicalize_existing_prefix`, so `cd -P`-derived paths survive too. Using `&&` rather than `;` keeps the
original failure status propagating to the existing `|| fail` handlers.

### TDD receipts

- RED (identity): `--ref 1.0.0` against a fixture answering `2.0.0` exited **0** and installed 2.0.0.
  GREEN after the comparison: exits non-zero, names both identities, and the request log shows the tags
  API call with no `releases/download/` request.
- RED (newline): with `ATOMIC_BIN_DIR` ending in a newline the install succeeded but
  `<binDir>/atomic` did not exist. GREEN: the launcher lands in the requested directory, the trimmed
  directory is never created, and the installed launcher runs `--version`.
- The two Windows repairs have always-on static contracts plus new gated fixture scenarios
  (`ref-identity`, `semicolon-bin`) that run on a real Windows PowerShell 5.1 host.

### Behavior change worth flagging

The Windows fixture's `install` scenario previously asserted that an explicit `-Ref "requested-alias"`
installed the API's canonical `1.0.0` — the opposite of finding 1. That assertion was the old contract and
is now wrong, so the scenario pins `1.0.0` directly and the alias case moved to the new `ref-identity`
scenario, where it must fail. The POSIX side already rejected `requested-alias` through the tag grammar.

### Iteration 8 validation receipts

- Focused installers: **34 passed, 13 skipped** (up from 31/11: three always-on additions, two new
  Windows-gated scenarios).
- CI contracts: **51 passed** (up from 48).
- Root unit **5,878 passed / 14 skipped**; integration **485 passed**; coding-agent **3,062 passed /
  29 skipped**.
- `npm run check`, `npm run hooks:run` (all hooks incl. Cargo fmt/clippy, Biome, check, unit suite),
  `npm run docs:check` (38 pages), `sh -n`, `dash -n`, Alpine 3.22 BusyBox `ash -n`, PowerShell 7.4
  Docker parse, and the Alpine + Debian restricted-PATH container installs all passed.
- The first coding-agent run failed one unrelated test,
  `bash-session-metadata > propagates the same snapshot into detached async jobs`. It is outside the
  delta (`git status` shows no `packages/coding-agent/src` or `test` change), passed 4/4 in isolation, and
  the full-suite rerun passed 3,062/29. Same load-sensitive async-bash class as the issue-5303 flake
  diagnosed in cycle 7. No source change was made for it.

### Base drift audit

`origin/main` advanced from `f9e672e4` to `536eb101` mid-run (Dependabot fixes, 0.9.12 and 0.9.12-alpha.1
changelogs, extension specifier canonicalization, README hero restructure). `git merge-tree --write-tree
HEAD origin/main` exited 0 with no conflicts, and the merged `README.md` extracted from that tree still
satisfies every README assertion in the docs contract. Main did not touch quickstart, windows, index,
containerization, or termux docs.

Decision: do not rebase. The branch did start from the latest fetched `origin/main` at creation; the merge
is verified clean; rebasing would rewrite commits the review loop cites by SHA and force a full
revalidation for no behavioral gain. Recorded rather than acted on. Merged-README detail: main's edit
touches the hero and capability list while the branch's edit touches Get started, Prerequisites, and
Install, so the hunks do not overlap; the merged file still carries both literal one-liners, the
``need `tar` and either `curl` or `wget` `` prerequisite, the physical-directory sentence, and the
stock-Alpine sentence, and does not carry the forbidden ``need `awk` `` string.

### Contract amendments received (inherited, cycle 7 — still binding)

> Convergence guidance for cycle 7: one cycle-6 reviewer approved with stop_review_loop=true and full evidence; the other introduced 11 findings, some of which conflict with the fixed task boundary. Prioritize literal issue #2159 acceptance and objective-backed safety defects: never delete a pre-existing unexpected launcher directory; reject impossible install/bin containment before I/O; preserve/report failed rollback restores; shell-quote PATH guidance; clean created empty parents; avoid token exposure if this can be done with allowed tools; and add focused regressions. Keep PowerShell 5.1/IEX behavior robust where the script can control it. Do NOT re-scope the task to gap-free concurrent launcher availability or installer locking—crash/concurrency handling was explicitly declared optional. Do not chase arbitrary Git refs beyond Atomic's supported release tag grammar if doing so adds forbidden dependencies; the new awk requirement conflicts with the issue's tool baseline and came from reviewer expansion around quote-containing tags. Distinguish the unsolvable pre-download TLS behavior of the literal irm bootstrap from logic inside install.ps1 rather than claiming the downloaded script can repair a connection that never fetched it. Treat UNC support as out of scope unless the ticket or existing docs promise it; a clear early rejection is enough if needed. Keep the graph bounded, run the exact acceptance tests/checks, remove issues.md when resolved, commit/push, and report any rejected reviewer findings with contract evidence.

No new user amendment arrived this cycle.

### Deferred (outside the contract, not implemented)

- Gap-free concurrent launcher availability and installer locking (declared optional).
- Arbitrary quote-containing Git refs beyond Atomic's stable/alpha tag grammar.
- UNC / network-path support on Windows.
- Pre-download TLS repair for the literal `irm` bootstrap (physically impossible from the fetched script).
- `atomic update` support for archive installs (issue #2159 names it as a separate PR).
- Hardening the post-exit idle heuristic behind the load-sensitive async-bash flakes (issue-5303 and
  `bash-session-metadata`); both are pre-existing and outside this delta.

### Iteration 8 delivery

- Committed `78fb36e4b4fb923d1570a2604585af85cbe8033a` as `fix(installer): pin exact refs and keep custom
  paths intact`, with the assistant-model trailer. Eight files, +222/-32.
- Pushed `feat/2159-curl-irm-installers`; local HEAD and the remote ref both resolve to `78fb36e4`.
- `origin/main` advanced again to `43557299` during the push. `git merge-tree --write-tree HEAD
  origin/main` still exits 0 with no conflicts, and the merged README keeps both literal one-liners and
  the `tar`/`curl`/`wget` prerequisite. The merge base remains `f9e672e4`; the no-rebase decision above
  still holds.
- Worktree clean, `gh pr list` returns `[]`, and the branch delta contains no `CHANGELOG.md` and no
  `config-self-update.ts`.

flora131:
## Implementation notes (full artifact)

Source artifact: `/Users/norinlavaee/.atomic/workflows/runs/ae3fe3e6-199b-4ba8-9770-8da0c52fd95a/artifact-b72d6903-ff25-4c8c-8664-1d2398d16ff6/implementation-notes.md`
HEAD at time of writing: `032d6bed`. Review did not converge; see the PR description for the unresolved blocking findings.

---

# Implementation Notes

Task: Continue issue #2159 on the existing branch feat/2159-curl-irm-installers in /Users/norinlavaee/atomic-curl-irm and repair the final review findings from draft PR https://github.com/bastani-inc/atomic/pull/2187. Freeze scope to these concrete items: (1) install.ps1 must detect and reject before download any pre-existing same-stem launcher such as ATOMIC_BIN_DIR\atomic.exe that Windows PATHEXT could resolve ahead of the installed atomic.cmd shim; add PowerShell 5.1-compatible static and native-gated regression tests. (2) install.sh must reject before network or filesystem mutation every impossible ATOMIC_BIN_DIR nested under transaction-owned paths such as ATOMIC_INSTALL_DIR/current or ATOMIC_INSTALL_DIR/versions/<resolved-tag>; add POSIX fixture tests proving zero requests and no mutation. Also fix the two directly relevant non-blocking correctness defects if reproduced: accept GNU sha256sum binary rows of the form HASH *asset, and restore a normal umask before creating install/bin/version payload paths so only temporary token/download state stays owner-only. Evaluate the Alpine APK URL durability note, but do not redesign release infrastructure unless a current deterministic failure violates issue #2159. Do not add concurrency locking, zero-downtime swap work, arbitrary Git-ref support, UNC support, pre-download TLS work, archive self-update, or other new scope. Remove temporary issues.md when resolved. Run focused installer suites, CI contracts, npm run check, container tests, and diff checks; commit and push focused repairs. Update the existing draft PR #2187 with accurate results and mark it ready only if all objective-required findings are resolved and fresh reviewers approve.

---

# Iteration C — four Windows PowerShell 5.1 repairs

## Contract amendments received

The research artifact carried a **new user amendment**, adopted as required behavior (verbatim):

> Include four reproduced Windows PowerShell 5.1 defects in final research/repair: (1) initialize $transaction before any preflight/API throw reaches finally; StrictMode masks the real error through an uninitialized variable. Test blocker/API errors keep original messages. (2) If effective PATHEXT omits .CMD, never report success when bare `atomic` cannot resolve; fail before I/O or use another contract-safe result, with native regression. (3) Enforce the supported Atomic semver/alpha tag grammar for PowerShell explicit refs and latest tags before archive requests, matching install.sh/docs; reject not-semver. (4) Never move/delete an unexpected regular installRoot\current directory or caller-owned entry; preflight type-check transaction pointers and preserve marker data. Keep this within issue #2159 safety/version/PATH scope and validate final committed HEAD.

The standing inherited clause about POSIX physical/dangling containment was restated too; the research
confirms it is already satisfied by `install.sh:107-143` / `:255-256` / `:269-283`, and my iteration-B
probe proved it independently. No new work was required for it.

## Reproduction before repair (real PowerShell 7.4, `--network none`, HEAD `0ea73cc3`)

| Finding | Reproduced? | Observed |
| ------- | ----------- | -------- |
| 1 — cleanup state initialized too late | **No** | The preflight throw surfaced verbatim (`ATOMIC_BIN_DIR contains atomic.exe, which PATHEXT resolves before atomic.cmd…`) and the API throw surfaced verbatim (`Failed to query GitHub release API at …`). Control flow explains it: the transaction `finally` belongs to the inner `try`, entered only *after* the initialization block, so a preflight or API throw never reached it. |
| 2 — `PATHEXT` without `.CMD` | **Yes** | `PATHEXT=".EXE;.BAT"` proceeded to the GitHub API instead of failing. |
| 3 — no tag grammar | **Yes** | `v1.0.0`, `1.0`, `1.0.0-alpha.0`, `1.0.0-beta.1`, `release/1.0` all reached the tags API. |
| 4 — regular `installRoot\current` | **Yes** | A regular `current` directory holding a marker file proceeded to the API with no preflight rejection. |

Finding 1 is an amendment clause, so I implemented the hardening anyway: it is memory-only, it makes
the invariant structural rather than incidental, and the error-preservation tests are worth having
either way. **I did not observe the defect it describes**, and say so rather than claiming a fix for
something I could not reproduce.

## Repairs (`install.ps1`)

1. **Cleanup state before the outer `try`.** `$tempDir`, `$versionStagePath`, `$currentNextPath`,
   `$atomicCurrentNextPath`, `$shimNextPath`, `$transaction`, `$transactionCommitted`,
   `$transactionMissingDirectories`, and `$rollbackRetryLimit` are initialized once, before
   `$previousSecurityProtocol`; the duplicate block after the download base was removed. The temp
   cleanup became null-safe (`$null -ne $tempDir -and (Test-Path …)`). No `catch` was added, so the
   original error record still propagates unchanged.
2. **Missing `.CMD`.** `Get-AtomicShimShadowingExtensions` tracks `$cmdSeen`, sets it at the `.CMD`
   break, and throws when the effective `PATHEXT` never lists `.CMD`. The call site is the existing
   preflight, so the throw lands before `$apiHeaders`, the temp directory, and every download.
   Normalization became `Trim().Trim('"').Trim()` so a quoted entry with internal whitespace still
   matches. Empty or whitespace `PATHEXT` still falls back to the stock list, which contains `.CMD`.
3. **Tag grammar.** New `Test-AtomicReleaseTag` uses
   `^(?:0|[1-9][0-9]*)\.(?:0|[1-9][0-9]*)\.(?:0|[1-9][0-9]*)(?:-alpha\.(?:[1-9][0-9]*))?$` with
   `-cmatch`, mirroring `install.sh:is_atomic_release_tag` exactly: no leading zeros, only an `-alpha.`
   prerelease, positive alpha revision. The requested ref is validated right after it is resolved from
   `-Ref`/`ATOMIC_VERSION` (before the API headers and the tags request); the resolved tag — from the
   redirect or the API — is validated before `$releaseBase`. Both throw the POSIX wording:
   `unsupported release tag: expected MAJOR.MINOR.PATCH or MAJOR.MINOR.PATCH-alpha.REVISION`.
4. **Transaction pointer types.** `$currentPath` and `$existingCurrentItem` are computed in the
   preflight, and a non-reparse `current` entry is rejected there. Reparse points (the installer's own
   junction, including a dangling one) still flow into the existing repair/rollback path, and the
   `atomic-current` guard is unchanged. Nothing is moved or deleted.

## Tests

- **Always-on static** (`test/unit/install-powershell.test.ts`): extended the PATHEXT shadow test with
  `$cmdSeen` and missing-`.CMD` ordering; added `Windows installer initializes cleanup state before
  preflight and API resolution` (each declaration appears exactly once and precedes the outer try),
  `Windows installer enforces Atomic release grammar before archive downloads`, and `Windows installer
  protects transaction pointer types before I/O`.
- **Engine-agnostic runnable probe**: new `preflightGuardProbeHarness` plus a per-engine test. It is
  fully offline — every case fails before any request — and covers missing `.CMD`, ten invalid explicit
  refs, an invalid `ATOMIC_VERSION`, a regular `current` directory and file with marker preservation,
  and preflight error-text preservation. **A local `pwsh` 7.6.4 exists on this host now, so it actually
  ran**: `PowerShell 7 enforces PATHEXT, tag grammar, and pointer guards before any request` passes,
  and against `git show 0ea73cc3:install.ps1` it fails with
  `PATHEXT without .CMD failed for the wrong reason: Failed to query GitHub…`.
- **Native Windows PowerShell 5.1 fixture scenarios**: `missing-cmd-pathext` (rejection plus `.cmd;.EXE`,
  `  " .CMD " ; ".EXE"  `, `.EXE;.CMD;.BAT`, and unset positive controls), `tag-grammar` (ten invalid
  refs, invalid `ATOMIC_VERSION`, valid `1.0.0` and `1.0.0-alpha.1`, invalid latest redirect tag,
  invalid latest API `tag_name`), `pointer-conflicts` (regular `current` and `atomic-current`, each as
  directory and file, with marker preservation, plus an installer-owned junction control upgrading
  1.0.0 → 2.0.0), and `preflight-errors` (blocker and API error text preserved, no strict-mode variable
  error). The harness now pins `$env:PATHEXT` to the stock default so scenarios do not inherit runner
  configuration.
- **Replaced** the `escaped-refs` scenario and its static test. It asserted that `release/1.0`,
  `hash#tag`, and `percent%tag` *install successfully*, which the amendment now forbids; its encoding
  coverage is unreachable once only semver tags are accepted. `ref-identity` moved from the non-semver
  `requested-alias` to `1.0.1`, since the grammar check would otherwise reject it before the identity
  mismatch could be exercised.
- **CI contract** (`test/ci/installers-safety-contract.test.ts`): new `Windows safety and version checks
  all precede temp creation and archive downloads` asserting all seven orderings the research listed,
  plus that the pointer preflight contains no `Move-Item`/`Remove-Item`/`New-Item`.
- **Docs**: `packages/coding-agent/docs/windows.md` gained the missing-`.CMD` and pointer-preservation
  sentences, with two matching assertions in the docs contract. That file's existing tag-grammar
  sentence was aspirational for PowerShell until this change and is now true.

## Iteration C validation

| Command | Observed |
| ------- | -------- |
| `pwsh` parser on `install.ps1` and on the extracted fixture harness | `PARSE_OK`, `HARNESS_PARSE_OK` |
| Engine-agnostic preflight probe under `pwsh` 7.6.4 | `PREFLIGHT_GUARDS_OK` (fails against `0ea73cc3`) |
| `npx vitest --run --project unit test/unit/install-shell.test.ts test/unit/install-powershell.test.ts` | **46 passed, 17 skipped** (was 41/14) |
| `npm run test:ci-contracts` | **56 passed** (was 55) |
| `npm run check` | pass (biome, `tsc --noEmit`, shrinkwrap) |
| `npm run test:unit` | **5890 passed, 18 skipped** (626 files) |
| `bash scripts/test-installers-containers.sh` | pass on stock `alpine:3.22` and `debian:bookworm-slim` |
| `sh -n install.sh`, `dash -n install.sh`, `git diff --check` | pass |

`install.sh` was not modified this iteration.

## Iteration C deferred

- **`cmd.exe` percent expansion at `install.ps1:762-763`** (both reviewer P3 findings). The research
  states plainly: "This is fail-closed, outside the frozen scope, and **must not be fixed in this repair
  cycle**." Not touched.
- Alpine apk revision-bump resilience; concurrency locking; arbitrary Git refs; UNC paths; pre-download
  TLS changes; archive self-update; release-infrastructure redesign.

---

# Iteration B (verification iteration) — HEAD 0ea73cc3

## Research artifact status

`research/2026-08-04-continue-issue-2159-...md` is **72 bytes** and contains exactly one line:
`Full unit suite passed: 5,882 tests passed, 17 skipped across 626 files.` That is a validation
statistic, not a findings list — the research stage again failed to record findings (0 bytes last
iteration, one stray line this one).

I recovered the authoritative findings from the review artifact instead:
`…/artifact-4b319e7a-dc06-47bf-974e-d86f87832936/review-round-latest.json`
(`convergence_decision`: `approved=false`, `stopReviewLoop=false`, `nextAction=implementation`;
6 consolidated findings, 5 blocking). No newer review artifact exists — the only file in the run
directory newer than that round is this stage's own transcript.

## Branch state found at start

`git log` shows six commits landed after my previous iteration's `87874c63`, all authored by a
later implementation cycle (`Assistant-model: GPT-5.6 Luna`):

```
0ea73cc3 fix(installer): support PowerShell 5.1 redirects
266c8a56 test(installer): honor common PowerShell parameters
7c3884bb test(installer): support Expand-Archive parameter aliases
d0c6ad43 fix(installer): cover preflight and hash fallbacks
75bfa390 fix(installer): support missing PowerShell hash cmdlet
52351057 fix(installer): close final review blockers
87874c63 fix(installer): refuse shadowed shims and transaction-owned bin paths   <- my prior cycle
```

Working tree clean; local HEAD == `origin/feat/2159-curl-irm-installers` == `0ea73cc3`.
PR #2187 is now **OPEN and no longer a draft**; every required check on `0ea73cc3` is SUCCESS.

Because the review round predates those six commits, this iteration's job was to **verify each
finding independently against HEAD** rather than re-implement, and to repair anything still
reproduced. Nothing was still reproduced, so **this iteration made no code changes**.

## Finding-by-finding verification at HEAD 0ea73cc3

| # | Finding (from `review-round-latest.json`) | Status | My evidence |
| - | ----------------------------------------- | ------ | ----------- |
| 1 | [P1, reviewer-a] Container test always fails cleanup on native Linux, keeping `static-checks` red | Resolved | `scripts/test-installers-containers.sh` now passes `--user "$(id -u):$(id -g)"` to both `docker run` calls. CI job `92167163701` (`static-checks (linux-x64)`, run 30960725545, HEAD `0ea73cc3`) ran `./scripts/test-installers-containers.sh`, logged both `Atomic 1.0.0 installed successfully.` lines, contains no `Permission denied` or `rm: cannot remove`, and concluded **success**. |
| 2 | [P1, reviewer-a] Dangling symlink components bypass the transaction-owned containment | Resolved | New `reject_dangling_symlink_path` in `install.sh`, called for both `ATOMIC_INSTALL_DIR` and `ATOMIC_BIN_DIR` after normalization. My own probe (below) reproduces the reviewer's exact scenario: RED at `87874c63`, GREEN at HEAD. |
| 3 | [P1, reviewer-b] Same container cleanup defect | Resolved | Same as #1. |
| 4 | [P1, reviewer-b] Native PowerShell 5.1 fixtures unrunnable — `Get-FileHash` not recognized, 13 failures | Resolved | New `Get-AtomicFileSha256` falls back to `[Security.Cryptography.SHA256]` when `Get-Command Get-FileHash` returns nothing. CI job `92167163390` (`suites (windows-x64)`) ran `test/unit/install-powershell.test.ts` with **33 tests, all passing on real Windows PowerShell 5.1**, including `PowerShell 5.1 fixture installs when Get-FileHash is unavailable`, `PowerShell 5.1 fixture refuses same-stem launchers that PATHEXT resolves before the shim` (the `shadowed-shim` scenario I authored last iteration), and `PowerShell 5.1 fixture honors custom PATHEXT order when .CMD precedes .EXE`. |
| 5 | [P2, reviewer-b] Dangling aliases into transaction-owned paths | Resolved | Same as #2; my probe covers both the `versions` and `current` alias variants. |
| 6 | [P3, reviewer-a, `beyond_objective`, confidence 0.55] Final smoke sends the shim path through `cmd.exe` unescaped, so a `%…%` token in `ATOMIC_BIN_DIR` is environment-expanded | **Deferred, not repaired** | The code pattern is confirmed present at `install.ps1:762-763` (`$shimCommand = '"' + $shimPath + '" --version'` then `& $env:ComSpec /d /c $shimCommand`). I could not execute it — no Windows and no `cmd.exe` on this host. Reasoning for deferring is below. |

### Dangling-alias probe (findings 2 and 5) — RED/GREEN

Stub `curl` on `PATH` logging every URL; `ATOMIC_INSTALL_DIR=$ROOT` where `$ROOT` does not exist;
`outside/versions-alias -> $ROOT/versions` and `outside/current-alias -> $ROOT/current` created as
dangling symlinks; `ATOMIC_BIN_DIR=outside/<alias>/bin`.

- **RED** at `87874c63` (`git show 87874c63:install.sh`), both aliases:
  `error: failed to resolve the GitHub release`, request log
  `GET https://api.github.com/repos/bastani-inc/atomic/releases/tags/1.0.0` — the run reached the
  network, exactly as the reviewer reported.
- **GREEN** at HEAD, both aliases: exit 1,
  `error: ATOMIC_BIN_DIR contains an unresolved symbolic link; refusing an unresolvable path: …`,
  **empty request log**, and `$ROOT` never created.
- **Over-rejection control**: with `outside/live-alias -> ` an existing directory, the installer
  passes the preflight and proceeds to the network (fails only on the stub), so a live symlinked
  bin directory is still accepted.

### Windows shadow rejection re-proved after a semantics change

`Get-AtomicShimShadowingExtensions` changed between my last verification and HEAD: it previously
unioned `$env:PATHEXT` with the stock list, and now uses the effective `$env:PATHEXT` alone, falling
back to the stock list only when `PATHEXT` is unset or blank. That is the more accurate rule — the
union could reject an `atomic.exe` in a session whose `PATHEXT` puts `.CMD` first, which is a false
rejection, and the new native test `honors custom PATHEXT order when .CMD precedes .EXE` pins the
corrected behavior. I re-ran the persisted probe harness (extracted verbatim from
`test/unit/install-powershell.test.ts`) against the current `install.ps1` under real PowerShell 7.4
with `--network none`: `SHIM_SHADOW_OK` for `atomic.exe`, `atomic.com`, `atomic.bat`, `ATOMIC.EXE`,
a same-stem directory, and a custom `PATHEXT=".WSF;.CMD;.EXE"` with `atomic.wsf`. `install.ps1`
parses clean (`PARSE_OK`).

### Prior-cycle repairs still present at HEAD

Verified by inspection: `for owned_child in current versions` preflight (`install.sh:269-281`),
`ORIGINAL_UMASK=$(umask)` / `umask 077` (388-389) with `umask "$ORIGINAL_UMASK"` (833) before the
extract and every install-path `mkdir`, and the GNU binary-row pattern `\**)` (803). The container
fixture now writes `HASH *asset` rows, so the stock Alpine/Debian smoke exercises binary-mode
checksums end to end.

## Validation run this iteration (all at HEAD 0ea73cc3, no local changes)

| Command | Observed |
| ------- | -------- |
| `sh -n install.sh`, `dash -n install.sh` | pass |
| PowerShell parser on `install.ps1` (docker pwsh 7.4) | `PARSE_OK` |
| Persisted shadow probe harness under docker pwsh 7.4, `--network none` | `SHIM_SHADOW_OK` |
| `npx vitest --run --project unit test/unit/install-shell.test.ts test/unit/install-powershell.test.ts` | **41 passed, 14 skipped** |
| `npm run test:ci-contracts` | **55 passed** |
| `npm run check` (biome + `tsc --noEmit` + shrinkwrap) | pass |
| `npm run test:unit` | **5885 passed, 15 skipped** (626 files) |
| `bash scripts/test-installers-containers.sh` | pass on stock `alpine:3.22` BusyBox ash and `debian:bookworm-slim` |
| `git merge-tree --write-tree HEAD origin/main` | exit 0, no conflicts (`origin/main` `43557299`, merge base `f9e672e4`) |
| `ls issues.md` | absent |
| `git status --short` | clean |

CI at HEAD (run `30960725545`): all nine jobs success — `static-checks (linux-x64)`,
`suites`/`agent-suite`/`test`/`release-archive` on both `linux-x64` and `windows-x64`.

## Deferred (outside the contract, not implemented)

- **Finding 6, `cmd.exe` percent expansion in the final smoke.** The pattern is real at
  `install.ps1:762-763`, but I did not repair it: the raising reviewer marked it
  `beyond_objective` at confidence 0.55 and wrote that "it should not gate this change"; the
  objective freezes scope to the two blockers plus the checksum and umask defects; it fails closed
  (rolls back rather than installing something wrong) and needs a bin directory containing a
  literal `%NAME%` pair naming a defined variable; and a new commit would invalidate a fully green
  CI run on a PR that is already out of draft. Recorded here so the decision is visible rather than
  silently dropped.
- Alpine apk revision-bump resilience in `scripts/build-binaries.sh`. Latent; all four pinned URLs
  returned HTTP 200 when checked last iteration.
- Carried forward: concurrency locking and gap-free launcher availability, arbitrary
  quote-containing Git refs, UNC support, pre-download TLS repair for the literal `irm` bootstrap,
  `atomic update` for archive installs.
- A docs clause naming the new `unresolved symbolic link` message. Not added: the existing
  quickstart/README wording ("impossible layouts fail before any download or filesystem change")
  already covers it truthfully, and `windows.md`'s PATHEXT sentence is *more* accurate under the
  effective-PATHEXT implementation than it was under the union. No documentation is stale or wrong.

## Contract amendments received

None. No user steering or follow-up message arrived during this iteration. The cycle-7 amendment
quoted in earlier notes remains inherited and binding; findings 2 and 5 repaired in the prior cycle
are its "reject impossible install/bin containment before I/O" clause.

## PR status

PR #2187 is OPEN, out of draft, head `0ea73cc3`, all required checks SUCCESS, `reviewDecision`
empty (no submitted human review). **I did not edit the PR**: the acceptance criteria reserve PR
action for `create_pr=true`, and this stage's constraints say to ignore requests to submit a PR.
A previous implementation cycle already updated the body and marked it ready.

## QA E2E Video

Not applicable. This is a terminal-only CLI installer with no web, frontend, or TUI surface, so
`playwright-cli` video does not apply and none was produced. The end-to-end evidence is the stock
container installs (`alpine:3.22`, `debian:bookworm-slim`) locally and in CI, the network-disabled
real PowerShell 7.4 probes, and — new this iteration — 33 passing native Windows PowerShell 5.1
fixture tests in CI job `92167163390`.

---

# Iteration A (previous, HEAD 87874c63) — retained for history

## Repairs

1. **[BLOCKING] Windows same-stem PATHEXT shadow (`install.ps1`).** Added
   `Get-AtomicShimShadowingExtensions` plus a preflight that throws before the API headers are
   built. Same-stem directories are rejected too, because `SearchPath` matches directories.
   Nothing is deleted — the installer reports and stops.
2. **[BLOCKING] POSIX bin path under transaction-owned install paths (`install.sh`).** Preflight
   rejecting `BIN_PATH`/`PHYSICAL_BIN_PATH` at or under `<root>/current` or `<root>/versions`, for
   both normalized and physical roots, before the required-command probe, the temp directory, and
   any request. `$INSTALL_ROOT` and `$INSTALL_ROOT/bin` stay accepted (positive control).
3. **GNU binary checksum row.** `case $checksum_name in \*)` → `\**)`.
4. **umask spillover.** Capture the original umask, keep `077` for the temp directory, download and
   token file, restore it immediately before `mkdir "$EXTRACT_ROOT"`.
5. **Alpine APK durability.** Evaluated only; all four pinned URLs returned HTTP 200, so no current
   deterministic failure and no redesign.

## RED/GREEN receipts (iteration A)

- POSIX containment: RED reproduced reviewer-b's exact
  `Too many levels of symbolic links` / `installed atomic --version check failed` after all three
  requests. GREEN: rejected with an empty request log and empty TMPDIR.
- GNU binary row: RED `error: SHA256SUMS must contain exactly one row for atomic-linux-x64.tar.gz`.
- umask: RED stock Alpine `drwx------` for bin and install root; GREEN `drwxr-xr-x`, payload
  `-rw-r--r--`.
- Windows shadow: RED `atomic.exe failed for the wrong reason: Failed to query GitHub release API`;
  GREEN `SHIM_SHADOW_OK`.

## Iteration A validation

Focused installers 38 passed / 15 skipped; CI contracts 54 passed; `npm run check` pass; unit 5882
passed / 16 skipped; integration 485 passed; coding-agent 3062 passed / 29 skipped (one
load-sensitive flake in `model-registry-dynamic-providers` that also passed in isolation and on a
clean tree, untouched by the delta). Committed `87874c63`, nine files, +454/-7.

flora131:
Implementation notes artifact for this run, reproduced in full as the last handoff action. Review did **not** converge (2 iterations); one blocking P2 finding at `install.ps1:940` remains unresolved and is listed in the PR body under "Unresolved review findings". This PR stays a draft.

---

# Implementation Notes

Task: Repair the single remaining blocker on draft PR https://github.com/bastani-inc/atomic/pull/2187 in /Users/norinlavaee/atomic-curl-irm on branch feat/2159-curl-irm-installers. Freeze scope to Windows temporary-directory cleanup in install.ps1 and its native PowerShell 5.1 regression. At committed HEAD 032d6bed, fresh Windows CI job 92181276918 failed its first unit attempt because the missing-cmd-pathext fixture root retained atomic-install-* residue; only the bounded whole-suite retry passed. install.ps1 currently removes $tempDir in finally with Remove-Item -Recurse -Force -ErrorAction SilentlyContinue after executing staged atomic.exe from that directory. Reproduce the race or create a deterministic one-shot deletion-failure probe, implement bounded verified temp-dir removal that handles delayed Windows process/file handles, preserves the original installer error when cleanup also fails, reports incomplete cleanup clearly if retries exhaust, and does not weaken rollback or caller-owned path safety. Add focused always-on contracts plus native Windows PowerShell 5.1 regression/stress coverage. Do not change any other installer behavior or add new findings/scope. Run focused installer tests, CI contracts, npm run check, Windows-native CI without relying on the suite retry, container smoke, and diff checks. Commit and push the focused repair, update PR #2187 accurately, and mark it ready only if fresh first-attempt Windows evidence and both reviewers approve.

## Running Notes

- Record implementation decisions, deviations from research, tradeoffs, blockers, validation outcomes, and user-relevant facts. Keep entries concise and readable.
- Before reporting progress, audit each claim against a tool result from this session. Report only work you can point to evidence for; say so explicitly when something is unverified.

## Root cause: corrected from the research artifact

`gh run view --log --job=92181276918` (fetched this session, saved to `/tmp/job92181276918.log`)
shows the attempt-1 failure was **not** at `test/unit/install-powershell.test.ts:1533`:

```
FAIL unit test/unit/install-powershell.test.ts > PowerShell 5.1 fixture rejects PATHEXT without .CMD before any request or mutation
AssertionError: PowerShell fixture missing-cmd-pathext failed.
stderr:
Fixture assertion failed: temporary installer directory was not cleaned
1 !== 0
```

`"temporary installer directory was not cleaned"` is the **end-of-scenario** assertion at
`test/unit/install-powershell.test.ts:2163`, which runs after the four valid installs in the
PATHEXT-acceptance loop (`:1535-1542`). So the residue was left by a *successful* install that
staged and executed `payload\atomic.exe`, not by the missing-`.CMD` rejection. Research
conclusion 1 (the rejection cannot create `$tempDir`) is confirmed and consistent; research
conclusion 2's framing of the failing line is wrong. Research conclusion 3 — silent, unverified
`Remove-Item -ErrorAction SilentlyContinue` cleanup at `install.ps1:848-850` — is the confirmed
production defect and is exactly what this repair fixes.

## Acceptance matrix

| # | Contract clause | Verification |
|---|---|---|
| 1 | Every install attempt removes its `atomic-install-*` temp directory after staged execution | `Remove-AtomicTemporaryDirectory` bounded verified removal + fixture `temp-cleanup` stress case asserting zero residue after each attempt |
| 2 | Deterministic one-shot deletion-failure probe | `temp-cleanup` case 1: real `FileShare.None` handle, released only after the first genuine failed removal of the exact temp root |
| 3 | Cleanup failure must not mask the primary error | `temp-cleanup` case 3: final-smoke failure + permanently locked temp; surfaced error is the primary one, cleanup evidence is a warning |
| 4 | Exhausted cleanup fails loudly with evidence | `temp-cleanup` case 2: sticky lock; installer throws naming exact path, attempt count, last exception |
| 5 | Rollback and caller-owned path safety unchanged | cleanup stays at its existing position; created-parent cleanup still runs before the cleanup error is surfaced; helper takes one exact path, no wildcard, no install/bin root deletion |
| 6 | Always-on contracts | `test/unit/install-powershell.test.ts` source contracts + `test/ci/installers-safety-contract.test.ts` mirror |
| 7 | Native PowerShell 5.1 regression | new `temp-cleanup` fixture scenario (Windows-only), must pass on the first CI attempt |

## Implementation

`install.ps1`

- New `Remove-AtomicTemporaryDirectory $Path $RetryLimit $RetryDelayMilliseconds` (declared after
  `Remove-AtomicDirectoryLinkOrTree`). Per attempt: clear `ReadOnly` on the root and every entry,
  `Remove-Item -Recurse -Force -ErrorAction Stop`, verify with `[IO.Directory]::Exists`, fall back to
  `[IO.Directory]::Delete($Path, $true)`, verify again, then linear backoff. On exhaustion it throws
  naming the exact path, the attempt count, and the last exception. It returns immediately for a null,
  blank, or missing path, so the preflight rejection paths (where `$tempDir` stays `$null`) are untouched.
- New state before the outer `try`: `$tempCleanupRetryLimit = 5`, `$tempCleanupRetryDelayMilliseconds = 125`
  (bounded backoff total 1.25 s), `$primaryError = $null`, `$tempCleanupError = $null`.
- The staged-work `try` gained `catch { $primaryError = $_; throw $primaryError }` — the same
  capture-and-rethrow shape the transaction catch at the old `:800-805` already used and that the
  real-console Ctrl+C fixture already exercises.
- Temp cleanup stays at its existing position in `finally`; its failure is captured, not thrown, so
  rollback, staged-pointer cleanup, and `Remove-AtomicCreatedEmptyDirectories` all still run. At the end
  of `finally`: warn with cleanup evidence if a primary error exists, otherwise throw the cleanup error.

`test/unit/install-powershell.test.ts`

- Always-on source contract: helper declaration, named finite limit, no unbounded loop, read-only
  clearing, terminating `Remove-Item`, `[IO.Directory]::Delete(..., $true)` fallback, two post-attempt
  existence verifications, path/attempt/last-error evidence, helper used for `$tempDir`, ordering
  (temp cleanup → created-parent cleanup → deferred report), and no removal/deletion of any path other
  than `$Path`.
- Fixture harness: `Get-TempResidueReport` (reports full leftover paths, not only a count), a recording
  `Write-Warning` wrapper, a pass-through `Remove-Item` wrapper that counts removals of the exact
  `atomic-install-<32 hex>` root and releases the one-shot handle before the second removal, and a
  `New-Item` hook that opens a real `FileShare.None` handle inside the temp root when a lock mode is armed.
- New Windows-only `temp-cleanup` scenario: direct `GetTempPath()` root assertion; one-shot real handle
  released only after a genuine failure; sticky handle exhausting the bounded budget with path/attempt/
  last-error evidence while the completed install survives; primary final-smoke error plus failing cleanup
  (primary error wins, cleanup warned once, rollback and created-parent cleanup still complete); and a
  stress pass of two valid installs plus one rolled-back install asserting zero residue after each.
- `TEMP_CLEANUP_FIXTURE_STRUCTURAL_TIMEOUT_MS = 180_000` at the call site. Justified as structural: a real
  Windows PowerShell 5.1 child that compiles a C# fixture executable, builds four ZIP releases, runs six
  real installs, and deliberately exhausts two bounded cleanup budgets against real open Win32 handles.

`test/ci/installers-safety-contract.test.ts` mirrors the safety-critical checks.
`packages/coding-agent/docs/windows.md` documents the new user-visible cleanup behavior.

## Validation (this session)

| Command | Result |
|---|---|
| `pwsh` parse of `install.ps1` | `PARSE_OK` |
| `pwsh` parse of the extracted fixture harness | `HARNESS_PARSE_OK` |
| `/tmp/probe-temp-cleanup.ps1` (helper behavior probe) | `NOOP_OK`, `NULL_OK`, `REMOVE_OK`, `EXHAUSTED_OK` with path + `after 3 attempts` + last error |
| `/tmp/probe-remove-item-wrapper.ps1` (fixture wrapper fidelity) | `SUPPRESSED_OK`, `INHERIT_OK`, `STOP_OK`, `RECURSE_OK`, `COUNT_OK` |
| `/tmp/probe-one-shot-release.ps1` (retry ↔ release interaction) | `ONE_SHOT_RETRY_OK after 2 removal calls` |
| `npm run test:unit -- test/unit/install-shell.test.ts test/unit/install-powershell.test.ts` | 48 passed, 18 skipped (Windows-only fixtures) |
| `npm run test:ci-contracts` | 57 passed |
| `npm run check` | biome clean, `tsc --noEmit` clean, shrinkwrap up to date |
| `bash scripts/test-installers-containers.sh` | Alpine and Debian installs succeeded |
| `sh -n install.sh`, `dash -n install.sh`, `git diff --check` | all clean |

The local probes ran under pwsh 7 on macOS; they prove the helper's control flow, the fixture wrapper's
parameter fidelity, and the retry/release interaction, **not** Windows sharing-violation semantics. Native
Windows PowerShell 5.1 evidence must come from CI attempt 1.

## Decisions and deferrals

- No `packages/*/CHANGELOG.md` entry. No commit on `feat/2159-curl-irm-installers` has touched a changelog
  (`git diff --name-only origin/main...HEAD` lists only docs), so the installer feature's release notes are
  not yet written on this branch; adding an entry for this sub-fix alone would be inconsistent and outside
  the frozen scope. Deferred to whoever writes the feature's changelog entry.
- Rename-then-delete quarantine was considered and rejected, per research §2: a rename can be blocked by the
  same handle, and the renamed tree still has to be deleted, so it adds a failure mode without adding proof.

## QA E2E Video

Not applicable, and no video was produced. This change touches `install.ps1` (a Windows PowerShell
installer script) and its test coverage. There is no web or frontend surface for `playwright-cli` and no
TUI flow for `tmux`: the installer is non-interactive, prints plain text, and its user-visible behavior is
Windows-only, so it cannot be exercised on this macOS host at all. The executable end-to-end evidence for
this change is the native Windows PowerShell 5.1 `temp-cleanup` fixture running under real Windows CI,
which performs six real installs with real open Win32 handles. Local end-to-end proof of the cleanup
control flow came from the three pwsh probes recorded in the validation table above.

## Signing note

The commit is unsigned. `commit.gpgsign=true` with `gpg.ssh.program=op-ssh-sign` is configured, but the
1Password signer failed with `1Password: failed to fill whole buffer` on every attempt, including a direct
`op-ssh-sign -Y sign` invocation outside git, so it needs an interactive desktop unlock that is not
available here. Two existing commits on this branch (`0ea73cc3`, `266c8a56`) are also unsigned, so this is
consistent with the branch. Re-sign with `git commit --amend -S` after unlocking 1Password if a signed
commit is required.

## Fresh Windows CI evidence (first attempt, no suite retry)

Commit `ee121f5f` pushed to `feat/2159-curl-irm-installers`. Run
[30970296937](https://github.com/bastani-inc/atomic/actions/runs/30970296937) concluded **success**;
every job green.

`suites (blacksmith-4vcpu-windows-2025, windows-x64)` job `92192947703`:

- `Test Files 622 passed | 9 skipped (631)` / `Tests 5887 passed | 59 skipped (5946)`.
- The only report written was `.ci-diagnostics/unit-tests-windows-x64-attempt-1.json`. There is no
  attempt-2 report, no `Retrying`, and no `Attempt 1 failed; bounded retry passed` annotation on this job,
  so the bounded suite retry was **not** consumed.
- `✓ test/unit/install-powershell.test.ts (45 tests) 42218ms` — 42 before, 45 now (three added tests).
- `✓ PowerShell 5.1 fixture removes every temporary download directory against real open handles 4817ms` —
  the new native Windows PowerShell 5.1 fixture ran and passed. 4817 ms is 2.7 % of its declared
  180 000 ms structural budget, far below the 40 % warn line, so the duration gate stayed silent.
- `✓ PowerShell 5.1 fixture rejects PATHEXT without .CMD before any request or mutation` passed on
  attempt 1 — the exact test that failed at `032d6bed`.

The run's single retry annotation belongs to a different job, `agent-suite (linux-x64)` `92192947713`:
`test/tools-04-01.suite.ts > Coding Agent Tools > bash tool > should not count a trailing newline as an
extra truncated bash output line`, `AssertionError: expected 'line-1001…' not to contain '4001'`. That is
the coding-agent bash-output truncation suite, unrelated to the installers and untouched by this change.
Recorded as an observation, not a finding to fix in this frozen scope.

---

## Iteration 2: no new findings to repair

The research artifact was rewritten to a single line for this iteration:

> The online-research subagent timed out, but the report's cited Microsoft documentation and local
> PowerShell probes were completed directly. No repository changes were made.

There are no consolidated unresolved findings in it, so there was no finding batch to repair. State
checked directly instead:

- `git status --short --branch` — clean, `feat/2159-curl-irm-installers` in sync with origin.
- `git log --oneline -3` — HEAD is `ee121f5f`, the temp-cleanup repair.
- `gh pr view 2187` — `headRefOid` is `ee121f5f` (the PR points at the repair), still draft, `reviewDecision`
  empty, and the only review on record is a `greptile-apps` `COMMENTED` review against the older commit
  `0ea73cc3`. No new blocking review round has been posted since the repair landed.
- `gh run list` — the newest run for `ee121f5f` is still `30970296937`, conclusion `success`. No re-run and
  no later run has changed that evidence.

### Re-verification against the committed checkout (this session)

| Command | Result |
|---|---|
| `npm run test:unit -- test/unit/install-shell.test.ts test/unit/install-powershell.test.ts` | 48 passed, 18 skipped |
| `npm run test:ci-contracts` | 57 passed, 8 files |
| `npm run check` | biome clean, `tsc --noEmit` clean, shrinkwrap up to date |
| `pwsh` parse of `install.ps1` | `INSTALLER_PARSE_OK` |
| `pwsh` parse of the re-extracted fixture harness | `HARNESS_PARSE_OK` |
| `/tmp/probe-temp-cleanup.ps1` | `NOOP_OK`, `NULL_OK`, `REMOVE_OK`, `EXHAUSTED_OK` |
| `/tmp/probe-remove-item-wrapper.ps1` | `SUPPRESSED_OK`, `INHERIT_OK`, `STOP_OK`, `RECURSE_OK`, `COUNT_OK` |
| `/tmp/probe-one-shot-release.ps1` | `ONE_SHOT_RETRY_OK after 2 removal calls` |
| `grep -n "Remove-AtomicTemporaryDirectory" install.ps1` | one declaration (`:276`), one call site (`:927`), for `$tempDir` only |
| `grep -n 'Remove-Item -LiteralPath $tempDir' install.ps1` | no match — the suppressed deletion is gone |

### Divergence audit against the current checkout

| Class | Probe | Outcome |
|---|---|---|
| States and transitions | `$tempDir` null (preflight reject), directory absent, removal succeeds first try, fails then succeeds, never succeeds without a primary error, never succeeds with a primary error | All six covered: first four by the local probes, the last two by the Windows `temp-cleanup` fixture cases that passed on CI |
| Omitted / empty / zero inputs | helper called with `$null` and with a missing path | Returns silently (`NULL_OK`, `NOOP_OK`); no probing or mutation of any caller path |
| Unusual filesystem input | read-only file inside the tree | Cleared and removed (`REMOVE_OK`) |
| Caller-controlled paths and precedence | `TEMP`/`TMP` redirection; helper restricted to one exact path | Fixture asserts the resolved `[IO.Path]::GetTempPath()` root directly; both test files assert the helper never removes or deletes a path other than `$Path`, and both suites pass |
| Enumerated errors | exhausted cleanup | Throws with exact path, attempt count, and last exception; asserted character-for-character by the source contracts and by the native fixture |
| Cross-state invariant | cleanup failure must not stop later cleanup | `Remove-AtomicCreatedEmptyDirectories` runs before the deferred report; ordering asserted in both test files and exercised by fixture case 3 |

Verified by inspection of `install.ps1:918-946` in this session: staged-pointer cleanup → temp cleanup
(captured, not thrown) → created-parent cleanup → deferred warn-or-throw.

### Deferred: greptile P1 comment, out of frozen scope

`gh api repos/bastani-inc/atomic/pulls/2187/comments` returns one bot comment, posted 2026-08-05T00:31:48Z
against commit `0ea73cc3` at `install.ps1:649` (prompt block cites lines 542-548):

> **Transaction-owned directories can be used as the shim location** — `ATOMIC_BIN_DIR` is normalized but
> never checked against `<installRoot>\current` or `<installRoot>\versions`.

The claim is factually accurate. `install.sh:269-278` has the guard
(`ATOMIC_BIN_DIR cannot be inside ATOMIC_INSTALL_DIR/$owned_child` for `current` and `versions`), and
`grep` finds no equivalent in `install.ps1`. So POSIX and Windows are asymmetric here.

It is **not** repaired in this iteration, deliberately. The objective freezes scope to "Windows temporary-
directory cleanup in install.ps1 and its native PowerShell 5.1 regression" and says "Do not change any
other installer behavior or add new findings/scope"; the acceptance criteria say "The only remaining
acceptance delta is deterministic Windows cleanup" and "No new scope." Under the literal contract the
objective controls over reviewer comments. It also predates the review round that produced the two
blocking findings this branch was returned to draft for: that round's consolidated batch, reproduced in the
PR body, contains exactly two findings, both about temp cleanup, both now fixed. Recorded here so the next
scope decision is made deliberately rather than by omission.

PR #2190, opened 2026-08-04
fix(tools): object-root union tool parameter schemas so they advertise parameters

### LINKED ISSUE #2189: Union-rooted tool parameter schemas are advertised with no parameters, making the tool uncallable
### What happened?

A tool whose `parameters` schema is union-rooted (`Type.Union([...])`) is advertised to the provider with **no parameters at all**, while validation still enforces the full union. The tool becomes uncallable.

`convertTools` builds the advertised `input_schema` from the root `properties`/`required` keywords alone — `@earendil-works/pi-ai/dist/api/anthropic-messages.js:1001`:

```js
const legacyInputSchema = {
    type: "object",
    properties: schema.properties ?? {},
    required: schema.required ?? [],
};
const inputSchema = strict === true ? { ...tool.parameters, ...legacyInputSchema } : legacyInputSchema;
```

A `Type.Union` root serializes to `{anyOf}` and carries neither keyword, so the tool ships as:

```json
{"name":"structured_output","input_schema":{"type":"object","properties":{},"required":[]}}
```

Note the `strict === true` branch spreads `legacyInputSchema` **last**, so it clobbers `properties`/`required` back to empty too. No provider or sampling setting rescues it.

Any client that types tool arguments from the advertised schema then sends every argument as a string, because no argument is declared. Validation — which still enforces the union — rejects them:

```
Validation failed for tool "structured_output":
  - options: must be array
  - root: must match a schema in anyOf
```

Measured on `typebox@1.3.7` with a three-branch union root:

```
advertised: {"type":"object","properties":{},"required":[]}
```

This is invisible to a client that emits raw JSON tool input, which is why it has not surfaced before.

**Blast radius.** It makes `createStructuredOutputTool({ schema })` unusable with a discriminated-union schema, which is the natural shape for a "return one of N turn kinds" contract. `packages/workflows` loops while `!structuredOutputCapture.called` and throws `STRUCTURED_OUTPUT_MISSING_ERROR` after 3 corrective prompts (`stage-runner-context.ts`), so the stage fails hard rather than degrading. Every branch of such a union requires a container (an array or an object), so **no valid call is constructible** from string-only arguments.

Two secondary defects in the same path:

1. **No coercion for union roots.** `validateToolArguments` calls `Value.Convert(tool.parameters, args)` for its in-place effect and discards the return value. TypeBox's union conversion works on a clone, so a union root gets nothing applied. Measured: `Value.Convert` leaves `42` as `42` on a union root but yields `"42"` on an object root. (In practice this is currently masked — `validateToolArguments` skips its `coerceWithJsonSchema` fallback only for schemas carrying the `TypeBox.Kind` own-symbol, and TypeBox 1.3.7 schemas carry no such symbol, so the fallback runs and coerces union roots anyway.)
2. **Nothing warns at registration.** The schema is accepted silently and fails at turn time, far from the cause.

### Steps to reproduce

```ts
import { Type } from "typebox";
import { createStructuredOutputTool } from "@bastani/atomic";

const schema = Type.Union([
  Type.Object({ kind: Type.Literal("question"), options: Type.Array(Type.String()) }, { additionalProperties: false }),
  Type.Object({ kind: Type.Literal("done"), result: Type.Object({ ok: Type.Boolean() }) }, { additionalProperties: false }),
]);

const tool = createStructuredOutputTool({ schema });

// What the provider is told the tool accepts:
console.log({ properties: tool.parameters.properties ?? {}, required: tool.parameters.required ?? [] });
// => { properties: {}, required: [] }
```

Register it on a session and prompt for a structured turn. Every attempt arrives with `options` as a JSON **string** and is rejected against the union; the step exhausts its corrective attempts.

### Expected behavior

A union of object-rooted branches should be advertised with real `properties`, so clients type arguments correctly — without loosening validation.

Concretely, the root can be rewritten to an object whose properties are the merged branch properties, with the original branches retained under `anyOf`:

```
{anyOf:[A,B]}  ->  {type:"object", properties:{...merged, optional}, required:[...common], anyOf:[A,B]}
```

`properties` makes it advertisable; the retained `anyOf` keeps validation exactly as strict, so the set of accepted arguments is unchanged. Two details matter: TypeBox encodes optionality with a marker rather than the `required` array (a raw JSON merge over-requires and rejects every branch but one), and a discriminator that differs across branches must be unioned rather than first-wins (otherwise the root is pinned to one branch's literal).

A root that is neither object-rooted nor a union of object-rooted branches cannot be advertised at all and should warn at registration instead of failing silently at turn time.

### Version

0.9.11

elefthei:
Closes #2189

## Problem

A tool whose `parameters` schema is union-rooted is advertised to the provider with **no parameters at all**, while validation still enforces the full union — so the tool is uncallable.

`convertTools` builds the advertised `input_schema` from the root `properties`/`required` keywords alone. A `Type.Union` root serializes to `{anyOf}` and carries neither, so the tool ships as `{"type":"object","properties":{},"required":[]}`. Clients that type tool arguments from the advertised schema then send every argument as a string; validation rejects them against the union that is still enforced. Every branch of a discriminated union requires a container, so no valid call is constructible.

This makes `createStructuredOutputTool({ schema })` unusable with a discriminated-union schema — the natural shape for a "return one of N turn kinds" contract — and a workflow stage with such a schema fails hard with `STRUCTURED_OUTPUT_MISSING_ERROR` after exhausting its corrective prompts. It is invisible to clients that emit raw JSON tool input, which is why it has not surfaced before.

## Solution

Rewrite a union root whose branches are all object-rooted into an equivalent object root:

```
{anyOf:[A,B]}  ->  {type:"object", properties:{...merged, optional}, required:[...common], anyOf:[A,B]}
```

`properties` makes it advertisable; the retained `anyOf` keeps validation exactly as strict, so **the set of accepted arguments is unchanged**.

Applied in `wrapToolDefinition` — the single adapter every tool passes through (built-ins, SDK `customTools`, extension `registerTool`, MCP, `createStructuredOutputTool`), and the one place where the advertised schema and the validated schema are guaranteed to be the same object.

Two details are load-bearing:

- **TypeBox encodes optionality with a marker, not the `required` array.** A raw JSON merge over-requires and rejects every branch but one, so merged properties are wrapped in `Type.Optional`.
- **A discriminator differs across branches.** First-wins merging pins `kind` to one branch's literal and rejects the others, so colliding keys are unioned.

Rewrites are memoized per authored schema, so repeated registry refreshes reuse one schema — and therefore one entry in the provider's identity-keyed compiled-validator cache.

Object-rooted schemas are returned unchanged (identity, no allocation). A root that is neither object-rooted nor a union of object roots cannot be advertised at all; it is left as authored and warns **once** at registration instead of failing silently at turn time.

The rewritten root also restores the in-place argument coercion that `Value.Convert` cannot apply to a union. Note this part is currently masked: `validateToolArguments` skips its `coerceWithJsonSchema` fallback only for schemas carrying the `TypeBox.Kind` own-symbol, and TypeBox 1.3.7 schemas carry no such symbol, so the fallback runs and already coerces union roots. Measured both ways.

## Tests

`packages/coding-agent/test/tool-parameter-schema.test.ts` (new, 8 tests). The load-bearing one is a differential table asserting the rewritten root accepts **exactly** what the union accepted, across valid branches, option-count boundaries, undeclared root/nested properties, cross-branch field leaks, a bare discriminator, and a container sent as a string.

```
npm run test --workspace=@bastani/atomic -- test/tool-parameter-schema.test.ts
  Test Files  1 passed (1)
       Tests  8 passed (8)

npm run check
  biome check --error-on-warnings .   -> clean (2376 files)
  tsc --noEmit                        -> clean
  check:shrinkwrap                    -> up to date
```

### Regression check

This change sits in the adapter every tool passes through, so I ran the suites on this branch **and** on `main` and compared the failing file *sets* rather than reading a raw pass count.

| suite | `main` | this branch | failing only on this branch |
| --- | --- | --- | --- |
| `@bastani/atomic` package | 13 files / 30 tests | 13 files / 29 tests | **none** |
| root `test:unit` | 11 files | 13 files | 2, both flaky (below) |

The package suite'''s failing set was **identical** to `main` — every one of those files fails on `main` too, and this branch adds 8 passing tests and no new failure.

Four files have appeared as branch-only failures across runs. All four are pre-existing flakes on this machine, not regressions:

- `test/unit/status-writer.test.ts`, `test/unit/interactive-engine-generation-lifecycle.test.ts` -> pass in isolation, **2 files / 19 tests passed**.
- `test/suite/regressions/1223-startup-lazy-builtins.test.ts`, `test/suite/regressions/1704-lazy-tool-lifecycle-round2.test.ts` -> pass in isolation, **2 files / 17 tests passed**. These two do exercise tool registration, so I checked them specifically; under full-suite load they fail with `spawnSync bun ETIMEDOUT`, an environment timeout rather than an assertion.
- None of the four reference `registerTool`, `parameters:`, `createStructuredOutputTool`, `wrapToolDefinition`, or `Type.Union`, except the two lifecycle fixtures noted above, which pass when not competing for the machine.
- The failure set drifts run to run on identical code: the package suite reported 34, then 29, then 30 failures across three runs.

This is a Windows dev box, not CI; trust CI over these numbers for the pre-existing failures.

## Breaking changes

None. Object-rooted schemas — every built-in and every documented example — take an identity path and are byte-identical to today. Only union roots change, and they change from "uncallable" to "callable with unchanged validation semantics".

## Follow-ups

Two upstream defects in `@earendil-works/pi-ai` are described in #2189 and are **not** addressed here, because this fix makes them unreachable from Atomic:

1. `convertTools` could merge `anyOf` branch properties itself; note its `strict === true` path spreads the legacy schema last, clobbering `properties`/`required` back to empty.
2. `validateToolArguments` discards the return value of `Value.Convert`, which is the reason union roots get no coercion of their own.

<h3>Greptile Summary</h3>

This change projects supported object-root union tool schemas into Anthropic-compatible object input schemas while retaining the authored schema for local validation. Valid union projection paths were exercised through the Anthropic request flow. One malformed object-union branch remains unsafe: a non-record `properties` value is serialized as character-indexed fields instead of falling back to an empty object schema.

Not safe to merge until malformed object-union property metadata is rejected from Anthropic schema projection.

The real request path correctly handles supported object unions, but a malformed `properties` value is accepted and serialized into unintended provider-facing fields.

**Files Needing Attention:** packages/ai/src/api/anthropic-messages.ts

**What T-Rex did**
- T-Rex produced a proof for the posted P1 finding and linked it to the review comment detailing the finding.
- The P1 finding involved comparing the regression harness outputs before and after the PR, confirming the malformed schema both before and after the change.
- A separate P1 finding proof was generated to corroborate the initial result.
- General contract validation showed that typeboxCompile was accepted and runtimeException was null across captures, but also identified a mismatched safe fallback and a bug in how branch.properties is handled in anthropic-messages.ts at the indicated lines.

<a href="https://app.greptile.com/trex/runs/23021179/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

<details><summary><h3>Comments Outside Diff (3)</h3></summary>

1. `packages/ai/src/api/anthropic-messages.ts`, line 1771-1782 ([link](https://github.com/bastani-inc/atomic/blob/ad118bd28d99251222eb725d97cceb798e3e1ce6/packages/ai/src/api/anthropic-messages.ts#L1771-L1782)) 

   <a href="#">P1</a> **Strict union projection unavailable**

   Object-root unions are now projected for advertising, but callers requesting `{ type: "json_schema", strict: "require" }` still receive a local unsupported-union error and no request is sent. With `strict: "prefer"`, the same schema is advertised as non-strict instead. As a result, consumers cannot use the new object-union projection when their tool contract requires Anthropic constrained sampling. Either implement a strict-compatible projection or explicitly keep this path unsupported rather than presenting the projected schema as a general solution.

   <details><summary><strong>Artifacts</strong></summary><br />

   **[Runtime probe source for root-union strict-tool conversion](https://app.greptile.com/trex/artifacts/2e77702a-2cd6-447e-a43e-7f231b49dd1a)**

   - Authored Bun probe starts a local HTTP/SSE endpoint and invokes the Anthropic adapter with strict-prefer and strict-require root object unions, showing the exact outgoing payload and terminal events.

   **[Baseline root-union endpoint capture before projection](https://app.greptile.com/trex/artifacts/287aee03-d234-4497-b632-dd1baadb79c8)**

   - Executed against revision 615cfcf3; the HTTP endpoint returned 200 and captured an empty legacy object schema, while strict-require made no additional request and returned the local unsupported-union error.

   **[Current root-union endpoint capture after projection](https://app.greptile.com/trex/artifacts/d60e4dfe-032d-46cf-ba9b-11859397a648)**

   - Executed against revision ad118bd2; the HTTP endpoint returned 200 and captured projected object fields without strict mode, while strict-require made no additional request and returned the local unsupported-union error.

   **[PR root-union regression test result](https://app.greptile.com/trex/artifacts/c65153dc-cc23-471b-b0b1-07551b8268c2)**

   - Runs the PR's Anthropic object-union tool-schema test with the package Vitest configuration and records the executed result.

   **[PR whitespace validation result](https://app.greptile.com/trex/artifacts/8a16bb7d-305f-48c3-bbc5-0378a4fea4b8)**

   - Runs git diff --check for the reviewed PR range and records that the patch has no whitespace errors.

   <a href="https://app.greptile.com/trex/runs/22476406/artifacts?artifact=2e77702a-2cd6-447e-a43e-7f231b49dd1a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewArtifacts.svg?v=4">View artifacts</picture></a>
   </details>

   <details><summary>Prompt To Fix With AI</summary>

   `````markdown
   This is a comment left during a code review.
   Path: packages/ai/src/api/anthropic-messages.ts
   Line: 1771-1782

   Comment:
   **Strict union projection unavailable**

   Object-root unions are now projected for advertising, but callers requesting `{ type: "json_schema", strict: "require" }` still receive a local unsupported-union error and no request is sent. With `strict: "prefer"`, the same schema is advertised as non-strict instead. As a result, consumers cannot use the new object-union projection when their tool contract requires Anthropic constrained sampling. Either implement a strict-compatible projection or explicitly keep this path unsupported rather than presenting the projected schema as a general solution.

   ---

   `````
   </details>

2. General comment 

   <a href="#">P1</a> **Projected root object-union tool schemas cannot satisfy requested strict sampling**

   - **Bug**
     - A root object `anyOf` schema with `constrainedSampling: { type: "json_schema", strict: "require" }` does not reach Anthropic: the adapter returns an error and makes zero outbound requests. With `strict: "prefer"`, it sends a projected non-strict schema instead. Thus callers cannot use the new projection for the strict-tools contract.
   - **Cause**
     - `makeStrictJsonSchema` rejects `anyOf` variants whose members are object or array schemas (`constrained-sampling.ts:63-72`), and `resolveJsonSchemaStrictSampling` suppresses that error for `prefer` or surfaces it for `require` (`208-220`). Moreover, should strict root-union support be enabled, the projection branch at `anthropic-messages.ts:1771-1776` discards strict schema constraints including `additionalProperties: false` while retaining `strict: true` at line 1782.
   - **Fix**
     - Either explicitly keep root object-union schemas unsupported for strict mode and document/test that limitation, or implement a contract-preserving strict projection: emit `additionalProperties: false` for the projected root and every nested object, make all projected property names required with nullable representations as Anthropic requires, and validate the result against Anthropic's strict JSON Schema subset before setting `strict: true`.

3. General comment 

   <a href="#">P1</a> **Malformed object-union properties are projected into invalid Anthropic tool fields**

   - **Bug**
     - When an `anyOf` object branch has `properties: "abc"`, the current PR accepts it as projectable and sends an Anthropic tool input schema with bogus properties `"0": "a"`, `"1": "b"`, and `"2": "c"`, rather than using the unsupported-root safe fallback. The request does not throw locally, so malformed schema metadata reaches the provider request.
   - **Cause**
     - `projectObjectUnionForAnthropic` checks only `branch.properties !== undefined` at line 1726. It does not establish that `properties` is a non-null, non-array record before `Object.entries` consumes it at line 1733. Strings are enumerable in JavaScript and therefore become character-indexed entries.
   - **Fix**
     - In `packages/ai/src/api/anthropic-messages.ts`, require each branch `properties` to be a non-null object and not an array (for example, a narrow schema-record guard) before projecting. If any branch fails this guard, return `undefined` so `convertTools` emits its existing `{ type: "object", properties: {}, required: [] }` fallback.

</details>

`````markdown
### Issue 1
packages/ai/src/api/anthropic-messages.ts:1726
**Malformed properties are projected as fields**

Checking only that `branch.properties` is defined admits non-record values such as `"abc"`. `Object.entries()` then interprets that string as enumerable character entries and advertises fields `"0"`, `"1"`, and `"2"` to Anthropic rather than using the existing empty-object fallback. Require a non-null, non-array record before projecting the union; otherwise return `undefined` so unsupported metadata follows the safe fallback.

---

`````

</details>

<sub>Reviews (5): Last reviewed commit: ["fix(ai): preserve Anthropic object-union..."](https://github.com/bastani-inc/atomic/commit/486654cc021cbe531ecd2ca7e380dddd997b2545) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50138284)</sub>


elefthei:
Updated: I found a working registry mirror, so the suite now actually runs and the PR body carries real output instead of an explanation of why it was missing.

Two changes since the first push:
- `npm run check` caught an import-ordering violation in the new file (`import { Type, type TSchema }` -> `import { type TSchema, Type }`). Fixed and folded into the commit, so the branch was force-updated.
- Full results are in the body: 8/8 new tests, `npm run check` clean, and a branch-vs-`main` comparison of both suites showing no new failures attributable to this change.

elefthei:
Rebased onto latest `main` (was conflicting after 0.9.12 shipped). The changelog entry moved from what is now the released `[0.9.12]` section into `[Unreleased]` -> `### Fixed`; the released section is untouched, and the changelog diff against `main` is a pure +4/-0 addition.

Re-verified on the rebased base after a fresh `npm ci`:
- `npm run check` -> clean (biome, `tsc --noEmit`, shrinkwrap)
- new test file -> 8/8 passed

Body updated with the final regression numbers.

flora131:
Hi @elefthei, thanks for your patience, and sorry for the delay. We have been working through a larger refactor and the pi 0.84.1 dependency upgrade.

Since your last update, `main` has moved and the branch now conflicts. Could you merge or rebase the latest `main` and obtain fresh CI?

Could you also address the concern in the latest Copilot review that open union branches may become stricter after normalization, either with focused coverage or an explanation of why the accepted input set remains unchanged?

Please let us know if you have any questions.

flora131:
Thanks for the careful diagnosis and implementation here. We reproduced #2189 on current `main` at `9c43ecc575ada84728f11875dba1e49441d54378`.

The latest provider contracts and current Atomic adapters narrow the defect to the Anthropic Messages conversion path: it rebuilds `input_schema` from root `properties` and `required`, dropping a root `anyOf`. OpenAI, Gemini, and Bedrock already preserve the authored union. Anthropic's current tool schema contract supports `anyOf` within an object schema.

We are going to update this PR from current `main` with a smaller provider-local fix. The authored schema will remain unchanged for runtime validation, while the Anthropic request keeps the union and adds the required object root. This avoids changing validation or coercion for other providers. We will preserve your contribution and attribution, verify it through a real Anthropic tool call using a focused test extension, and post the exact evidence here.

flora131:
Updated this PR to `ad118bd28d99251222eb725d97cceb798e3e1ce6`, rebased on current `main` at `9c43ecc575ada84728f11875dba1e49441d54378`. Lef's original contribution remains as its own authored commit (`615cfcf3b`).

One correction to the earlier note: the live Anthropic API rejects `anyOf`, `oneOf`, and `allOf` at the top level of a tool's `input_schema`. The final fix therefore projects object-union branch fields only in the Anthropic Messages adapter, intersects required fields, and keeps property variants as nested unions. Atomic retains the untouched authored union for runtime validation. Other provider adapters and shared validation/coercion are unchanged.

Verification:

- `npx --no-install vitest --run --config packages/ai/vitest.config.ts packages/ai/test/anthropic-object-union-tool-schema.test.ts` — 1 test passed.
- A real direct `anthropic/claude-haiku-4-5` run through a temporary Atomic extension called `store_container` with `{"requestId":"live-check","kind":"items","values":["alpha","beta","gamma"]}`. The call passed runtime validation, reached `execute`, and returned `isError: false`.
- An independent live check against the final head used a natural-language prompt that named no schema fields; Claude still produced the correct container-valued branch. A second live check exercised the nested-object branch successfully.
- Local validation continues to reject invalid cross-branch mixtures against the original union.

Thanks again to @elefthei for the original diagnosis and implementation. The final branch preserves that attribution while narrowing the workaround to the provider that needs it. Normal PR CI can now provide the merge gate.

PR #2281, opened 2026-08-10
feat(tui): add fullscreen input navigation

lavaman131:
Port fullscreen transcript navigation, prompt history precedence, and Windows right-click paste while preserving the isolated interactive input route.

Workflow collisions tested in both TUI modes: raw pageUp/pageDown stage-chat input, tui.select.pageUp/pageDown prompt-card movement, and fullscreen half-page/home/end bindings while overlays own focus.

Assistant-model: GPT-5.6 Luna

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

---
<a href="https://app.blacksmith.sh/bastani-inc/codesmith/atomic/pr/2281"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788926864&installation_model_id=10562&pr_number=2281&repository=bastani-inc%2Fatomic&return_to=https%3A%2F%2Fgithub.com%2Fbastani-inc%2Fatomic%2Fpull%2F2281&signature=2448c4c7a77e17e74d7125589f20c55a0ac87264f032703405f7fab5818bc91c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

<h3>Greptile Summary</h3>

This PR adds fullscreen transcript navigation, focused-component input precedence, configurable prompt-history shortcuts, and right-click clipboard paste in fullscreen mode. One test-harness maintainability issue remains: the right-click paste test bypasses TypeScript’s structural checks when invoking a private method.

The product behavior changes are covered by focused tests; the remaining issue is limited to the type safety and refactor resilience of a test helper.

There is one independent non-security P2 finding. Under the scoring rules, a nonempty set containing only P2 findings receives a score of 4.

**Files Needing Attention:** packages/coding-agent/test/interactive-right-click-paste.test.ts

`````markdown
### Issue 1
packages/coding-agent/test/interactive-right-click-paste.test.ts:20-22
**Avoid the unchecked prototype cast**

Casting `InteractiveMode.prototype` through `unknown` bypasses compile-time validation between the fabricated test context and the private implementation method, making this test brittle during refactoring. Use a specifically typed test seam instead of the ambiguous cast.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(tui): add fullscreen input navigati..."](https://github.com/bastani-inc/atomic/commit/c37880aa05882dbd65b213116b2dcf793bb2799c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=52094625)</sub>

**Context used:**

- Context used - CLAUDE.md ([source](https://github.com/bastani-inc/atomic/blob/main/CLAUDE.md))


lavaman131:
## Reviewer findings carried forward

Passed all gates; two items are being handled in the next layer (`fullscreen-polish`) rather than left open.

**1. Focused-component pass-through is broader than the docs claim (behavior change).**
`docs/keybindings.md:99` says a focused component's *matching* `pageUp`/`pageDown`/`home`/`end` bindings take precedence. The implementation passes through unconditionally for **any** focused non-editor component that has `handleInput`, whether or not it binds the key. So a focused overlay that ignores `home` now swallows it instead of scrolling the transcript — a change from the pre-layer state where `TuiAltScreen` consumed those keys globally.

This matters more here than upstream: Atomic ships overlays in `packages/workflows/src/tui/` (`inline-form-overlay.ts`, `store-widget-installer.ts`) that upstream has no equivalent of. Being fixed in `fullscreen-polish` by making the gate consult the focused component's return value, rather than by narrowing the sentence.

**2. `interactive-tui.ts:24-49` is coupled to pi-tui constructor ordering by convention.**
The WeakSet assumes the first `addInputListener` call on the instance is the viewport listener. True at 0.84.1 (`tui-alt-screen.js:77`), but a future pi-tui registering any listener earlier would silently gate the wrong one and hand the workflow paging collision back. The `preserves workflow paging precedence` test would catch it; a comment naming the load-bearing assumption is being added.

**3. Recorded for the next reviewer:** multi-click selection and the tmux/Zellij/Screen mouse-tracking reduction were verified present in pi-tui 0.84.1 and deliberately **not** ported. The commit message omitted that confirmation.

flora131:
## Decision on the pass-through defect: add a reply channel

Repository owner has chosen **option (a): add a reply channel to `engine_custom_input`**, rather than declaring that remote components always consume, or deferring.

### Why this was escalated

The partial fix attempted in `fullscreen-polish` could not work, and the reviewer diagnosed exactly why:

- `RemoteComponent.handleInput` (`interactive-engine/remote-component.ts:64-74`) returns `true` unconditionally because it **cannot** do otherwise — `sendEngineCommand({type:"engine_custom_input"})` is fire-and-forget, `engine-custom-ui.ts:280` calls `record.component.handleInput?.(command.data)` and discards the result, and `protocol.ts:107` carries no reply for that command.
- So under `IsolatedInteractiveRuntime` — Atomic's shipped interactive path — a focused remote component that ignores `home` still reports handled, and the transcript does not scroll. The fix landed only on the in-process branch.
- A partial fix also **introduced** a regression: `tui.select.pageUp` and `tui.altScreen.pageUp` share the `pageUp` default (pi-tui `keybindings.js:80-83,91-99`), so three selector lists both paged their selection and scrolled the transcript.

That work has been reverted out of `fullscreen-polish`, which now ships only its seven genuine polish fixes plus `HANDOFF-pass-through-gate.md`.

### What lands next

A dedicated layer on top of `fullscreen-polish`:

1. A reply channel on `engine_custom_input` so the child's `handleInput` return value reaches the parent.
2. `RemoteComponent.handleInput` reports what the child actually did.
3. The ~16 shipped host components declaring `handleInput(): void` converted to report handling, so the gate's "returned non-`true` means not handled" rule is sound rather than accidentally true.
4. The settling test the reviewer specified: mount a `RemoteComponent`, focus it, feed `\x1bOH`, assert `transcript.scrollTop === 0` with a no-op child `handleInput`.
5. `docs/keybindings.md:99` updated to describe the behaviour only once it is actually true.

This touches the isolated engine's protocol, which is Atomic's most load-bearing divergence — `settleActiveResponseBeforeTeardown` must remain a no-op and no teardown path may gain `await this.session.abort()`. It gets its own reviewable diff for that reason.

flora131:
Content merged to `main` via #2300, which squashed all 14 commits of the fullscreen stack (core → dock → scrollbar → input → polish → pass-through). Closing the layer PR: its diff is in main, but the squash means git cannot auto-close it.

Verified present in main, e.g. `src/modes/interactive/interactive-tui.ts` and the `scrollbarThumb` schema keys.

PR #2384, opened 2026-08-14
feat: durable remote session adapters and session telemetry

jkyamog:
This branch adds the Atomic-side support needed for extension-provided remote
session adapters, safe workflow continuation, and useful telemetry for stages
that do not expose a local `AgentSession`.

The branch is based on `origin/main` and contains three logical commits:

1. `feat(workflows): add named session adapters`
2. `fix(workflows): harden remote session continuation and resume`
3. `feat(ui): show adapter session telemetry`

## What changed

- Added a versioned named-session-adapter registry with extension discovery,
  conflict detection, and late-registration support.
- Added serializable stage adapter selection and an explicit adapter retry
  policy, including `retryPolicy: "never"` for adapters that require explicit
  workflow control for reconnects.
- Hardened durable stage continuation and workflow resume:
  - preserves stable remote session identity;
  - recovers an accepted but unfinished turn with `resume_unfinished_turn`
    instead of submitting the prompt twice;
  - recovers a completed remote tail without another model turn;
  - settles queued steering/follow-up messages without creating synthetic
    duplicate continuation work;
  - keeps replay/postmortem stage chat aligned with durable session state.
- Added adapter-backed session statistics to the interactive usage footer,
  including context usage, input/output tokens, cache read/write totals, cache
  hit rate, and cost when supplied by the adapter.
- Refreshes the usage ribbon for adapter `session_stats` events without routing
  those telemetry-only events through the ordinary transcript reducer.
- Added documentation and regression coverage for adapter registration,
  continuation, resume, queued steering, replay, and telemetry behavior.

---
<a href="https://app.blacksmith.sh/bastani-inc/codesmith/atomic/pr/2384"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789322860&installation_model_id=10562&pr_number=2384&repository=bastani-inc%2Fatomic&return_to=https%3A%2F%2Fgithub.com%2Fbastani-inc%2Fatomic%2Fpull%2F2384&signature=7fd1f5c3a0a50b7d5bb11c760be542d06431e2e35fe596b4f9c4acb4d54afaab"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

<h3>Greptile Summary</h3>

This change restores durable workflows for read-only inspection and adds follow-up chat for retained conversations.

A completed workflow can still register a follow-up chat using a named adapter that is no longer available. The chat appears available but fails only when it attempts to restore the retained session.

### T-Rex validation blocked
The focused check could not run because the required workspace package `@bastani/pi-ai/compat` could not be resolved. Bun stopped during module loading before the completed-workflow restore and attachment path could execute.

A follow-up chat for a completed workflow can fail at first use when its persisted named adapter is unavailable.

One actionable non-security reliability issue remains in the completed-workflow chat registration path.

**Files Needing Attention:** packages/workflows/src/durable/completed-inspection.ts

**What T-Rex did**
- The focused command bun trex-artifacts/completed-inspection-missing-named-adapter-02-after.ts was executed from the repository, but Bun exited with code 1 because packages/coding-agent/src/core/typebox-compat.ts could not resolve the workspace package @bastani/pi-ai/compat, so the completed workflow could not be opened.
- The failure was validated as a Cannot find module '@bastani/pi-ai/compat' error in the execution record, indicating an environment/workspace dependency-resolution blocker.

<a href="https://app.greptile.com/trex/runs/20947707/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

`````markdown
### Issue 1
packages/workflows/src/durable/completed-inspection.ts:132
**Adapter validation remains bypassed**

When completed durable inspection opens a run using a named-only adapter, the default `agentSession` check returns before registering follow-up chat. When the persisted named adapter is unavailable, this direct handle construction bypasses `findSessionAdapter` and exposes an apparently usable handle that throws during attachment instead of leaving the transcript read-only.

---

`````

</details>

<sub>Reviews (5): Last reviewed commit: ["fix(workflows): honor per-stage sessionA..."](https://github.com/bastani-inc/atomic/commit/821013033da81014003a01ee038283236afab9d6) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53417567)</sub>


lavaman131:
## Make the run, not the stage, the unit of remote placement

Thanks for the thought put into this. The continuation and resume hardening is the hard part of the problem, and most of it survives what I am about to propose. But `sessionAdapter` is a per-stage option, and I think placement belongs at the run.

Why the stage is the wrong place to split:

- **Tool gates check the wrong tree.** `ctx.tool` callbacks run in the host process against the local filesystem. A remote stage writes somewhere else, so the build gate after it passes on unchanged source and nothing reports that.
- **Artifacts and `reads` are host paths.** `workflowArtifactRunPath()` resolves under the local agent dir, so "Read the file at `<path>`" fails on another node. Copying artifacts across nodes means writing a distributed filesystem.
- **Two parallel stages can write one file.** Split across nodes there is no lock, no last writer, and no conflict to report — just two trees, and the next tool node reads one of them. Worktrees are local paths.
- **Steering, prompts, and intercom run in process today.** Each becomes a network protocol with its own ordering rules, and stages in one group stop reaching each other.
- **Recovery repeats per stage.** This PR already needs `resume_unfinished_turn`, `retryPolicy: "never"`, and "never resend on an uncertain transcript tail". That is exactly-once delivery over a lossy link. Per-stage placement solves it once per stage, with the checkpoint on one node and the session on another.
- **A dead node leaves a run neither failed nor resumable**, which breaks the root failure boundary we promise authors.
- **The selector crosses ~18 durable, persistence, replay, and TUI files.** At run level it is one field, read once at launch.

A run already owns one durable record, one artifact root, one intercom group, one worktree contract. Move a whole run and all of that still holds. Throughput does not suffer: run independent workflows at once, which is how we dispatch them anyway.

### Where this is going

I am looking at Coder, and at GitHub, so we can back several cloud providers. The model I want is one VM per workflow run, with the durable store on a hosted PostgreSQL so any VM can resume any run.

One VM per run matches the run boundary exactly. A VM has one filesystem, one clone, one worktree, one set of local paths, which is what every stage and tool node in a run already assumes. Stage-level placement leaves two choices and both are bad: a VM per stage, which costs more and brings back every problem above, or all stages in one VM, which is run-level placement with extra plumbing.

The hosted database needs little from us. The backend is already DBOS on Postgres, `effectiveSystemDatabaseUrl()` already reads `DBOS_SYSTEM_DATABASE_URL`, and the embedded local Postgres becomes one choice among several. Point a fresh VM at the shared database and it can claim a run and replay it.

Note what that does not solve. DBOS keys durable state by workflow, and stage and tool checkpoints hang off a workflow ID, so a shared database hands us run-level resumption and nothing finer. It cannot give stage-level resumption, because what a remote stage needs back is the live session and its transcript, and that is not in Postgres. So the shared database makes the run the recoverable unit, and per-stage placement then splits recovery between a database one node owns and a session another node owns.

Placing at the run also keeps the provider out of the workflow definition. `sessionAdapter` on a stage writes a provider choice into the authoring contract, where it turns wrong as soon as there is a second provider. At the run it is a launch input, and Coder, GitHub, or a local host is one more backend behind it.

### Keep and change

Keep: `resumeUnfinishedTurn` and its RPC command, `success: false` on a rejected continuation, and the no-double-submit path. A remote run needs all of it.

Change: drop `StageOptions.sessionAdapter` and its threading, keep `AgentSessionAdapter` internal, and select at launch. The registry shape is fine — it needs a run-level caller.

Tell me if there is a plan for the filesystem problem that I have missed. Otherwise let us talk through the rework before you spend time on it, since the part I want cut is separable from the part I want kept.

jkyamog:
I am pivoting and changing how I approach this.  So moved my local AI to the same gateway as cloud AI, I do still need the work to be done on remote hosts.  Just like normal pi --ssh I needed something that atomic can spin out agents with remote access.  The process is still on the same host but execution of tools are on the remote host.  https://github.com/jkyamog/atomic-remote-ssh

PR #2407, opened 2026-08-15
fix(compaction): use symmetric heuristic stats pair and authoritative display

### LINKED ISSUE #2052: bug: compaction stats mix authoritative and heuristic token estimates, producing negative percentReduction
## What happened?

A compaction can report that it made the context **larger**. Observed in a real session's JSONL:

```json
{"rung": "planned", "strategy": "verbatim-lines",
 "stats": {"linesBefore": 288, "linesDeleted": 144, "linesKept": 144,
           "rangeCount": 89, "tokensBefore": 3126, "tokensAfter": 9869,
           "percentReduction": -215.7}}
```

144 of 288 lines were deleted, yet `tokensAfter` is triple `tokensBefore` and `percentReduction` is −215.7%.

The deletions themselves are correct. Only the reported numbers are wrong.

## Steps to reproduce

1. Run a session through several compactions until the compactable region is small relative to the `preserve_recent` tail (repeated `/compact` on a short session does it).
2. Inspect the resulting `compaction` entries in the session JSONL.
3. `details.stats.tokensAfter` exceeds `tokensBefore` and `percentReduction` is negative.

Observed on `anthropic` / `claude-opus-5`.

## Expected behavior

`tokensBefore` and `tokensAfter` are measured the same way, so `percentReduction` describes a real change and is never negative for a compaction that deleted lines.

## Additional context

**Root cause: three incommensurable token estimators feed one subtraction.**

```ts
// compaction-boundary.ts:136 — provider-AUTHORITATIVE when usage is available
tokensBefore: estimateContextTokens(buildSessionContext(entries).messages).tokens

// compaction-boundary.ts:140 — crude per-message heuristic
keptTailTokensByPreparation.set(preparation,
  tailMessages.reduce((sum, item) => sum + estimateTokens(item.message), 0));

// transcript-serialization.ts:155 — char/4 on serialized text
tokenEstimate: Math.ceil(text.length / 4)
```

and they are combined in `compaction-runner.ts:91-94`:

```ts
const tokensAfter = result.stats.tokensAfter + (keptTail ? getKeptTailTokenEstimate(preparation) : 0);
const percentReduction = preparation.tokensBefore === 0
  ? 0
  : Math.round((1 - tokensAfter / preparation.tokensBefore) * 1000) / 10;
```

The asymmetry is in `estimateContextTokens` (`compaction.ts:112-130`): when the last assistant message carries provider usage, it returns `calculateContextTokens(usage, api)` — a **real tokenizer count from the API** — and only falls back to summing `estimateTokens` when no usage exists.

So `tokensBefore` is authoritative while `tokensAfter` is entirely heuristic (`char/4` of the compacted text plus a per-message estimate of the tail). Subtracting one from the other is apples-to-oranges. When the region is small and the kept tail is proportionally large, the heuristic sum overshoots the real count and the reduction goes negative.

`scrubPreCompactionAssistantUsage` compounds it: post-compaction the stale usage is deliberately scrubbed, so the "after" side has no authoritative number available even in principle.

**Severity: display only.** `stats.tokensAfter` appears exactly once in the source — at `compaction-runner.ts:95`, where it is computed. Nothing reads it. It never feeds `shouldCompact`, the auto-compaction threshold, the post-tool hard-limit gate, or the rung ladder. The impact is a confusing footer line and a nonsense number in the durable entry.

Same family as [earendil-works/pi#6340](https://github.com/earendil-works/pi/issues/6340) and [earendil-works/pi#6464](https://github.com/earendil-works/pi/issues/6464) — stale/mismatched usage accounting around compaction.

### Suggested direction

Use one estimator for both sides of the reported comparison: compute the stats-facing `tokensBefore` with the same heuristic that produces `tokensAfter`, and keep the provider-authoritative number for the **threshold decision**, where it belongs and is correct.

Clamping `percentReduction` at 0 would hide the mismatch rather than fix it, and would leave `tokensAfter` overstated. Clamp only after the estimators agree, if at all.

Found while capturing E2E evidence for #2050; predates that change and is untouched by it.

yangjj-iso:
## Summary

Closes #2052.

The compaction stats mixed three incommensurable token estimators (provider-authoritative, per-message heuristic, and char/4 on serialized text), producing negative `percentReduction` values when the region was small relative to the preserved tail.

Implements **option 3** from the issue discussion, as approved by @flora131.

### 1. Centralized widening/ratio logic (`computeWholeContextStats`)

Extracted from the old `withWholeContextStats` into an exported function in `compaction-runner.ts`, used by planned, fresh, and extension paths. Both sides of the reported comparison now use the **same heuristic**: region char/4 + explicit tail estimate.

- `tokensBefore = region.tokenEstimate + tailEstimate` (when tail kept)
- `tokensAfter = regionStats.tokensAfter + tailEstimate` (when tail kept)
- When the fresh rung drops the tail, the tail term correctly does not appear on either side.

### 2. Removed authoritative-minus-heuristic tail fallback

`getKeptTailTokenEstimate` in `compaction-boundary.ts` now defaults to `0` instead of `preparation.tokensBefore - preparation.region.tokenEstimate` (authoritative minus heuristic). The old fallback mixed units and could overshoot the real count.

Exported `setKeptTailTokenEstimate` so test fixtures can set the tail estimate on preparations constructed directly (not through `prepareCompactionBoundary`).

### 3. Authoritative display through adapter

- Added `VerbatimCompactionDetails.tokensBefore` as an optional field carrying the authoritative count through to the TUI display.
- `CompactionBoundaryMessageComponent` uses `displayTokensBefore` (from `entry.tokensBefore` / `result.tokensBefore`) instead of `stats.tokensBefore` for the "Compacted from N tokens" line, projected through the `BoundaryView` adapter.
- Legacy entries without `details.tokensBefore` fall back to `stats.tokensBefore`.
- The authoritative `tokensBefore` is preserved on `CompactionEntry` and `VerbatimCompactionResult` for budgeting and threshold decisions.

### 4. No clamping

The percentage is **not** clamped: a genuine expansion may still be negative.

## Regression tests

Added `test/unit/compaction-stats-symmetry.test.ts` (11 tests) covering:
- Large kept tails in planned and extension paths
- Fresh compaction with tail kept and dropped
- `preserve_recent: 0`
- Live and resumed rendering where display count differs from heuristic stats count
- Legacy entries without `details.tokensBefore` fallback
- `computeWholeContextStats` symmetric ratio properties

Updated existing tests in `compaction-fresh-window.test.ts` to use `setKeptTailTokenEstimate` instead of the removed mixed-unit fallback.

## Test results

All 141 affected tests pass. TypeScript type check and Biome lint/format all pass.

<h3>Greptile Summary</h3>

This change makes compaction reduction statistics compare like-for-like estimates while preserving the provider-authoritative token count in boundary messages. Focused unit and integration coverage confirms that dropping a protected tail includes that tail in the original-context baseline, and that live and resumed views display the authoritative original token count.

No blocking failure remains; the compaction and boundary-rendering paths exercised by the focused checks behave as intended.

No accepted blocking findings remain after exercising fresh compaction with a dropped tail and boundary rendering with distinct heuristic and authoritative counts.

**What T-Rex did**
- Ran the focused 11-test compaction regression unit suite and the focused four-test fresh-compaction integration suite on the current checkout; both suites passed.
- Compared the current test outcomes with the parent checkout and observed that the unit and integration runs meet the expected behavior for the targeted dropped-tail scenarios, with all targeted tests passing.
- Documented the validation coverage showing the computeWholeContextStats and related boundary display paths align with the intended token counts and tail handling as described in the test-driven validation.

<a href="https://app.greptile.com/trex/runs/19097306/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

<sub>Reviews (4): Last reviewed commit: ["Merge remote-tracking branch &#39;origin/mai..."](https://github.com/bastani-inc/atomic/commit/accc98b7ec03f25632b33044bb3a5113abec8afd) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53557062)</sub>


flora131:
Thanks for the work on this. It looks like #2449 is the newer, focused version of the same #2052 change, and we’re reviewing that PR now.

Would you be okay with us closing #2407 in favor of #2449? That would keep the review queue clear and avoid us reviewing or commenting on the older implementation by mistake. If #2407 contains any behavior that you intentionally did not carry into #2449, please let us know so we can account for it before closing.

flora131:
@yangjj-iso, checking in on #2407. Your August 20 update selected this as the canonical branch for #2052 and mentioned a rebase, protected-tail and documentation work, and fresh checks. The assignment has already been inactive beyond the normal seven-day window in our `CONTRIBUTING.md` guidelines.

Please let us know by September 4 whether you intend to continue. If we don't hear from you by then, we may release the #2052 assignment and close #2407 and #2449, while leaving the issue available for another contributor to pick up.

Thanks!

flora131:
@yangjj-iso, following up on our September 2 message: the September 4 response deadline has passed, and we haven’t heard back, so we’re closing this PR and releasing the #2052 assignment for now.

We appreciate the investigation and implementation work. We also recognize that our earlier uncertainty about which branch to continue made the process less straightforward.

We confirmed that #2052 still affects current `main`, so the issue will remain open with links to your work and the agreed approach. You’re welcome to return—please comment on the issue so we can coordinate ownership before work resumes.

PR #2412, opened 2026-08-15
refactor(workflows): one live review session per design

flora131:
## Summary

`open-claude-design` wrapped a bounded `generate-N` → `user-feedback-N` loop around an `impeccable live` session that is itself unbounded. This collapses it to one review session per generated design.

## Why

Inside a live session the user already picks elements, accepts variants — written into `preview.html` **in place** — and steers the page until they leave. So when a review stage returned, its work was already applied, and the following `generate-N+1` re-read that file and rewrote it whole, with `live_changes` threaded in mainly so the rewrite would not revert what the user accepted.

Two defects followed:

1. **A no-op refinement round.** The stage prompt required `revise` "whenever the user asked for ANYTHING at all", so a session where the user only accepted variants triggered a full regeneration whose real job was to avoid undoing live's work.
2. **`max_refinements` capped the wrong loop.** It bounded review rounds while the inner session had no bound, so the number users read as their iteration budget did not describe it.

## What changed

- The feedback schema's `decision` becomes `export` | `regenerate`, replacing `approve` | `revise`.
- `refineOpenClaudeDesign` runs one review session per design. `export` goes to the exporter; `regenerate` runs one more `generate-*` as a **fresh pass** from the brief, design context, references, and the user's notes, then opens the next session.
- `max_refinements` bounds regenerations. When the budget is spent, the run exports and states the dropped request in the design summary rather than discarding it silently.
- Deleted: the pending-feedback threading, the final-round revision special case, `assertUserAnnotationsThreaded`, and the `live_changes` preservation plumbing that existed only because a whole-file rewrite followed every session.
- The approve-carrying-work coercion is gone: notes the user still wants applied are exactly why they would choose `regenerate`, and an explicit `export` is their own decision about an artifact they were just editing.

## Breaking change

Anything reading `<artifact_dir>/feedback/iteration-N.json` must handle `export`/`regenerate` instead of `approve`/`revise`.

## Tests

- an `export` session runs no `generate-2` and reaches `exporter`
- a `regenerate` session runs exactly one `generate-2` carrying the notes, then reviews again
- the regeneration budget: a second `regenerate` at `max_refinements: 1` exports and records the exhausted budget
- the durable artifact is still written per session and its path still reaches the next generate stage's `reads`
- `test/unit/executor-structured-output-continuation.test.ts` passes unchanged — the #2401 regression proving a continuation cannot erase a finalized structured result

## Verification

[...]

Issue #2401, opened 2026-08-14
bug(workflows): late continuation can erase open-claude-design feedback and silently approve a stale preview


flora131:
## What happened?

`open-claude-design` can silently export a stale preview after a live-review stage captured real user feedback.

Observed event order:

1. `user-feedback-1` emitted a complete labeled report containing `user_notes`, `live_changes`, and `next_action_hint`.
2. 15 ms later, the stage admitted a synthetic `Continue where you left off...` turn.
3. The agent replied with a short unlabeled wrap-up.
4. The stage returned that later wrap-up as `WorkflowTaskResult.text`, replacing the labeled report.
5. `open-claude-design` parsed only the replacement text, found no meaningful feedback, set `approvedForExport = true`, skipped `generate-2`, and exported the stale preview.

The user then saw changes described during live review that were never applied. One directly accepted live edit persisted because it changed `preview.html` in place; the remaining requested changes existed only in the displaced report.

This is an interaction between a shared runtime behavior and the bundled workflow:

- **Runtime:** a late continuation reply can replace an earlier finalized stage deliverable.
- **Workflow:** `open-claude-design` consumes only `result.text` and interprets missing parsed feedback as approval.

The user-visible silent approval occurs in `open-claude-design`, but the result-displacement mechanism can affect any workflow that relies on plain `WorkflowTaskResult.text`.

### Focused code-path reproduction

On `@bastani/atomic@0.9.14-alpha.1`:

1. Create a feedback-stage transcript with this order:
   - assistant: a labeled report containing `user_notes` and `live_changes`
   - user: `RESUME_CONTINUATION_PROMPT` / `Continue where you left off...`
   - assistant: a short unlabeled completion message
2. Let stage-result selection choose the last assistant message.
3. Pass that text to `toPreviewFeedback(...)`.
4. Observe:
   - `userNotes === undefined`
   - `liveChanges === undefined`
   - `hasMeaningfulFeedback(...) === false`
5. Follow `refineOpenClaudeDesign(...)` and observe that the workflow sets `approvedForExport = true` rather than starting the next `generate-*` refinement.

A local direct-import/temp-script reproduction against the installed alpha passed 9/9 assertions. It used the actual intended report and displaced stage result from the observed run. This was a code-path reproduction, not a full browser E2E run.

### Observed live result

The live run exported with:

- `refinements_completed: 1`
- `approved_for_export: true`
- no `generate-2` stage

[...]

PR #2413, opened 2026-08-15
feat(workflows): let the workflow drive the impeccable live poll loop

flora131:
## Summary

Stacked on #2412. The review stage owned the `impeccable live` poll loop, which meant the model decided when a review was over. This moves the loop into the workflow, so termination is an event the helper emits rather than a judgment a stage makes.

## Why

The live contract says `"timeout" -> LOOP`. That is an instruction to a model, not a constraint the runtime enforces. A stage that read a poll timeout as "the user is done" would end its turn, the runtime would force a schema-valid structured answer out of it, and the refinement loop would route on a review the user was still in.

In a chat session that ambiguity costs nothing — the user just keeps talking. A workflow stage boundary is a commit point, so the same ambiguity silently discards a review round.

This is the same failure class as #2401. That fix stopped a late turn from *replacing* a finalized answer; it did nothing about an answer produced too early.

## What changed

When a project has the impeccable scripts installed:

| Node | Kind | Role |
|---|---|---|
| `user-feedback-N-start` | stage | boots `live.mjs`, opens the page, prints the review URL, then stops |
| `live-poll-N-M` | tool | polls the helper; absorbs `timeout` internally |
| `live-generate-*`, `live-steer-*`, `live-manual_edit_apply-*` | stage | handle exactly the events that need a model |
| `live-reply-N-M` | tool | acknowledges each model-handled event (`done` / `steer_done`) |
| `user-feedback-N` | stage | reports the finished session as the structured deliverable |

- `accept`, `discard`, and `prefetch` are acknowledged by the poll script itself and mint no stage.
- `timeout` is absorbed inside the poll node, so an idle hour costs no graph nodes and never looks like an ending.
- Unreadable poll output is treated as a timeout, not an ending.
- The loop ends only on the helper's own `exit` event.

## Fallback

A project without the impeccable scripts, or a host without `ctx.tool`, falls back to the single model-driven review stage unchanged.

## Tests

New `test/unit/open-claude-design-live-protocol.test.ts` (9): event parsing, unreadable output as timeout, timeout absorption, missing scripts reporting `exit`, which events call the model back, reply tokens, and the two prompts.

New workflow-level tests: the loop polls, mints a stage per model-facing event, replies to exactly those, ends on `exit`, and reports the deliverable; and a project without the scripts takes the fallback path.

`test/helpers` change: the mock ctx accepts `cwd` so a test can point a builtin at a temp project.

## Verification

`npm run check`, `npm run test:unit` (6320), `npm run test:integration` (486), `npm run test:ci-contracts` (42) all green.

Refs #2401

---
<a href="https://app.blacksmith.sh/bastani-inc/codesmith/atomic/pr/2413"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789368904&installation_model_id=10562&pr_number=2413&repository=bastani-inc%2Fatomic&return_to=https%3A%2F%2Fgithub.com%2Fbastani-inc%2Fatomic%2Fpull%2F2413&signature=aaed427ecb9077d880dc10e17da339f54c6aef4b6b5307f0c12d214a3168931f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

<h3>Greptile Summary</h3>

The live-review workflow cannot reliably complete model-handled interactions: acknowledgement commands omit fields required by the helper and rejected replies are ignored. In addition, poll-helper failures are retried as idle timeouts instead of ending the run with an actionable error. These two reliability issues should be corrected before merge.

<h3>Confidence Score: 3/5</h3>

Not merge-safe until live acknowledgements and poll-helper error handling are corrected.

Two distinct non-security P1 failures remain: model-handled live events cannot be acknowledged through the current adapter, and failed poll-helper executions retry indefinitely. Under the scoring table, two non-security P1 findings produce a confidence score of 3.

**Files Needing Attention:** packages/workflows/builtin/open-claude-design-live-protocol.ts; packages/workflows/builtin/open-claude-design-phases.ts

**What T-Rex did**
- T-Rex produced a proof for a posted P1 finding and linked it to the corresponding review comment.
- To validate the P1 finding, T-Rex inspected the successful complete live-helper reply contract and opened the focused live reply contract repro source.
- A second P1 finding proof was produced, including a controlled nonzero helper repro script.
- The second proof also captured the repro source, the observed poll output, and the current poll implementation lines to support validation.

<a href="https://app.greptile.com/trex/runs/19017389/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

`````markdown
### Issue 1
packages/workflows/builtin/open-claude-design-live-protocol.ts:184
**Live reply adapter omits required acknowledgement fields**

`replyLiveEvent` only invokes the helper as `--reply <token>`, although the live helper requires `--reply <event-id> <status>`. The workflow also has no way to send the required structured `--data` result for `manual_edit_apply`. Consequently, generate, steer, and manual-edit acknowledgements are rejected before they reach the helper endpoint, leaving the corresponding live events pending. Pass the event ID and any manual-edit result through the reply API, and treat a nonzero helper result as a failed workflow operation.

### Issue 2
packages/workflows/builtin/open-claude-design-live-protocol.ts:169-170
**Failed live polls are retried as timeouts**

`pollLiveEvent` parses stdout without checking `ScriptResult.code`. A failed `live-poll.mjs` invocation with empty output is parsed as a timeout, so the loop immediately launches another poll rather than surfacing the unavailable helper, authentication failure, or other terminal error. Check the exit status before parsing stdout and throw an error that includes the helper status and stderr.

---

`````

</details>

<sub>Reviews (3): Last reviewed commit: ["docs(workflows): say how a live review e..."](https://github.com/bastani-inc/atomic/commit/a16b8f5235b93b47f5560f5e88adc1f355fd759d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53581707)</sub>


flora131:
Superseded by #2423, which lands the same end state as one commit from `main`. The regenerate path this PR's stack introduced is deleted there, so reviewing the two in sequence would mean reading churn.

Issue #2210, opened 2026-08-06
Builtin Herdr reporter with workflow lifecycle authority


flora131:
# Builtin Herdr Reporter with Workflow Lifecycle Authority — Technical Design

| Document Metadata | Details |
| --- | --- |
| Author | Norin Lavaee |
| Status | Draft (RFC) — requirements frozen, open questions in §9 |
| Compatibility posture | **Additive only. No breaking changes.** See §10. |
| Delivery | Two phases, two stacked PRs (`gh stack`) |

## 1. Executive Summary

Atomic running inside a [Herdr](https://herdr.dev) pane is invisible to it: Herdr falls back to scraping the terminal, and a backgrounded workflow run leaves the pane reading *done* while the run works for hours — or sits blocked on a prompt nobody sees.

This adds a builtin reporter that makes Atomic a **lifecycle authority** in Herdr, plus the core primitive it needs. Two doors carry the change. `awaitUserDecision` is the one honest way for any Atomic surface to declare that the agent is parked on a human — a dialog, a trust prompt, a workflow HIL prompt. `reportPaneState` ⚠ is the single chokepoint through which any pane state reaches Herdr's socket. Everything else is a pure reducer over observed lifecycle.

Unlike every existing Herdr integration, Atomic also reports **workflow run lifecycle**, so a pane tracks the work Atomic actually dispatched rather than only the chat turn in front of it.

### 2.1 Current State

A Herdr pane exports `HERDR_ENV`, `HERDR_PANE_ID`, `HERDR_BIN_PATH`, `HERDR_SOCKET_PATH`. An integration reports semantic state over the pane socket and, when it does, Herdr stops using screen-manifest fallback for that pane. Agents split into *lifecycle authority* (Pi, OMP, Kimi, OpenCode, Kilo, MastraCode) and *session identity only* (Claude Code, Codex, Copilot, Devin, Droid, …).

Atomic is in neither group. `grep -ril herdr` across this repo matches one unrelated eval fixture.

**Leaking doors today:**

- **No door for "waiting on a human."** `pi.ui.select` / `confirm` / `input` / `custom` / `editor` each park the agent on a person, and each does so silently. No observer — Herdr, a statusline, a notifier — can distinguish *thinking* from *waiting*. The obligation exists; no door discharges it.
[...]

Issue #2022, opened 2026-07-26
bug(workflows): /new re-registers process-global DBOS workflow and degrades durability


flora131:
## What happened?

After DBOS has initialized, starting a new chat with `/new` in the same Atomic process can reload the bundled workflows extension and make the replacement session attempt to register the process-global DBOS workflow a second time.

The second registration fails:

```text
atomic-workflows: durable backend unavailable — continuing NON-DURABLY with an in-memory backend. Workflow runs will execute, but their state will not survive this process and `/workflow resume` after exit will not work. Restore durability by fixing Postgres provisioning: DBOS workflow durability configuration failed: Operation (Name: .atomicWorkflowHandle) is already registered.
Prior registration occurred at:
>>> DBOS.registerWorkflow (.../@dbos-inc/dbos-sdk/dist/src/dbos.js:1221:40)
>>> configureDbosDurableBackend (.../@bastani/atomic/dist/builtin/workflows/src/durable/dbos-backend.ts:107:28)
>>> async configureDbosOnce (.../@bastani/atomic/dist/builtin/workflows/src/durable/dbos-lifecycle.ts:76:10)
>>> async getReadyDbosBackend (.../@bastani/atomic/dist/builtin/workflows/src/durable/dbos-lifecycle.ts:111:3)
>>> async initializeDurableBackend (.../@bastani/atomic/dist/builtin/workflows/src/durable/factory.ts:54:10)
>>> async ensureDbosReady (.../@bastani/atomic/dist/builtin/workflows/src/extension/runtime.ts:141:5)
```

The subsequently launched `overnight-social-listening` workflow completed, confirming the documented in-memory degradation path. However, its workflow state was process-local and could not be relied on for cross-process `/workflow resume`.

This is not a Postgres connectivity/provisioning failure. The actionable cause is a duplicate DBOS operation registration, so the fallback warning's final sentence is misleading in this case.

This appears to be a follow-on regression to #1957 / #1958. That fix correctly keeps the DBOS executor and registry alive across process-preserving session replacement, but a freshly evaluated workflows extension has fresh module-local lifecycle state and registers the same fixed operation name into the still-live DBOS SDK registry.

### Impact

- Workflow actions in the replacement session silently switch from DBOS/Postgres durability to `InMemoryDurableBackend`.
- Runs can complete while the process remains alive, but checkpoints do not survive process exit.
- A crash/restart can therefore lose resumability and durable `ctx.tool(...)` replay protection.
[...]

PR #2416, opened 2026-08-15
fix(herdr,workflows): workflow lifecycle reporting and survival across session replacement (phase 2)

### LINKED ISSUE #2247: Non-quit session shutdown (/reload, /new, fork) orphans in-flight workflow runs instead of pausing them like quit does
## Summary

A `session_shutdown` with reason `quit` gracefully pauses every in-flight workflow run (`quitAllRuns`), leaving them resumable via `/workflow resume`. Every other shutdown reason — `reload`, `new`, `resume`, `fork` — clears stage handles synchronously without pausing anything, so an in-flight run is orphaned: its stage sessions stop mid-turn, and it is not resumable afterward. `/reload` is a routine user action, so a settings tweak or extension change can silently kill hours of background workflow work.

Observed on the released `@bastani/atomic` 0.9.12.

## Where

`packages/workflows/src/extension/extension-lifecycle.ts:116-155`:

- `reason === "quit"`: `quitAllRuns({ store, stageControlRegistry })` pauses runs, handles are disposed after pausing, `shutdownDbosQuietly()`. The comment states the intent: *"CLI/orchestrator quit is a resumable process boundary, not destructive cancellation. Durable-progress workflows stay available through `/workflow resume`."*
- any other reason: `stageControlRegistry.clear()` (comment: invalidate detached lazy handles "before they can attach to the replacement session"), `setNotificationsActive(false)`, `flushDbosQuietly()` — no `quitAllRuns`, no pause.

## Reproduction (what happened to us)

1. Launch a long multi-stage workflow run (`ralph`) from an interactive TUI session; let it run in the background across several stages.
2. Run `/reload` in that session for an unrelated reason (an extension change).
3. The in-flight stage's session file stops being written mid-turn and never resumes. The replacement session's workflow registry lists no runs; a resume attempt for the run id returns "No resumable workflow found".
4. Durable state (run directory, artifacts, committed work in the run's worktree) survives; the continuation is lost. In our case a stage that had been working ~90 minutes died silently.

## Why this seems like a gap rather than a choice

The quit path shows process boundaries were considered and graceful resumability built — but a mid-run `/reload` lands in the worst state: not running (the engine lived in the replaced extension host) and not resumable (the run was never paused into a resumable state). If auto-reattach across a reload is deliberately out of scope (plausibly hard — live stage sessions would have to survive the host swap), two cheaper mitigations would still remove the data-loss edge:

1. Treat non-quit shutdown like quit for run lifecycle: `quitAllRuns` before clearing handles, so runs land paused/resumable.
2. Or at minimum, warn before `/reload`/`/new`/fork when runs are in flight ("N workflow runs are active and will be terminated — continue?").

Happy to work on this if assigned.

Related but distinct neighbors:

   - #2022 (open) is the complementary half of the same non-quit-boundary family: this issue is the in-flight run dying **at** the boundary; #2022 is the replacement session's durability degrading **after** it (duplicate DBOS registration → silent in-memory fallback). A
 fix for either should probably consider both.
   - #2022 cites #1957/#1958, which deliberately keep the DBOS executor and registry alive across process-preserving session replacement — i.e. the design intent appears to be that runs survive these boundaries. That suggests the orphaning here is a gap (or an interaction
 with the same lifecycle-state reset #2022 describes) rather than a deliberate choice, since the surviving executor has no one driving the run once `stageControlRegistry.clear()` discards the handles without pausing.
   - #1722 (closed) covered re-attaching a live background workflow on `--session` resume — adjacent, but it presumes the run survived; in this path it never does.

makgunay:
> **Stacked PR:** builds on #2415 (`herdr-phase1`). This branch contains PR #2415's 33 commits plus the 12 phase-2 commits; review the range `herdr-phase1..herdr-phase2` (the last 12 commits). Will be rebased once #2415 merges.

**Base:** `herdr-phase1` · **Head:** `herdr-phase2` @ `b06ef6d4304ddc317905f50f2093f771e9d95f61` · 12 commits · 43 files (+3845/−190)
**Closes #2247** · Related: #2210 · Herdr-side: herdrdev/herdr#2423

> Stacked on the phase-1 PR (`herdr-phase1` → `main`). Review this diff against `herdr-phase1`, not `main`.

## Problem

Three, in escalating severity:

1. **The pane cannot see workflow runs.** Phase 1 reports chat-turn state only, so a pane running a workflow reads `idle` while stages execute, and a human-in-the-loop gate never reads `blocked`.
2. **Session replacement kills in-flight workflows (#2247).** Every successor `session_start` ran `killAllRuns()` + `store.clear()` (`extension-lifecycle.ts:91-92`), so `/reload`, `/fork`, and `/new` destroyed live runs. Upstream still does this on current `main` (`04e83e38e`) — verified during the rebase.
3. **Even without the kill, `/reload` loses the runs anyway.** `/reload` re-evaluates the workflows extension's whole module graph while the session keeps running (`clearExtensionCache()` + generation-busted specifiers, `loader-virtual-modules.ts:361-366`, jiti `moduleCache: false` at `:484`), so every run-scoped module singleton restarts empty: `/workflow status` answered "run not found", the bridge dropped the live contribution (pane idle mid-run), and the completion notice never arrived — the predecessor's notifications were uninstalled at shutdown and its API facades staled by `invalidate()`.

Two further defects were found by live testing of the stack itself; both are phase-1-introduced and disclosed in PR1:

4. **(A) Blocked-contribution leak:** a two-gate workflow wedged the pane at `blocked` from gate 1 to pane close, through answers and completion.
5. **(B) pi-asset coexistence:** with Herdr's installed `~/.pi/agent/extensions/herdr-agent-state.ts` present (the real-user default), the pane sat at `none:unknown` forever.

## Solution

**Lifecycle bridge** (`fa40536b7`, refined by `4c3756bdd`, `c23dad4ec`, `f30fbff7b`, `0f756b5c4`): the workflows extension publishes run lifecycle events; the Herdr builtin consumes them as pane contributions. State is derived once and committed before publishing; chat notices and the bridge share one `deriveRunLifecycle` path; the bridge handoff is bounded to the handoff itself.

**Survival across replacement** (`6556cb6ac`, `74e2cce85`, `58a90e881`, `9e605fdf1`): stop the successor's `killAllRuns`/`store.clear` teardown, preserve live stages and reconstruct pane contributions across every session replacement.

**Survival across `/reload`'s module re-evaluation** (`df630c97d`): new host API `sessionScopedExtensionState(scope, key, create)` (`packages/coding-agent/src/core/extension-session-state.ts`, exported from `src/index.ts`) — a WeakMap of keyed state per canonical event bus, the one object that outlives every extension load generation. The workflows package's six run-scoped singletons (`store`, `stageControlRegistry`, `cancellationRegistry`, `toolControlRegistry`, `jobTracker`, `stageUiBroker`) become stable facades adopted to the host session at factory time (`shared/session-scoped-singleton.ts`, `extension/adopt-session-run-state.ts`). Without a host bus (unit tests, embedded use) the module-local instance is used — prior behavior unchanged. Containment: `makePersistencePort` swallows exactly the stale-extension-context rejection on transcript appends from the orphaned predecessor graph.

**Defect A fix** (`10b430bbe`): the bridge was proven **innocent** — reproduced three ways, it published the correct `awaiting_input → resumed → … → completed` sequence with an empty contribution map at completion. The leak was a user-decision block minted for the workflow graph overlay: the overlay hides on Ctrl+X via its `OverlayHandle` without settling (a deliberate, preserved remount-avoidance contract, pre-existing on `main`), and the phase-1 wrapper released the "Custom dialog" block only at settlement. The single-vs-multi-gate discriminator was the auto-attached graph overlay (inputs form), **not** gate count — do not chase the bridge. Fix in the core wrapper (`runner-ui-blocks.ts`): interpose on `options.onHandle`; hidden overlay ⇒ block released; re-shown ⇒ block re-minted; settlement ends bookkeeping exactly once; callers without `onHandle` keep exact argument identity.

**Defect B fix** (`b06ef6d43`): the builtin blindly stood down whenever `herdr-agent-state.ts/js` merely loaded, while the current Herdr server ACKs but discards that asset's `herdr:pi` reports. Replaced with explicit load-time, env-gated supersession: inside a Herdr pane the resource loader skips the installed integration files by basename at both extension-path assembly points (never a second writer, even if a future server accepts `herdr:pi`); the builtin reports whenever active. Out-of-pane behavior is byte-identical.

## Regression evidence (fail first, then pass)

- **#2247 / reload survival:** `test/unit/workflow-run-state-real-reload.test.ts` replays the host's exact `/reload` sequence — `session_shutdown("reload")` → `runtime.invalidate()` → genuine module-graph re-evaluation → fresh generation → `session_start("reload")`. On pre-fix sources (verified at `503019133`, the pre-rebase twin of `58a90e881`) it fails at the live symptom: `AssertionError: successor session must still observe the in-flight run after /reload`. Post-fix: passes.
- **Defect A:** `test/unit/herdr-pane-workflow-multi-gate.test.ts` (real executor run, two `ctx.ui` gates, real bridge, real builtin, real block door, real overlay adapter, socket fixture). With the fix stashed it fails at the live symptom: `timed out waiting for working; saw [... {"state":"blocked","message":"Custom dialog"}]`. Post-fix: passes. Plus 4 new wrapper unit tests in `extensions-ui-block-wrapping.test.ts` (all 39 in file pass).
- **Defect B:** new `herdr-supersession.test.ts` (real `DefaultResourceLoader`; asset skipped with pane env set, loaded without) and flipped stand-down tests in `herdr-activation.test.ts` / `herdr-extension-integration.test.ts`. With the B wiring stashed, **5 tests fail**; post-fix all pass.

## Test output

At `b06ef6d43` (post-rebase verification):

- `npm run typecheck` — clean on the first run (root `tsc --noEmit` + coding-agent `tsgo`). `npx biome check` on the hand-merged `extension-runtime-state.ts` — clean.
- Root unit, all green: bridge-session-replacement, real-reload, lifecycle-notifications-01, delivery-failures, herdr-pane-multi-gate, extension (**90 tests**) and all four workflow-heartbeat suites (**102 tests**).
- coding-agent, branch-touched files: 7 files, **126 tests**, green.

Earlier full runs on pre-rebase twins (content lines byte-identical per the transplant-fidelity check): at `e302bf24f`, `npm run test:unit` **6193 passed / 1 skipped** (639 files) and `npm run test:integration` **487 passed** (37 files); at `d9ee0587a`, final fix-effort runs **202 coding-agent + 136 root tests** green, `npm run typecheck` and `npm run check` clean.

## Live verification

**Scenario matrix at this exact tip** (`b06ef6d43`, live Herdr pane `w1J:pF`, real session, real socket, real workflow runs, pi asset present): **9/9 pass** (`.atomic/workflows/runs/herdr-live-matrix/matrix-i1.json`).

| # | Scenario | Observed |
|---|----------|----------|
| s1 | idle-baseline | `launch=atomic:idle` |
| s2 | turn-working | `idle` → `working` sustained across 11 polls → `done` |
| s3 | dialog-blocked | `dialog=atomic:blocked after=atomic:done` |
| s4 | run-while-idle | `early=working chatIdle=working after=done` |
| s5 | concurrent-refcount | `bothLive=working shortDone=working bothDone=done` |
| s6 | block-precedence | `run=working dialog=blocked back=working settled=done` |
| s7 | hil-awaiting-input | `blockedSeen=true settled=atomic:done completed=true` |
| s8 | reload-survival | `pre=working`, 4 post-reload polls all `working`, `notice=true` |
| s9 | quit-release | `final=none:unknown` |

**W8 `/reload` probe** (pane `w1J:p5`, at `e302bf24f`): `livetest seconds=45`, `/reload` mid-run → pane stayed `working` (8 samples), background panel survived, `/workflow status` found the run, both stages completed (~1m30s), completion notice **delivered in the successor**, pane → `done`. `runSurvived: true`, `pass: true`.

**Two-gate + coexistence probe** (pane `w1J:pD`, at `d9ee0587a`, pi asset present throughout): `coexistOk`, `g1BlockedSeen`, `postG1WorkingSeen`, `g2BlockedSeen`, `postG2WorkingSeen`, `completedIdle`, `postSettleIdle` all true; `flapSeen` false.

## Reviewer scrutiny points

1. **`makeTerminalNotice` merge** (`lifecycle-notifications.ts`, lines 531–567) — the one place upstream's #2330 author-exit refactor and this branch overlap inside single functions. The merged function carries both the #2330 fields (`outputs`, failed-author-exit `resumable`, `exitReason` fallback) and phase 2's shape; `deriveRunLifecycle`/`deriveCurrentRunState` route through it. `outputs` never enters a bridge contribution (kind/label only), so nothing new crosses the Herdr socket. Verified by hand; please read it.
2. **Facade-registry placement** — `registerEventBusFacade`/`canonicalEventBusFor` live in `event-bus.ts`, not the lifecycle module, so the loader does not import workflow code. If upstream later gives extensions direct bus access, the registry becomes a resolvable no-op.
3. **Loader-survival assumption** — the `/reload` fix assumes the `ResourceLoader` (and its `eventBus`) survives session replacement; confirmed on this base (`resource-loader-reload.ts` passes `state.eventBus` through). A future upstream change that rebuilds the loader per session would empty the handoff again; the "later load" test covers facade churn over one bus only.
4. **`sessionScopedExtensionState` host API** — a new public core API (WeakMap keyed on the canonical event bus). Keys carry `@1` version suffixes to force a fresh start on deliberate shape changes; an incompatible *unstamped* change would hand new code an old-shaped instance. Kept in its own conventional commit rather than a fixup so the API is not buried in a workflows-only change.
5. **`extension-runtime-state.ts` interleave** — commits 1, 3, and 7 merged the bridge-handoff rewrites (`takeWorkflowLifecycleBridgeHandoff`) into upstream's new heartbeat-scheduler context (commit 1 had an import-block conflict; 3 and 7 auto-merged). Line placement hand-verified, tests cover the handoff path, but this is the file where the two series interleave most tightly.
6. **Heartbeat-across-replacement suppression (open decision, no test pins it):** a heartbeat card admitted before a replacement and consumed after it fails exact pending ownership and is excluded from model context even though its run is preserved and live. The next future boundary re-raises; worst case one suppressed steer per replacement, never a stale one. Decide: accept-by-design (add a pinning test) or follow up to let a preserved run's card survive the boundary.
7. **Non-blocking review notes / residual uncertainties:**
   - Why Herdr's server discards `herdr:pi` reports for an Atomic pane is inferred from black-box socket probes (likely agent-vs-process validation); the server source was not read. The fix does not depend on the reason.
   - If the TUI host ever hides an overlay internally (not via the caller's handle), the block would not suspend; no such path exists today for `custom()` overlays.
   - Pre-fix, the pane label with a hidden overlay plus a waiting gate was `"Custom dialog"`, not the gate label; the original findings never recorded the label, so that detail could not be cross-checked.
   - Transcript entries appended by the orphaned predecessor graph after `/reload` are dropped (stale-context containment) rather than rerouted; run state, notices, and pane reporting are unaffected.
   - The phase-2 replacement tests still inject raw buses as `pi.events`; only the scope test exercises the loader-built facade path.
   - Why runs appeared to survive reload on the pre-rebase base is unexplained (likely a build flavor with retained module graphs); the fix makes preservation independent of build flavor.

## Breaking changes / follow-ups

- **No breaking changes.** New additive host API: `sessionScopedExtensionState` (see scrutiny point 4). Out-of-pane and embedded behavior unchanged.
- **CHANGELOG placement:** the workflows package's #2210 bullets share `[Unreleased]` with upstream's #2401 entries after `0.9.14-alpha.1` was cut; confirm the shipping vehicle.
- **Follow-ups:** the heartbeat-suppression decision (scrutiny point 6); watch during dogfooding for panes stuck `blocked` after custom UI other than the graph overlay, refcount drift across long sessions, and label behavior if a future Herdr server starts accepting `herdr:pi` (the supersession fix already prevents a second writer).

---
<a href="https://app.blacksmith.sh/bastani-inc/codesmith/atomic/pr/2416"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789370708&installation_model_id=10562&pr_number=2416&repository=bastani-inc%2Fatomic&return_to=https%3A%2F%2Fgithub.com%2Fbastani-inc%2Fatomic%2Fpull%2F2416&signature=7785fcaffcda4f316f483b3271bd43eec9ed7ae81f07cad8ec6af589eeaf7f69"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

<h3>Greptile Summary</h3>

This change preserves active top-level workflow reporting when sessions are reloaded, replaced, resumed, or forked, and ensures Herdr reflects terminal workflow cleanup, user-dialog priority, and custom-overlay visibility correctly.

The exercised workflow replacement path confirms that live stage controls now survive the session boundary. The same behavior failed before these changes, while the updated workflow, Herdr socket, lifecycle-priority, and hidden-overlay checks pass with the change. No defects were found.

The changed workflow lifecycle and Herdr reporting behavior is safe to merge based on direct execution of session replacement, workflow-state reconciliation, socket reporting, and overlay visibility flows.

No actionable findings remain. The focused checks exercised every session replacement mode, terminal workflow cleanup, user-dialog precedence, and hidden custom-overlay behavior, with the updated implementation passing all covered assertions.

**Files Needing Attention:** No files require follow-up. The primary exercised areas were workflow lifecycle state, session-scoped run adoption, Herdr reporting, and UI block wrapping.

**What T-Rex did**
- Ran the focused session-replacement regression against the pre-change baseline and observed 17 test failures due to the predecessor clearing the live stage-control handle at the session boundary.
- Re-ran the replacement flow and the real Herdr multi-gate workflow; all 26 tests passed, showing live workflow state persists across replacement and terminal runs stop contributing.
- Ran the narrow Herdr lifecycle-priority and custom-overlay visibility suites; all 74 tests passed, confirming open user dialogs outrank workflow state and hidden custom overlays release their blocks.
- Before: the current replacement regression test failed against HEAD~8; After: the same regression suite passed 26/26, and the Herdr reducer/UI-block suites passed 74/74.

<a href="https://app.greptile.com/trex/runs/19016064/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(herdr): supersede the installed pi i..."](https://github.com/bastani-inc/atomic/commit/b06ef6d4304ddc317905f50f2093f771e9d95f61) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53584300)</sub>


makgunay:
Note for reviewers: maintainer guidance on #2247 ([comment](https://github.com/bastani-inc/atomic/issues/2247#issuecomment-5301142430:~:text=)) crossed with this PR's submission by minutes. This PR implements the boundary as *continuation* (runs keep executing across replacement) rather than the shared graceful-pause shape proposed there; the mapping against each point of that guidance, and our offer to rework toward pause-parity if preferred, is in [this #2247 reply](https://github.com/bastani-inc/atomic/issues/2247#issuecomment-5302029955).

lavaman131:
@makgunay Thanks for this one. Clearing the bridge on defect A before pinning it on the overlay's hide-without-settle contract was the right order to work in, and writing down that the discriminator was the auto-attached graph overlay rather than gate count will save whoever hits this next. The live matrix and the fail-first evidence are useful.

Two problems: how this landed, and how big it is.

## This PR isn't actually stacked

The description says the base is `herdr-phase1` and asks reviewers to read `herdr-phase1..herdr-phase2`. The real base on GitHub is `main`, so the PR shows all **45 commits, +9,626 / −134 across 75 files** — phase 1 in full plus your 12.

Not your mistake. Both PRs come from your fork, and a fork PR can only target a branch that exists upstream. `herdr-phase1` is in `makgunay/atomic`, so GitHub retargeted to `main`. `gh stack` can't chain across a fork boundary.

The practical consequence is the important part: **upstream never renders this as a stack.** There's no per-layer view here, so the "review `herdr-phase1..herdr-phase2`" instruction can't be honored by the UI — a reviewer opens the PR and gets one 9,626-line cumulative diff with phase 1 in it twice. Asking people to mentally subtract 33 commits isn't a review process.

**So our clear preference is separate, smaller PRs rather than a stack.** One focused PR against `main` at a time — we review and land it, you rebase and open the next. It costs you a rebase per layer, but it's the only shape that's genuinely reviewable on our side, and each piece moves quickly instead of the whole thing blocking on one enormous review. Rough target: something readable in a sitting, a few hundred to ~1,000 lines.

## Size

Worth saying that even your intended scope is too big for one PR. The 12 phase-2 commits alone are +3,845 / −190 across 43 files and do at least five separable things — a new public core API, two independent lifecycle fixes, the bridge, and two defect fixes. [`CONTRIBUTING.md`](../blob/main/CONTRIBUTING.md) asks for changes "small enough to review," and that applies to the intended diff, not just the accidental one.

Suggested split, following your commit structure:

1. **`sessionScopedExtensionState`** — `packages/coding-agent/src/core/extension-session-state.ts` plus the `src/index.ts` export. New public core API; should be reviewed on its own, not as a detail inside a workflows fix. You already isolated it in its own commit (scrutiny point 4) — same reasoning, one step further.
2. **Run-state survival across `/reload`** — the six singleton facades, `session-scoped-singleton.ts`, `adopt-session-run-state.ts`, `workflow-run-state-real-reload.test.ts`.
3. **Session-replacement lifecycle** — `6556cb6ac`, `74e2cce85`, `58a90e881`, `9e605fdf1`.
4. **Workflow lifecycle bridge** — `fa40536b7` and refinements, plus the Herdr consumption side.
5. **Defect A / B fixes** — `10b430bbe`, `b06ef6d43`, folded into the PRs that introduce what they fix (see #2415).

## #2247 and #2022

We've moved #2247 to P0 and are taking it, together with #2022 (the `/new` DBOS re-registration bug) as one piece of work — they're the two halves of the same non-quit-boundary problem, as you noted in the issue. Losing hours of durable background work on a routine `/reload` is data loss, and we don't want it gated behind a 9,626-line feature review.

**We're building on your commits, not around them.** `df630c97d` cherry-picks onto current `main` with conflicts only in the two CHANGELOG files — all 15 source and test files apply clean, including your fail-first test. Cherry-pick keeps your authorship; we'll add `-x` provenance and a co-author trailer. `6556cb6ac` and `58a90e881` conflict against the phase-2 bridge, so we'll port those by hand.

Reasoning, including where we land on continuation vs pause-parity, is in [the issue](https://github.com/bastani-inc/atomic/issues/2247). For you: those commits drop out on rebase and this PR ends up scoped to the lifecycle reporting.

## What we'd want fixed

- **Scrutiny point 6 — heartbeat suppression across replacement.** You flag it as an open decision with no test pinning it. We shouldn't merge an undecided behavior. Our preference is accept-by-design plus the pinning test, so it lives in a test rather than a PR description nobody reads after merge.
- **`makeTerminalNotice`** (`lifecycle-notifications.ts:531–567`, scrutiny point 1) — agreed, highest-risk merge on the branch. Much easier to review in a small PR than as one of 75 files.
- **`.ts` / `.js` import inconsistency** carried up from #2415.
- **Replacement tests still inject raw buses as `pi.events`** — you note only the scope test covers the loader-built facade path, which is the one that actually runs. Worth closing.
- **Required CI hasn't run** (fork PR needs approval). We'll approve on the split PRs.

We want the Herdr work. Splitting it is how it gets reviewed and lands sooner.

flora131:
Following the closure of #2415 and the updated direction in #2551, we're closing this cumulative second phase as superseded as well.

The workflow-survival and session-replacement concerns have since been addressed through focused changes on `main`. The remaining Herdr lifecycle bridge should be rebuilt more narrowly on the current lifecycle events after the Herdr maintainers settle the upstream contract in herdrdev/herdr#2423.

We'll retain this PR's implementation, tests, and investigation as reference material and preserve your attribution for any code we carry forward. No additional work is needed from you on this branch. Thanks again for the thorough work.

PR #2423, opened 2026-08-15
refactor(workflows): drive open-claude-design's live review from the workflow

### LINKED ISSUE #2411: refactor(workflows): open-claude-design wraps a bounded loop around impeccable live's unbounded one
## Summary

`open-claude-design` wraps a bounded refinement loop (`generate-N` → `user-feedback-N`, capped by `max_refinements`) around an `impeccable live` session that is itself an unbounded interactive loop. The outer loop duplicates the inner one, and `max_refinements` bounds the wrong thing.

## Evidence

`packages/workflows/builtin/open-claude-design-setup.ts` instructs the feedback stage to run `/skill:impeccable live` and "keep iterating until the user signals they are done with this round."

The live reference (`impeccable/reference/live.md`) shows what that session already covers:

- element pick → three on-brand variants → the accepted one is written into `preview.html` **in place**
- `steer` → the agent performs page edits and navigation help **during the session**
- the poll loop continues until the user emits `exit`

So by the time a `user-feedback-N` stage returns, every accepted variant and most steered direction is already applied to the artifact. The subsequent `generate-N+1` round then re-reads that file and rewrites it whole, with `live_changes` passed in mainly so it does not revert work the user already accepted.

## Two concrete defects that follow

1. **A no-op refinement round.** The stage prompt says `decision` must be `revise` "whenever the user asked for ANYTHING at all — any note, any accepted variant, any steer." A session where the user only accepted variants therefore reports `revise`, and a full `generate-N+1` fires whose only real job is to avoid undoing what live already did. It applies nothing new.
2. **`max_refinements` caps the wrong loop.** It bounds review *rounds*, but the inner live session is unbounded, so the number that users read as "how many times can I iterate" does not describe their actual iteration budget.

## Proposed shape

One review session, with the outer loop reduced to a rare escape hatch:

```
discovery -> research -> generate-1 -> gate -> live review session (unbounded)
                              ^                        |
                              |                    decision
                              +---- "regenerate" ------+
                                                       |
                                    "export" ----------+---> exporter -> final-display
```

- `decision` becomes `export | regenerate`.
- `regenerate` exists for the one thing live cannot do: start over from the brief, design context, and references when the user rejects the direction rather than the details. Live only ever edits in place.
- `user_notes` becomes the brief for a `regenerate` only. Accepted variants need no round at all; they are already in the file.
- `max_refinements` becomes a cap on regenerations, which is what the name already implies to users.

## Expected deletions

- the pending-feedback threading through the refinement loop
- the final-round revision special case (`generate-(max+1)`)
- `assertUserAnnotationsThreaded` and the `live_changes` preservation plumbing that exists only because a whole-file rewrite follows the session
- most of the remaining body of `refineOpenClaudeDesign`

## Acceptance criteria

- [ ] A session where the user accepts variants and writes no notes exports directly, with no `generate-2`.
- [ ] A session where the user rejects the direction produces exactly one `generate-2` from the brief plus their notes, then exports.
- [ ] `max_refinements` bounds regenerations, and the docs say so.
- [ ] The durable per-round feedback artifact and its path in the exporter/generate `reads` survive.
- [ ] The `#2401` continuation regression test still passes unchanged: the structured answer stays the authoritative deliverable.
- [ ] `npm run check`, `test:unit`, `test:integration`, and `test:ci-contracts` are green.

## Non-goals

- Changing `impeccable live` itself.
- Changing shared workflow-runtime result selection.
- Reintroducing any prose parsing of the feedback report. Structured output stays the only decision channel (#2401).

## Context

Follow-up to #2401 (PR #2408), which made the structured answer the authoritative deliverable and deleted 1,344 lines of prose-parsing defense. This issue removes the redundant loop that the same investigation exposed.

flora131:
## Summary

`open-claude-design` wrapped a bounded generate/review loop around an `impeccable live` session that is itself unbounded, and the review stage owned the poll loop. This makes the workflow drive the live protocol, ends a review on the helper's own `exit` event, and exports what the session produced.

Net: **+741 / −1879**.

## Why

Both problems came from the skill's chat-session design, where "the review is over" needs no ceremony because the user simply keeps talking. A workflow stage boundary is a commit point, so the same ambiguity silently discarded work.

1. **A model decided when a review ended.** The live contract's `"timeout" -> LOOP` is an instruction to a model, not a constraint the runtime enforces. A stage that concluded on a poll timeout returned a perfectly valid structured answer for a review the user was still in, and the refinement loop routed on it. Same failure class as #2401, which stopped a late turn from *replacing* a finalized answer but did nothing about one produced too early.
2. **The outer loop duplicated the inner one.** Accepted variants and steered edits are written into `preview.html` during the session, so the following `generate-N+1` existed mainly to avoid undoing them, and `max_refinements` bounded review rounds while the session inside them had no bound at all.

## What the review looks like now

| Node | Kind | Role |
|---|---|---|
| `user-feedback-N-start` | stage | boots `/skill:impeccable live`, opens the page, prints the review URL and how to end the review, then stops |
| `live-poll-N-M` | tool | polls the helper; absorbs `timeout` internally |
| `live-generate-*`, `live-steer-*`, `live-manual_edit_apply-*` | stage | handle exactly the events that need a model |
| `live-reply-N-M` | tool | acknowledges each one (`done` / `steer_done`) |

- `accept`, `discard`, and `prefetch` are acknowledged by the poll script itself and mint no stage.
- `timeout` is absorbed inside the poll node, so an idle hour costs no graph nodes and never looks like an ending.
- Unreadable poll output is treated as a timeout, not an ending.
- The loop ends only on the helper's `exit`, and the run exports the preview as it then stands.

Termination is an event, not a judgment.

## What that made redundant

With the session unbounded and its work already applied, the second opinion after it had nothing left to decide. Deleted: the refinement loop, the feedback schema, the summary stage, `open-claude-design-feedback.ts` in full, and `<artifact_dir>/feedback/`. Wanting a different direction means re-running the workflow with a better brief, which is where a direction actually comes from.

## Ending a review

Nothing but the user ends one. The run-level gate now states the three ways — exit in the Impeccable overlay, closing the tab, or `exit live` — that the session waits indefinitely until then, and that ending it exports immediately with no further round. The session-start stage prints the same under the live review URL.

## Breaking changes

- removed inputs: `max_refinements`
- removed outputs: `approved_for_export`, `refinements_completed`
- removed artifacts: `<artifact_dir>/feedback/` (JSON record, Markdown copy, annotated-snapshot copies)
- removed module: `open-claude-design-feedback.ts` and every export
- the bundled impeccable skill is always used; a project-vendored copy under `.agents/skills` is deliberately ignored, because the loop depends on `live-poll.mjs`'s CLI surface, reply tokens, and event vocabulary, and the bundled scripts are versioned and tested with this workflow
- no model-driven fallback path

Remaining inputs: `prompt`, `discover_references`.

## Tests

New `test/unit/open-claude-design-live-protocol.test.ts`: event parsing, unreadable output as a timeout, timeout absorption, bundled-script resolution, which events call the model back, reply tokens, and the prompts including how a review ends.

Workflow-level: the loop polls, mints a stage per model-facing event, replies to exactly those, ends on `exit`, and reaches `exporter`; the gate's skip option still exports without a session.

`test/unit/executor-structured-output-continuation.test.ts` is untouched and passing — the #2401 regression proving a continuation cannot erase a finalized structured result.

The mock workflow context stubs `live-poll-*` / `live-reply-*` so no unit test spawns a real ten-minute long-poll.

## Verification

`npm run check`, `npm run test:unit` (6279), `npm run test:integration` (486), `npm run test:ci-contracts` (42) all green.

Closes #2411
Refs #2401

---
<a href="https://app.blacksmith.sh/bastani-inc/codesmith/atomic/pr/2423"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789405089&installation_model_id=10562&pr_number=2423&repository=bastani-inc%2Fatomic&return_to=https%3A%2F%2Fgithub.com%2Fbastani-inc%2Fatomic%2Fpull%2F2423&signature=cf4b9c77a2de8e7637715ea73d481b988205937dd1cc3ff85a6ad13f5d47f639"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

<h3>Greptile Summary</h3>

This change moves the Impeccable live-review loop into the workflow, including helper polling, model-event handling, and session exit. Executable checks confirmed that model-event acknowledgements omit the event identifier required by the bundled helper, so handled events remain pending and are delivered again. They also confirmed that failed poll commands with empty output are treated as timeouts and retried instead of reporting the helper error.

<h3>Confidence Score: 3/5</h3>

Not safe to merge until live-event acknowledgements include their event IDs and failed helper polls are surfaced to the workflow.

Two independent runtime failures were reproduced: acknowledgement requests are rejected and repeat the same event, while nonzero helper exits are silently retried. Neither failure is a security issue.

**Files Needing Attention:** packages/workflows/builtin/open-claude-design-live-protocol.ts needs the reply API and poll error handling corrected; packages/workflows/builtin/open-claude-design-phases.ts must pass the event ID into the acknowledgement call.

**What T-Rex did**
- T-Rex produced proofs for three posted P1 findings and linked them to the corresponding review comments.
- A general-contract validation exercise showed the harness injecting a ScriptResult and recording invocations, a first nonzero run with code 23 and transport unavailable, then a successful run ending with exit code 0, revealing an error-handling defect where failures can lead to unbounded retries.

<a href="https://app.greptile.com/trex/runs/19041954/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

1. General comment 

   <a href="#">P1</a> **Nonzero live-poll exits are silently retried as timeouts**

   - **Bug**
     - `pollLiveEvent` receives a nonzero `ScriptResult` with empty/non-event stdout, maps stdout to `timeout`, and loops. The injected runner reproduced this with exit code 23 and recorded a second poll rather than an error outcome.
   - **Cause**
     - Lines 159-161 parse only `result.stdout` and branch only on the parsed event type; `result.code` and `result.stderr` are never checked.
   - **Fix**
     - Immediately after `await run(...)`, check `result.code !== 0` and throw or return a structured failure that preserves the exit code and stderr before calling `parseLiveEvent`. Preserve the existing timeout retry only for successful (`code === 0`) polls.

</details>

`````markdown
### Issue 1
packages/workflows/builtin/open-claude-design-live-protocol.ts:174
**Reply arguments cannot acknowledge events**

`replyLiveEvent` invokes the bundled helper as `--reply <token>`, but the helper requires `--reply <event-id> <status>`. The workflow therefore sends `--reply steer_done`, which the helper interprets as an event ID without a status and exits without posting an acknowledgement. Carry the received event ID into this call and pass it before `replyTokenFor(event)`.

### Issue 2
packages/workflows/builtin/open-claude-design-live-protocol.ts:159-161
**Nonzero poll exits are retried as timeouts**

`pollLiveEvent` parses only stdout and never checks `result.code`. A failed helper invocation with empty or non-event stdout becomes a synthetic `timeout`, so the loop begins another poll rather than surfacing the transport, authentication, or server error. Check for a nonzero exit after `run` and preserve stderr in the error; retry only successful timeout responses.

### Issue 3
packages/workflows/builtin/open-claude-design-live-protocol.ts:122-123
**Event payload shape remains ambiguous**

The new parser represents decoded protocol data as `unknown` and then casts it to `Record<string, unknown>` rather than defining a specific parsed-event shape. This weakens static checking when the live protocol gains or changes fields and conflicts with the repository requirement to use specific types.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["refactor(workflows): drive open-claude-d..."](https://github.com/bastani-inc/atomic/commit/ed6b4f3b071b06cbaf4bb6bd706dc53e8d6b24f8) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53637729)</sub>

> Greptile also left **3 inline comments** on this PR.

**Context used:**

### CODE REVIEW

**Comment by flora131** [comment]
## The review after this change

```mermaid
flowchart TD
  D["discovery<br/>brief, output type, references"] --> R["research<br/>design-context.md + references.md"]
  R --> G["generate-1<br/>writes preview.html"]
  G --> GATE{{"run-level gate — ctx.ui.select<br/>Start live review / Skip and export"}}
  GATE -->|"skip"| EXP
  GATE -->|"start"| START["user-feedback-1-start — STAGE<br/>boots /skill:impeccable live<br/>prints review URL + how to end it<br/>then stops"]

  subgraph LOOP["workflow-owned poll loop"]
    direction TB
    POLL["live-poll-1-M — TOOL<br/>timeout absorbed here"]
    EVT{"event"}
    HANDLE["live-generate / live-steer /<br/>live-manual_edit_apply — STAGE"]
    REPLY["live-reply-1-M — TOOL<br/>done / steer_done"]
    POLL --> EVT
    EVT -->|"generate, steer,<br/>manual_edit_apply"| HANDLE --> REPLY --> POLL
    EVT -->|"accept, discard, prefetch<br/>acked by the poll script"| POLL
  end

  START --> POLL
  EVT -->|"exit"| EXP["exporter<br/>HTML spec + handoff"]
  EXP --> FD["final-display"]

  USER[/"the user, in the browser overlay:<br/>exit button, close tab, or 'exit live'"/] -.->|"produces exit"| EVT
  SKILL[/"impeccable skill bundled in Atomic<br/>live-poll.mjs"/] -.-> POLL
```

**Legend.** `{{...}}` is a human prompt, `TOOL` nodes are durable `ctx.tool` calls, `STAGE` nodes are model turns.

**What the shape guarantees**

- Every arrow leaving the loop is a tool result or a helper event. The model boots the session, handles individual events, and never decides the review is over.
- `timeout` is absorbed inside the poll node, so silence costs no graph nodes and cannot end a review.
- There is exactly one terminator: `{"type":"exit"}`, produced by the user in the browser.
- Nothing runs between `exit` and `exporter` — no summary stage, no decision, no artifact.

**Before, for contrast:** `generate-N → gate → user-feedback-N (owned the poll loop) → decision → generate-N+1 …`, bounded by `max_refinements`, with the feedback round persisted to `<artifact_dir>/feedback/iteration-N.json`. A poll timeout inside that stage could end a review the user was still in.

PR #2452, opened 2026-08-16
fix(subagents): bound a non-terminating foreground child turn with a content-progress stream guard

### LINKED ISSUE #2446: subagents: non-terminating model turn can block foreground child indefinitely and prevent fallback
### Plain-English summary

An Atomic foreground subagent can get a model response that stays open but never finishes. The parent workflow then waits for that child indefinitely. Because the active turn never ends with a failure that can be handled, a configured fallback model is not reached.

This is a liveness and recovery gap in the current in-process design. It is not a claim that a documented timeout setting is broken: current Atomic intentionally has no per-attempt idle watchdog or child wall-clock kill cap.

### Technical summary

The current in-process runner records attempt state and subscribes to session events, then awaits the model turn directly:

```ts
await session.prompt(admitted.spec.task);
await terminating;
```

Source at current main (`66a10d1c858ebbd2aa9ea5e82b5dc3f88210c285`):

- Runner execution and prompt await: [runner.ts#L899-L960](https://github.com/bastani-inc/atomic/blob/66a10d1c858ebbd2aa9ea5e82b5dc3f88210c285/packages/subagents/src/runs/inprocess/runner.ts#L899-L960)
- Current behavior contract: [subagents.md#L210-L216](https://github.com/bastani-inc/atomic/blob/66a10d1c858ebbd2aa9ea5e82b5dc3f88210c285/packages/coding-agent/docs/subagents.md#L210-L216)

The documentation says there is no per-attempt idle watchdog, child wall-clock kill cap, or timeout-regex classification. A quiet provider response is allowed to finish; an explicit termination or provider failure must supply a retryable cause. The 0.9.13 migration notes record the intentional removal of the process-era watchdog: [subagents CHANGELOG.md#L50-L53](https://github.com/bastani-inc/atomic/blob/66a10d1c858ebbd2aa9ea5e82b5dc3f88210c285/packages/subagents/CHANGELOG.md#L50-L53).

If a model turn never settles and emits no terminal failure, the runner has no event on which to complete the attempt and make fallback handling reachable. The trigger that produced the non-terminating stream is not known; it could be in the provider path or the client/parser path, and raw wire evidence is needed to distinguish those possibilities.

### Reporter-observed incident

1. Run a foreground workflow with a primary and configured fallback candidate.
2. Complete one debugger helper normally and admit its result.
3. Launch a second foreground debugger helper.
4. Observe a model turn that remains open after tool arguments have completed and does not emit `message_end`.
5. Wait for the parent workflow to complete.

Reporter-observed result:

```text
consecutive whitespace-only toolcall_delta events: 32749
message_end: not observed
foreground child: pending
fallback attempt count: 0
parent foreground call: pending
pending duration: approximately 6.5 hours
```

All values above are reporter-observed. Private session data, prompts, model identifiers, run identifiers, raw streams, and the large local artifact are intentionally not included.

### Minimal maintainer probe

A deterministic repository-level liveness probe can use the existing `testSession.promptGate` seam:

1. Create a deferred promise and pass it as `testSession.promptGate` when admitting the in-process child.
2. Start the foreground child and leave the gate unresolved, modeling a prompt that has not produced a terminal result.
3. Wait through a short, fixed observation window.
4. Verify that the child remains pending and no automatic runner-level settlement occurs during that window.
5. Interrupt the child or resolve the gate for cleanup.

This probe demonstrates runner liveness only. The test session does not exercise the production SDK fallback machinery, so fallback-attempt count and fallback reachability remain reporter-observed incident evidence. Gate release, eventual provider settlement, explicit interruption, or process failure are separate ways the pending call may later end.

No public one-command production reproduction is included here. An end-to-end fallback assertion needs a controllable provider or instrumented real session that keeps the primary stream open without a terminal event while recording whether a configured fallback starts.

## Expected behavior

A non-terminating foreground subagent model turn should not hold an unattended parent forever without actionable state or a recovery path. Atomic should eventually either make an appropriate fallback candidate reachable or return control safely, while preserving valid long-running work where possible.

The invariant is bounded recoverability for an unattended foreground child, not a specific timeout or watchdog implementation.

## Impact

- A foreground parent workflow can remain pending indefinitely.
- Configured fallback candidates can remain unreachable because the active attempt never terminates with a classifiable failure.
- Human intervention may be required to recover the session.
- The issue affects recovery from a non-terminating in-process subagent turn; it does not establish that every quiet or long-running model response is faulty.

## Scope and unknowns

In scope is the foreground in-process subagent call and fallback reachability while its active model turn never settles.

Out of scope are claims that a removed process-era timeout setting is currently supported or broken, and claims that the provider or parser caused the observed stream. The malformed-stream trigger remains unknown without raw transport capture. It is also unknown whether a provider or SDK layer has an independent timeout in a particular deployment; the observed gap is that Atomic has no confirmed runner-level recovery when no terminal event arrives.

## Related issues

- [#1580](https://github.com/bastani-inc/atomic/issues/1580) covered the old process runner hanging without attempt bounds. It is closed and describes the process-era architecture, not this current in-process path.
- [#2188](https://github.com/bastani-inc/atomic/issues/2188) and [PR #2205](https://github.com/bastani-inc/atomic/pull/2205) introduced the in-process architecture and intentionally removed the process-era watchdog. This report concerns the liveness consequence of that current design, not a request to restore the old implementation.
- [#2425](https://github.com/bastani-inc/atomic/issues/2425) concerns workflow stages waiting on silent or contentless model streams. This report is narrower: a foreground in-process subagent call can remain pending and keep its configured fallback unreachable.

## Investigation guidance

Please evaluate the invariant that an unattended foreground child cannot remain indefinitely without actionable state and a safe recovery path. The implementation choice should follow from that invariant rather than from this report prescribing one code shape.

Options to evaluate include:

- an attempt wall-clock or idle bound, with care for long legitimate reasoning and in-flight tools;
- content-progress or transport-level guards for streams that stay open without useful progress;
- an attention notice plus explicit operator control when automatic recovery risks false positives; or
- another mechanism that preserves cleanup, abort/interrupt races, and fallback reachability.

Tests should cover long legitimate turns, in-flight tools, contentless deltas, abort/interrupt races, cleanup after recovery, and the case where fallback becomes reachable. Capturing raw SSE or equivalent transport chunks would help determine whether the observed stream is sustained by the provider or by client/parser behavior. That trigger question should remain separate from the confirmed liveness consequence.

## Evidence

Sanitized evidence attachment: https://gist.github.com/flora131/fd6818004aa46197219dc2fa8c493a24

## Version

Observed on Atomic `0.9.14-alpha.1`. Behavior and cited source were rechecked on 2026-08-16 at current main commit `66a10d1c858ebbd2aa9ea5e82b5dc3f88210c285`.

yangjj-iso:
## Summary

Fixes #2446. A non-terminating foreground in-process subagent model turn could block its parent indefinitely and prevent a configured fallback from ever being reached. The reporter observed **32749 consecutive whitespace-only `toolcall_delta` events**, `message_end` never observed, the foreground child pending, **fallback attempt count 0**, and the parent blocked **~6.5 hours**.

### Root cause

pi-agent-core's consumption loop (`for await (const event of response)`) has **no inter-event timeout and no `signal.aborted` check**. A stream that floods contentless deltas — or goes silent — never reaches a terminal event, so `prompt()` never settles. Fallback is **throw-driven**: it advances only when a throw during consumption is stamped `stopReason:"error"`. A stream that emits no terminal error throws nothing, so fallback is never consulted.

### Fix — a content-progress liveness guard at the stream layer

Wrap the `AssistantMessageEventStream` returned by `createAgentSession`'s streamFn factory with a guard that tracks time since the last **content-bearing** event. If none arrives within `streamStallMs`, the guard **throws** a retryable failure whose message matches the classifier's `stream ended before a terminal response event` pattern and normalizes to `provider_unavailable`. Same-model retry and configured `fallbackModels` then advance within the same `prompt()` call — or, with no fallback, `prompt()` resolves with an error and the runner returns control instead of hanging.

Key properties:

- **Classification stays pi-agent-core's job.** The guard only decides *when* to end the stream. An abort during a stall is stamped `"aborted"` → `cancelled` → **no** fallback; a stall while not aborted is `"error"` → fallbackable. The abort/interrupt race is handled for free.
- **Content-progress, not wall-clock.** The window measures time since the last content-bearing event (any non-empty text/thinking/tool-call delta, plus every structural and terminal event). Long reasoning passes and slow in-flight tools keep resetting it and are never affected. It bounds forward progress *within a single stream* — not socket idleness, not total turn/tool duration.
- **Throw, never return-done.** After the for-await loop, agent-loop calls `await response.result()`, which would hang forever without a terminal event — so on stall/abort the guarded iterator throws.
- **Inert by default.** The guard is off unless `streamStallMs > 0`. Only unattended in-process subagent sessions set it; **the interactive main session is byte-for-byte unchanged**.

### Config surface

Subagent control config gains `streamStallMs` (default **300000 ms**, `0` disables), resolved from per-run `control.streamStallMs` or global config and threaded through the child spec into the SDK session. With defaults it trips *after* the observational needs-attention (60000 ms) and long-running (240000 ms) notices, which surface a stalled child but never terminate it.

## Files

- **New** `packages/coding-agent/src/core/stream-liveness-guard.ts` — the guard (`isProgressEvent`, `guardStreamLiveness`, `GuardedAssistantMessageEventStream`).
- `packages/coding-agent/src/core/sdk-types.ts` + `sdk.ts` — `streamStallMs` option; all three streamFn return paths wrapped.
- `packages/subagents/src/{shared/types-results,extension/schemas,runs/shared/subagent-control,runs/inprocess/runner,runs/foreground/inprocess-run-sync}.ts` — config + spec threading (`parseStallMs` admits the `0` disable sentinel that `parsePositiveInt` would reject).
- `packages/coding-agent/docs/subagents.md` + both `packages/*/CHANGELOG.md`.

## Tests

Maps to the issue's required cases — long legitimate turns, in-flight tools, contentless deltas, abort/interrupt races, cleanup after recovery, and fallback becoming reachable.

- **`test/unit/stream-liveness-guard.test.ts`** (deterministic injected clock/timer): long turn spanning many windows never trips; non-empty tool-call deltas keep it alive; whitespace flood trips bounded with one stall; silent hang trips via the timer with no unhandled rejection; already-aborted and abort-during-stall both outrank the stall; cleanup removes the listener and cancels timers; terminal `done` completes; terminalless close ends retryably; a 16-case `isProgressEvent` table; and a classification bridge asserting the real stall error normalizes to a same-model-retryable, fallbackable `provider_unavailable`.
- **`test/integration/subagent-stream-stall-fallback.test.ts`**: drives the **real** `createAgentSession` streamFn factory — a stalled provider stream ends with a fallbackable failure the classifier advances on, and an abort during a stall surfaces as an abort (not a fallbackable failure).

## Verification

`npm run check` (biome + `tsc --noEmit` + coding-agent `tsgo -p tsconfig.build.json --noEmit` + shrinkwrap) is green on the rebased tree. New unit (28) and integration (2) tests pass. Remaining full-suite failures on my Windows machine are the pre-existing symlink-EPERM / load-sensitive / tar-C: environmental cases documented in AGENTS.md, unrelated to this change.

---
<a href="https://app.blacksmith.sh/bastani-inc/codesmith/atomic/pr/2452"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789488736&installation_model_id=10562&pr_number=2452&repository=bastani-inc%2Fatomic&return_to=https%3A%2F%2Fgithub.com%2Fbastani-inc%2Fatomic%2Fpull%2F2452&signature=aed1b91632cd11f459ebfa0dd8e3534262322da52bb88a7a87302b927ff3b966"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

<h3>Greptile Summary</h3>

This change adds a configurable content-progress timeout for unattended in-process subagent streams and connects stalled streams to retry and model fallback handling. Direct final-result consumers in compaction and summarization do not enter the guarded iterator, so a provider that never produces a terminal result can still leave those flows waiting indefinitely. The integration test should also use the repository runtime filesystem helper to retain cross-runtime filesystem behavior.

Do not merge until direct final-result consumers receive the same stall protection as iterated streams.

A nonterminating-stream reproduction showed that direct result calls did not pull the guarded iterator or invoke its stall callback, while the focused existing guard tests passed without covering that path.

**Files Needing Attention:** packages/coding-agent/src/core/stream-liveness-guard.ts needs a direct-result liveness policy; test/integration/subagent-stream-stall-fallback.test.ts needs the runtime filesystem helper.

**What T-Rex did**
- Ran a focused reproduction of the raw and guarded direct-result calls to address the posted P1 finding.
- Compared the raw direct-result reproduction output, the guarded direct-result reproduction output, and the existing stream liveness guard unit-suite results to validate behavior across paths.
- Validated contract behavior by wrapping the same stream with guardStreamLiveness(..., { stallMs: 10 }) and observed that it still times out when no iterator is pulled.
- Summarized the plain-language takeaway that compaction/summarization paths requesting only the final result can wait forever instead of yielding the configured retryable liveness failure.

<a href="https://app.greptile.com/trex/runs/19110602/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

1. General comment 

   <a href="#">P1</a> **Direct result bypasses the stream liveness guard**

   - **Bug**
     - `GuardedAssistantMessageEventStream.result()` delegates to the underlying nonterminating provider stream. The liveness timeout exists only inside `[Symbol.asyncIterator]`, so a caller that uses `await stream.result()` without iterating never starts the timeout and can hang indefinitely. This is reachable from the compaction session-summary, branch-summary, and range-planner request paths.
   - **Cause**
     - `packages/coding-agent/src/core/stream-liveness-guard.ts:137-139` returns `this.source.result()` directly, while timer setup and the failure race occur only after the guarded async iterator is consumed at lines 141-227.
   - **Fix**
     - Make `result()` observe the same liveness policy as iteration—for example, have it consume the guarded iterator through terminal completion and return/throw the corresponding final result, with safe single-consumer coordination—or change every direct-result compaction/summarization caller to consume the guarded stream before awaiting its result. Add focused tests for silent and whitespace-flooding sources invoked through direct `result()`.

</details>

`````markdown
### Issue 1
packages/coding-agent/src/core/stream-liveness-guard.ts:137-139
**Direct result calls bypass the liveness guard**

`result()` delegates directly to the underlying provider stream, while the stall timer is only created when the guarded async iterator is consumed. Compaction and summarization paths that await `stream.result()` without iteration therefore never arm the configured timeout; a provider stream that never produces a terminal result leaves the child and its parent waiting indefinitely rather than surfacing a retryable failure for retry or model fallback.

### Issue 2
test/integration/subagent-stream-stall-fallback.test.ts:2
**Test bypasses runtime filesystem helpers**

This root integration test imports and invokes `node:fs` operations directly instead of using `test/helpers/runtime.ts`, bypassing the repository's centralized cross-runtime filesystem behavior and platform conventions.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(subagents): bound a non-terminating ..."](https://github.com/bastani-inc/atomic/commit/e98b62e6b408a6bdc34e0a2e11e772a1d6ecf392) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53781377)</sub>

**Context used:**


flora131:
Thanks for taking the time to investigate #2446 and build out the guard and tests.

After reviewing this again as a team, we’ve decided not to add an automatic content-progress timeout. A stream that appears stuck can be indistinguishable from valid slow reasoning or tool work, so any threshold remains a heuristic that may terminate legitimate turns. For now, we prefer the existing attention/operator controls and transport-level deadlines rather than adding another automatic termination policy.

This is a product decision, not a judgment on the effort or quality of the contribution. We’re going to close #2446 as not planned and close this PR with it.

For future substantial issue work, please follow our [contribution guide](https://github.com/bastani-inc/atomic/blob/main/CONTRIBUTING.md#claiming-an-issue): comment with your intended approach and wait for assignment or explicit maintainer approval before starting. #2446 was unassigned, and that coordination step would have helped us settle this product decision before you invested the time. Thanks again for picking it up and for understanding.

PR #2450, opened 2026-08-16
feat(workflows): add per-attempt ctx.tool timeouts

flora131:
## Summary

Add an optional per-attempt `timeoutMs` deadline to `ctx.tool` so a callback that never settles cannot leave its workflow node running forever. Timeout expiry aborts the attempt signal, records a distinct timeout error with the elapsed bound, and then follows the existing retry and failure-mode rules.

## Changes

- Add `timeoutMs` to the typed `WorkflowToolOptions` API and reject non-positive or non-finite values before callback execution.
- Race each timed attempt against its deadline and run cancellation, with a fresh `AbortSignal` and deadline for every retry.
- Preserve timeout evidence in typed `failureMode: "return"` outcomes, thrown failures, durable checkpoints, tool-node state, and repair-stage handoffs.
- Release a tool node at its deadline even if its callback ignores the signal, while keeping untimed calls on the existing path.
- Handle cancellation/timeout races, deadline cleanup, cached replay, and deadlines larger than the host timer limit.
- Document why retries do not bound hangs and require child-process and network callbacks to set `timeoutMs` and forward `signal`.
- Add the feature to the workflows package's `[Unreleased]` changelog.

## Tests

Final workflow verification passed:

- `npm run check`
- `npm run test:unit` — 6,419 passed, 2 skipped
- `npm run test:integration` — 494 passed, 1 skipped
- `npm run test:ci-contracts` — 45 passed
- `npm --workspace=@bastani/atomic run test` — 3,865 passed, 39 skipped
- `git diff --check HEAD^..HEAD`

## Breaking Changes

None. Existing `ctx.tool` calls that omit `timeoutMs` keep their current behavior.

Closes #2424

---
[...]

PR #2454, opened 2026-08-16
feat(workflows): inspect ctx.tool graph nodes

### LINKED ISSUE #2427: Workflow tool nodes are not inspectable: clicking a ctx.tool node should show its args, calling code, and output
## What happened?

`ctx.tool` nodes render as cards in the workflow graph view, but they are not inspectable. Clicking one (or focusing it and pressing Enter) does nothing: there is no detail view, no way to see what the tool was called with, and no way to see what it returned. The card body is the entire available surface — `stage.error ?? stage.result ?? "durable tool"` truncated to card width (`packages/workflows/src/tui/node-card.ts:276-278`).

This is a hard opt-out in the snapshot model, not a missing feature in the render path: `ToolNodeSnapshot` declares `readonly attachable: false` (`packages/workflows/src/shared/store-types.ts:38`), so the attach flow that Enter and direct node clicks trigger (`_activateFocusedNode` → `onStageAttach`, `packages/workflows/src/tui/graph-view-input.ts:243-261`) skips tool nodes entirely. `ToolNodeSnapshot` also retains only `argsHash` and `resultSummary` — the full args and full result are not on the snapshot, so even a renderer that wanted them has nothing to show.

Compare the two adjacent experiences that already work:

- A **model stage** is attachable: Enter/click opens the stage-chat view with its transcript, so you can read exactly what the stage did.
- An **agent tool call** inside that transcript is inspectable end-to-end — call and result.
- A `ctx.ui.custom` (**context.ui**) mount gets a first-class clickable custom UI in the stage chat (`packages/workflows/src/tui/stage-chat-view-custom-ui.ts`).

The `ctx.tool` node — the primitive the docs push for every workflow-owned external check and side effect — is the only node type with no click-through. During a run this means a failed or long-running gate is a card that says `failed` and nothing else. Debugging requires dropping to `workflow status`/durable inspection outside the TUI.

## Steps to reproduce

1. Author a workflow with a tool node:

```ts
await ctx.tool("checks", { branch: "feat/x" }, async () => {
	const r = spawnSyncCollect("gh", ["pr", "checks", "feat/x"]);
	return { stdout: r.stdout, exitCode: r.status };
});
```

2. Run it, open the graph view, click the `checks` node (or focus it and press Enter).
3. Nothing opens. The card shows `failed`/`completed` plus a one-line summary. There is no path to the args, the output, or the error body.

## Expected behavior

Clicking a workflow tool node attaches a detail view for that node, the same way stage nodes attach the stage-chat view. Concretely:

1. **Full tool call input** — the node's `name` and the complete `args` object it was invoked with. Today only `argsHash` survives on the snapshot; the view needs the full payload (see plumbing notes below).
2. **Calling code** — the source of the `fn` callback passed to `ctx.tool`, so the reader can see *what* the node does, not just what it was named. The engine has the function in hand at invocation time; capturing `fn.toString()` at registration is the cheap path.
3. **Output** — the full serialized result once the call settles (`completed`), the full error for `failed`, plus the existing status, timing (`startedAt`/`endedAt`/`durationMs`), and the `cached`/`replayed` markers. `resultSummary` is a preview, not a substitute.
4. **Card body drops the output** — once the detail view exists, the card should stop trying to be a viewport onto the result. Today the body is `stage.error ?? stage.result ?? "durable tool"` (`packages/workflows/src/tui/node-card.ts:276-278`), so a settled node renders a serialized payload chopped at card width:

```
╭──── gate-L13-3-1 ────╮
│{"ok":true,"value":{"…│
│     ✓ completed      │
│        1 dep         │
╰──────────────────────╯
```

That line carries no information — it is a JSON prefix, not a summary — and the same is true of a truncated error. The body should be the constant `durable tool` in every state; status, timing, and dependency count already occupy their own rows, and the full args/result/error live one click away. Slimming the card is part of this change, not a follow-up: the card is only bloated today because it is the *only* surface, and adding the detail view removes that excuse.

## Plumbing notes (what has to move)

- `ToolNodeSnapshot.attachable: false` (`packages/workflows/src/shared/store-types.ts:38`) — flip to a real attach target. Whatever consumes `attachable` (attach-pane eligibility, stage-attach routing) needs to accept `kind: "tool"` targets with a `tool:<argsHash>` stage-id shape, which `workflow({ action: "interrupt", stageId })` already accepts for control.
- Full args/result are durably checkpointed by the DBOS backend (`packages/workflows/src/durable/tool-primitive.ts`); the snapshot projection drops them. Either widen `ToolNodeSnapshot` or expose a targeted-inspection fetch the detail view calls on open — the latter keeps snapshots lean for large payloads.
- `node-card.ts` cards already render for `nodeKind === "tool"`; only the input path and the attach shell change. The body expression at `node-card.ts:276-278` collapses to the literal `"durable tool"` — the `stage.error ?? stage.result` fallbacks go away with it. Check whether `result`/`error` on the tool snapshot still has another consumer before narrowing the projection.
- Note that the card's `stage.result` is not a snapshot field: `ToolNodeSnapshot` carries `resultSummary`, and `expanded-workflow-graph.ts:199` and `:352` (first projection and refresh path) rename it to `result` on the view model. Both sites are card-only, but the underlying `resultSummary` is **not** dead — `workflow-status-summary.ts:188` puts it in the `workflow({ action: "status" })` payload, which is the fallback inspection path this issue is trying to make unnecessary. Do not delete the field while slimming the card.
- `graph-store-snapshot.ts:105-118` strips `result` from agent stages when compacting, with a comment saying graph cards do not render it and "durable tool cards use ToolNodeSnapshot.resultSummary instead." That comment goes stale the moment the card body becomes a constant — after this change tool cards render neither. Update it in the same pass.

This also closes an asymmetry with `workflow send`/`interrupt`, which already address tool nodes by id — control can reach the node but inspection cannot.

flora131:
## Summary

`ctx.tool` nodes render as cards in the workflow graph view but could not be inspected: clicking one or pressing Enter did nothing, so there was no way to see what a durable tool node was called with or what it returned. The card body was the entire surface, showing `stage.error ?? stage.result ?? "durable tool"` truncated to card width, and `ToolNodeSnapshot` declares `readonly attachable: false`, so the attach flow behind Enter skipped these nodes.

This adds a **read-only** detail view for tool nodes. Enter on a focused node, a direct click, or selection from the `/` switcher opens a snapshot-backed panel showing the tool name, the exact args, the retained result or error, the callback source, status, and timing. `ctrl+x` or `escape` returns to the graph.

Closes #2427.

> Stacked on #2450 (`feat/2424-ctx-tool-timeout`), which is now merged. The base branch still exists on the remote, so this PR is targeted at it to keep the diff to just this change. Retarget to `main` once the base branch is deleted or after a rebase.

## What changed

**Inspection data on the snapshot.** `ToolNodeSnapshot` gains optional `args`, `result`, `durationMs`, and `source`. `source` is captured from `fn.toString()` at call registration — capture never re-executes the callback and never reads a file. `readonly attachable: false` is unchanged and now carries a comment saying why.

**The detail view.** `src/tui/tool-detail.ts` renders the panel; `graph-view-input.ts` routes activation for tool nodes to it instead of the stage attach path. The view is scrollable with `↑`/`↓`, `PageUp`/`PageDown`, `Home`/`End`, the wheel, and the scrollbar, so a long payload is readable without a mouse.

**Bounded, hostile-input-safe payload handling.** `src/shared/tool-payload-bounds.ts` replaces `structuredClone`/`JSON.stringify` on these paths. Both throw on hostile input, and both cost what the payload costs rather than what the cap allows. The new walker charges every branch against a single 16,384-character cap and makes every bound explicit in its output: truncation appends `… [truncated]`, a back-reference becomes `<cycle>`, an unreadable property becomes `<unreadable>`. A cyclic payload, a throwing `toJSON`, a throwing getter, or a result holding a function, promise, proxy, symbol, or weakmap renders a placeholder instead of crashing the graph or the session list. Nesting is bounded at depth 256 so a pathological payload cannot exhaust the stack before the character cap stops it.

**Card contract.** The tool card body becomes the constant `"durable tool"`. The card is a fixed-size graph node, not a result preview; the detail view owns args, result, error, and timing. Status, meta, and dependency rows are untouched.

**Durable state is unaffected.** The checkpoint keeps the exact output, so cached replay and `argsHash` identity are unchanged. The cap governs the live run snapshot and the display, not the durable record.

## Constraints held

- Tool nodes never become attachable chat stages. `attachable: false` stands; inspection is not an attach.
- The view exposes no interrupt, resume, or steering affordance.
- Nothing is rendered that the snapshot does not already carry.
- The view cannot crash on a big or unusual payload, and truncation is always marked.
- Stage-node attach behavior is unchanged; stages without an attach callback keep their historical no-op.

## Tests

`test/unit/workflow-tool-detail.test.ts` (1,074 lines) covers the four required cases, each failing without the fix: activation opens the view; the view renders name, args, and result; a failed node renders its error; an oversized payload truncates with a marker rather than throwing. It also covers the hostile-payload shapes, keyboard scrolling, and narrow-frame rendering down to 40 columns.

Review found and closed a retention defect during the run: the walker charged nothing for `null` or for values JSON omits from objects (`undefined`, functions, symbols, array holes), so the cap never tripped on those shapes. An array of nulls was retained whole and re-walked on every graph snapshot, and a sparse `new Array(20_000_000)` — free to create and a legal tool return — was densified into twenty million retained elements. Every branch now charges. Measured after the fix: 5,000,000 nulls retain 3,278 characters; three consecutive `graphSnapshot()` calls run 1.4/0.6/0.6 ms where they previously took 1.1–3.5 s each.

## Known non-blocking issue

All three reviewers independently found the same P3: the object truncation marker is written under the literal key `"…"`, so an args record that itself owns a key named `"…"` has that author value replaced when the record hits the cap. Display-only, no durable effect, and not worth widening this diff for.

## Checks

| Command | Result |
|---|---|
| `npm run check` | pass |
| `npm run test:unit` | pass |
| `npm run test:integration` | pass |
| `npm run test:ci-contracts` | pass |
| `npm --workspace=@bastani/atomic run test` | pass |

CHANGELOG entry is under `[Unreleased]` in `packages/workflows/CHANGELOG.md`.

---
<a href="https://app.blacksmith.sh/bastani-inc/codesmith/atomic/pr/2454"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789496894&installation_model_id=10562&pr_number=2454&repository=bastani-inc%2Fatomic&return_to=https%3A%2F%2Fgithub.com%2Fbastani-inc%2Fatomic%2Fpull%2F2454&signature=b0ae5e2736fd906bbb15bb4085206ff6752818e74fbb71ec83815963b097fd07"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

<h3>Greptile Summary</h3>

This change adds durable inspection details for `ctx.tool` executions, including arguments, results, callback source, timing, and a terminal detail panel. Verification reproduced a responsiveness and memory-risk issue in the payload helper: extremely wide objects are fully enumerated before truncation can stop traversal, so inspecting a tool payload can still perform work proportional to the full object.

The change should not merge until payload key discovery is bounded as well as payload output.

A focused executable reproduction against the changed helper confirmed that wide objects are fully enumerated before truncation takes effect. The linked type-boundary requirement also remains unmet.

**Files Needing Attention:** packages/workflows/src/shared/tool-payload-bounds.ts needs bounded key discovery for all shared payload-walking paths and clearer boundary types.

**What T-Rex did**
- Ran the enumerable-key experiment using the reproduction source to count ownKeys, descriptor, and value reads.
- Captured a baseline showing the complete enumeration of keys.
- Captured a bounded scenario where keys are truncated while descriptor enumeration stays around 200,000, showing the value limit does not bound enumeration.
- Two finding-proof entries were produced for the P1 finding and linked to the corresponding review comments.

<a href="https://app.greptile.com/trex/runs/19118482/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

1. General comment 

   <a href="#">P1</a> **Payload cap is bypassed while collecting enumerable object keys**

   - **Bug**
     - `boundedToolPayload` and `boundedToolPayloadText` call `safeOwnKeys`, which executes `Object.keys(value)` and creates a complete key array before their loops check remaining budget. An object/proxy with 200,000 enumerable keys caused 200,000 descriptor checks even with a payload limit of 32, while only 5 or 3 property values were consumed before truncation.
   - **Cause**
     - `safeOwnKeys` returns the eagerly materialized result of `Object.keys` at lines 60-64; the cap checks at lines 154 and 313 run only after that full enumeration has completed.
   - **Fix**
     - Replace eager `Object.keys` collection with a traversal mechanism whose key discovery is itself bounded, or impose a key-count/work bound before materializing a complete key list. Apply it to the shared helper/call sites so both retention and text serialization honor the advertised work bound.

</details>

`````markdown
### Issue 1
packages/workflows/src/shared/tool-payload-bounds.ts:60-66
**Payload cap is bypassed while collecting enumerable object keys**

`safeOwnKeys` eagerly calls `Object.keys(value)`, which materializes every enumerable key before either bounded walker reaches its remaining-budget check. A 200,000-key proxy with a limit of 32 still performed 200,000 property-descriptor checks, although the retention and display paths consumed only 5 and 3 values before truncating. This makes tool completion and snapshot rendering perform payload-sized work and allocate a payload-sized key array despite the configured limit. Use key discovery that is itself bounded, or impose a key-count/work limit before collecting keys, for all callers of this shared helper.

### Issue 2
packages/workflows/src/shared/tool-payload-bounds.ts:36
**Ambiguous payload boundary types**

The new payload helpers accept and return `unknown` and rely on repeated casts, contrary to the repository requirement to use specific types; this obscures which hostile values each boundary supports and moves safety guarantees out of the type system.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(workflows): inspect ctx.tool graph ..."](https://github.com/bastani-inc/atomic/commit/2ad08c1c737452c22a90c67d015bd3b32b9896d5) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53797001)</sub>

**Context used:**

### CODE REVIEW

**Comment by flora131** [comment]
# End-to-end verdict: **PASS**

PR #2454 was exercised in a real tmux-hosted Atomic session built from this worktree. A successful tool, a failed tool, and a large-result tool all opened in the new read-only detail view with name, args, result/error, source, timing, and bounded rendering. The large result displayed the explicit `… [truncated]` marker without freezing or crashing.

## Environment

- Worktree: `/Users/tonystark/Documents/projects/atomic-ctx-tool-inspect`
- Branch: `feat/2427-tool-node-inspection`
- HEAD: `2ad08c1c737452c22a90c67d015bd3b32b9896d5`
- Atomic binary executed: `/Users/tonystark/Documents/projects/atomic-ctx-tool-inspect/packages/coding-agent/dist/cli.js`
- Atomic version: `0.0.0`
- Node: `v26.7.0`
- tmux: `tmux 3.7b`
- Workflow run: `dc3f240e-af9d-4e2b-9750-d22c8faae3a6`

## Exact commands and interaction

```sh
npm --workspace=@bastani/atomic run build

tmux new-session -d -s evidence2454 -x 120 -y 40 \
  'cd /Users/tonystark/Documents/projects/atomic-ctx-tool-inspect && \
   exec node packages/coding-agent/dist/cli.js --no-session'

# In the Atomic prompt:
/workflow evidence-tool-inspection
/workflow connect dc3f240e-af9d-4e2b-9750-d22c8faae3a6
# Focused each graph card in order; pressed Enter; captured each detail pane.
# Pressed Escape to return to graph, Down to focus the next card, Enter to inspect.
# Pressed End in the large-result detail to show its bounded tail.
```

The temporary workflow created three tracked nodes:

- `evidence-success`: `{kind: "success", answer: "small-result"}` → successful result.
- `evidence-failure`: `{kind: "failure", expected: true}` → intentional failure under `failureMode: "return"`.
- `evidence-large`: `{kind: "large", bytes: 30000}` → 30,000-character payload.

## Graph pane before inspection

```text

 ╭──────────────╮                                                                                                       
 │ ORCHESTRATOR │  evidence-tool-inspection                                                                   ✓ 2  ✗ 1  
 ╰──────────────╯                                                                                                       
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                              ╭── evidence-success ──╮                                                  
                                              │     durable tool     │                                                  
                                              │     ✓ completed      │                                                  
                                              │         root         │                                                  
                                              ╰──────────────────────╯                                                  
                                                          │                                                             
                                                          │                                                             
                                                          │                                                             
                                              ╭── evidence-failure ──╮                                                  
                                              │     durable tool     │                                                  
                                              │       ✗ failed       │                                                  
                                              │        1 dep         │                                                  
                                              ╰──────────────────────╯                                                  
                                                          │                                                             
                                                          │                                                             
                                                          │                                                             
                                              ╭─── evidence-large ───╮                                                  
                                              │     durable tool     │                                                  
                                              │     ✓ completed      │                                                  
                                              │        1 dep         │                                                  
                                              ╰──────────────────────╯                                                  
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
 ╭───────╮                                                                                                              
 │ GRAPH │                            ctrl+x return to main chat  ·  ↵ open tool detail  ·  ↑↓←→ navigate  ·  / stages  
 ╰───────╯```

The graph cards are explicitly `durable tool`, and the footer advertises `↵ open tool detail` rather than stage-chat attach.

## Successful tool detail

```text

 ╭──────────────╮                                                                                                       
 │ ORCHESTRATOR │  evidence-tool-inspection                                                                   ✓ 2  ✗ 1  
 ╰──────────────╯                                                                                                       
            ╭ TOOL evidence-success ───────────────────────────────────────────────────────────────────────╮            
            │NAME      "evidence-success"                                                                  │            
            │STATUS    completed                                                                           │            
            │ARGS      {"kind":"success","answer":"small-result"}                                          │            
            │RESULT    {"status":"ok","message":"small-result"}                                            │            
            │SOURCE    async () => ({ status: "ok", message: "small-result" })                             │            
            │TIMING    startedAt=1786905472107 · endedAt=1786905472107 · duration=0ms (0s)                 │            
            │MARKERS   —                                                                                   │            
            ╰──────────────────────────────────────────────────────────────────────────────────────────────╯            
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
 ╭───────╮                                                                                                              
 │ GRAPH │                                                              ctrl+x return to graph  ·  ↑↓ pgup/pgdn scroll  
 ╰───────╯```

## Failed tool detail

```text

 ╭──────────────╮                                                                                                       
 │ ORCHESTRATOR │  evidence-tool-inspection                                                                   ✓ 2  ✗ 1  
 ╰──────────────╯                                                                                                       
            ╭ TOOL evidence-failure ───────────────────────────────────────────────────────────────────────╮            
            │NAME      "evidence-failure"                                                                  │            
            │STATUS    failed                                                                              │            
            │ARGS      {"kind":"failure","expected":true}                                                  │            
            │RESULT    {"ok":false,"error":{"name":"Error","message":"intentional evidence failure"},"attem│            
            │          pts":1,"cached":false}                                                              │            
            │ERROR     "intentional evidence failure"                                                      │            
            │SOURCE    async () => {                                                                       │            
            │                  throw new Error("intentional evidence failure");                            │            
            │                }                                                                             │            
            │TIMING    startedAt=1786905472110 · endedAt=1786905472111 · duration=1ms (0s)                 │            
            │MARKERS   —                                                                                   │            
            ╰──────────────────────────────────────────────────────────────────────────────────────────────╯            
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
                                                                                                                        
 ╭───────╮                                                                                                              
 │ GRAPH │                                                              ctrl+x return to graph  ·  ↑↓ pgup/pgdn scroll  
 ╰───────╯```

## Large payload detail and truncation

The tail after pressing `End` contains the renderer's explicit marker:

```text

 ╭──────────────╮                                                                                                       
 │ ORCHESTRATOR │  evidence-tool-inspection                                                                   ✓ 2  ✗ 1  
 ╰──────────────╯                                                                                                       
           │          bcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcde│             
           │          f0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef012│             
           │          3456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456│             
           │          789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789a│             
           │          bcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcde│             
           │          f0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef012│             
           │          3456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456│             
           │          789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789a│             
           │          bcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcde│             
           │          f0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef012│             
           │          3456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456│             
           │          789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789a│             
           │          bcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcde│             
           │          f0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef012│             
           │          3456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456│             
           │          789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789a│             
           │          bcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcde│             
           │          f0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef012│             
           │          3456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456│             
           │          789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789a│             
           │          bcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcde│             
           │          f0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef012│             
           │          3456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456│             
           │          789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef01… [trunca│             
           │          ted]                                                                                │             
           │SOURCE    async () => ({                                                                      │             
           │                  payload: "LARGE-PAYLOAD-BEGIN\n" + "0123456789abcdef".repeat(2_000) + "\nLAR│             
           │          GE-PAYLOAD-END"                                                                     │             
           │                })                                                                            │             
           │TIMING    startedAt=1786905472115 · endedAt=1786905472116 · duration=1ms (0s)                 │             
           │MARKERS   —                                                                                   │             
           ╰──────────────────────────────────────────────────────────────────────────────────────────────╯             
 ╭───────╮                                                                                                              
 │ GRAPH │                                                              ctrl+x return to graph  ·  ↑↓ pgup/pgdn scroll  
 ╰───────╯```

## Read-only / non-attachable verification

Inspection was verified as read-only in the live pane: after Enter, the detail footer contained only `ctrl+x return to graph · ↑↓ pgup/pgdn scroll`; there was no chat composer, attach affordance, steering input, or tool-control action. Enter was consumed while in the detail pane, and Escape returned to the graph. The source contract independently confirms the invariant:

```text
packages/workflows/src/shared/store-types.ts:57: readonly attachable: false;
packages/workflows/src/tui/graph-view-constants.ts:15: { key: "ctrl+x", label: "return to graph" },
```

Thus the tool node was inspected without creating or attaching a stage session.

The temporary workflow file was deleted afterward and `git status --porcelain` is clean.

PR #2457, opened 2026-08-16
fix(workflows): open ctx.tool nodes as a collapsed tool message block

### LINKED ISSUE #2427: Workflow tool nodes are not inspectable: clicking a ctx.tool node should show its args, calling code, and output

flora131:
## Summary

Follow-up to merged PR #2454 (`feat(workflows): inspect ctx.tool graph nodes`). It replaces that PR's rounded standalone detail panel with the same message-block shape agent tool calls already use, and closes both unresolved Greptile threads left on #2454:

- [#discussion_r3792534342](https://github.com/bastani-inc/atomic/pull/2454#discussion_r3792534342) — **P1**: the payload cap was bypassed while collecting enumerable object keys.
- [#discussion_r3792534347](https://github.com/bastani-inc/atomic/pull/2454#discussion_r3792534347) — **P2**: ambiguous `unknown` payload boundary types and repeated casts.

It also fixes the truncation-key collision that #2454 shipped as a known non-blocking display defect.

## The UI change

`ctx.tool` inspection is now one read-only, agent-chat-style tool message block opened from the graph — not a dropdown, accordion, expanded graph card, rounded standalone card, chat attachment, or editable transcript.

- Enter on a focused node, a direct click, or switcher selection opens the block **collapsed**, showing the tool name/call and a compact bounded result or error preview.
- The configured `app.tools.expand` action (`ctrl+o` by default) toggles the same block into full bounded args, result or error, callback source, timing, and cached/replayed markers. A raw `\x0f` fallback runs **only** where no host keybindings object exists, so a user remap wins and an explicitly unbound action stays unbound.
- The frame is flat with a one-cell inset (`renderRoundedBoxLines` is no longer imported), matching `packages/coding-agent/src/modes/interactive/components/tool-execution.ts` without a circular import from `@bastani/atomic` into the workflows TUI.
- Escape / `ctrl+x` return, keyboard and wheel scrolling, scrollbar, narrow-terminal safety, status styling, `attachable: false`, and the absence of attach/steer/interrupt/resume are all preserved.

Collapsed multi-line errors keep the block inset on every continuation row, are capped at `COLLAPSED_RESULT_LINES`, and say what was dropped with `... (N more lines, ctrl+o Expand)` rather than letting a single-row fit silently swallow them.

## Thread 1 — bounded key discovery (P1)

`safeOwnKeys` no longer returns an eagerly materialized `Object.keys` array. It is a generator-backed scan that calls the `[[OwnPropertyKeys]]` trap **once** and then bounds every operation this module controls:

- at most `TOOL_PAYLOAD_MAX_KEYS` (2,048) candidate keys have a descriptor read;
- only candidates passing the enumerable-string check have a value read;
- budget is charged **before** filtering, deliberately including symbol and non-enumerable candidates, so a hostile key list cannot make filtering work unbounded.

Both the retention path (`boundedToolPayload`) and the text path (`boundedToolPayloadText`) share that budget. Own enumerable string-key order is preserved.

**Honest limit, not a claimed one:** the engine's own `Reflect.ownKeys` materialization of the trap result is outside JavaScript's control. That is documented in the module header and in the docs rather than asserted away. The budget is one total-work bound across the recursive walk, so a very wide member can consume budget its siblings would have used — a deliberate trade, documented at the constant.

`test/unit/workflow-tool-payload-bounds.test.ts` proves the bound with a hostile proxy that counts `ownKeys`, `getOwnPropertyDescriptor`, and `get` traps on both paths.

## Thread 2 — precise boundary types (P2)

`unknown` and the cast chains are gone from the payload boundary. The module now exports `ToolPayloadValue`, `ToolPayloadObject`, `ToolPayloadArray`, `ToolPayloadCallable`, and `ToolPayloadRetainedValue`, with guards `isToolPayloadObject`, `isToolPayloadArray`, `isToolPayloadObjectLike`, and `isToolPayloadCallable`. Property reads go through a `PropertyRead | PropertyReadFailure` result rather than a cast plus try/catch at each site. The guards drive the real walkers rather than sitting unused; `hasCallableToolPayloadToJSON` and its dead type export were removed. Runtime containment is unchanged: `<cycle>`, `<unserializable>`, `<unreadable>` placeholders all still apply.

## Truncation-key collision

`addTruncationMarker` now probes for a free key (`…`, then `…1`, `…2`, …) instead of writing unconditionally under the literal `"…"`. An author key named `"…"` is no longer overwritten. The bound is not weakened — one marker is still added, once.

## Unchanged by construction

Durable exact output, `argsHash` identity over raw args, replay behavior, checkpoint compatibility, `ToolNodeSnapshot.resultSummary`, and workflow discovery/runtime semantics outside inspection.

## Tests

TDD, red before green. `test/unit/workflow-tool-detail.test.ts` and `test/unit/workflow-tool-payload-bounds.test.ts` were updated rather than left asserting the old panel contract — the border assertions became `assert.doesNotMatch` checks.

New and rewritten cases cover: collapsed message-block parity with the host grammar; `ctrl+o` expand/collapse through configured keybindings and through the no-keybindings fallback; full fields when expanded; failed, cached, and replayed nodes; truncation markers; hostile accessors, `toJSON`, and cycles; bounded wide-key work measured by trap counters; narrow frames down to 40 columns; scrolling; and the absence of any attach or control action.

The multi-line regression case asserts the exact wide-width row array, the two-space continuation inset, that markers on lines 2 and 3 survive at width 60, that no row exceeds 60 columns, that a 200-line error stays under 12 rows, and that the last row carries the remaining-lines marker.

## Checks

| Command | Result |
|---|---|
| `npm run check` | pass |
| focused workflow tool tests (38) | pass |
| `npm run test:unit` | pass (6,492 passed / 2 skipped) |
| `npm run test:integration` | pass (495 / 1 skipped) |
| `npm run test:ci-contracts` | pass (45) |
| `npm --workspace=@bastani/atomic run test` | pass (3,891 / 39 skipped) |
| `build:binary` | pass |

`test:unit`, `test:integration`, and `test:ci-contracts` also passed again in the pre-push hook for this branch.

## Live probe

The local Atomic CLI was built and a temporary workflow with successful, failed, and oversized tool nodes was driven under tmux at commit `3c90c45c20`.

Wide frame, collapsed, four-line error — every line present and inset:

```
 ✗ e2e-failure {"phase":"failure"}
   ✗ AssertionError: expected 1 to equal 2
     at foo (/x/y.ts:3:1)
     at bar (/x/y.ts:9:2)
   RR-LAST-ERROR-LINE
   (ctrl+o to expand)
```

Same node after `ctrl+o`:

```
 ✗ e2e-failure {"phase":"failure"}
 ARGS      {"phase":"failure"}
 ERROR     AssertionError: expected 1 to equal 2
             at foo (/x/y.ts:3:1)
             at bar (/x/y.ts:9:2)
           RR-LAST-ERROR-LINE
 SOURCE    async () => {
                   throw new Error("AssertionError: expected 1 to equal 2\n  at foo …
                 }
 TIMING    startedAt=1786913312370 · endedAt=1786913312371 · duration=1ms (0s)
 MARKERS   —
   (ctrl+o to collapse)
```

A second `ctrl+o` returned the collapsed frame byte-for-byte. At 60 columns the same error kept all four lines, and the oversized node collapsed to `✓ {"payload":"xxx…` with its truncation marker. Grepping the captures for `attach|steer|interrupt|resume|composer` returned nothing. Temporary fixtures were deleted.

## Docs and changelog

`packages/coding-agent/docs/workflows.md` and `packages/workflows/README.md` describe the collapsed-by-default message block, the `app.tools.expand` toggle, and the honest key-discovery bound. The existing `[Unreleased]` entry in `packages/workflows/CHANGELOG.md` was revised in place — no new released section touched.

## Known non-blocking

Reviewers left two P3 observations, both display-only and neither blocking:

1. The collapsed error's `N more lines` count is derived from the 320-character preview, so a very long error under-reports how many lines expansion would add.
2. An own **enumerable** `__proto__` key is dropped from the retained clone rather than carried as data.

Refs #2454. Closes #2427.

---
<a href="https://app.blacksmith.sh/bastani-inc/codesmith/atomic/pr/2457"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789507286&installation_model_id=10562&pr_number=2457&repository=bastani-inc%2Fatomic&return_to=https%3A%2F%2Fgithub.com%2Fbastani-inc%2Fatomic%2Fpull%2F2457&signature=9981edb28eb7665467bda77b514325cdb6101e289c8166c48fcc2216b4263514"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

<h3>Greptile Summary</h3>

The workflow graph view now renders tool output as expandable inline message blocks and bounds payload processing during rendering.

No blocking failure remains.

No accepted P0 or P1 finding remains, so the score is 5.

<sub>Reviews (3): Last reviewed commit: ["fix(workflows): type toolExpandKey again..."](https://github.com/bastani-inc/atomic/commit/c16aed3a8a57b498ed2d88c3fd9fb8d56dc7c646) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53818026)</sub>

**Context used:**


flora131:
# PR #2457 — live end-to-end verification

**Verdict: CONFIRMED.** In a live Atomic TUI session, Enter on a focused `ctx.tool` graph node opens a collapsed, flat, agent-chat-style tool message block with a bounded preview; `ctrl+o` expands it to full bounded args/result/error/source/timing/markers and collapses it back byte-for-byte; a failed node's four-line error survives collapsed with a two-space continuation inset at both wide and 60-column widths; and the same workflow driven against a pre-fix build of the direct parent commit shows the old rounded standalone `╭ TOOL … ╰` panel with no collapsed state and a `ctrl+o` no-op — so the captures prove causation, not coincidence.

## Environment

| Item | Value |
|---|---|
| Worktree | `/Users/tonystark/Documents/projects/atomic-pr-2454-followup` (branch `pr-2457`, clean tree) |
| Fixed-build HEAD | `754f5a3f960f945579201202146b0d1ada94f387` — `fix(workflows): preserve multiline tool error previews` |
| Binary | built in this worktree: `packages/coding-agent/dist/atomic` (reporting `0.0.0`, versionless release base) |
| Pre-fix baseline | detached scratch worktree `/tmp/pr2457-prefix` at `7bb38b2bdc9e3c720c55f7ea6f684cb18bacfebd` (`docs(workflows): state full discovery extension set and shared lib layout`) — the merge-base of `pr-2457` and `origin/main`, i.e. the commit immediately under the two PR commits |
| Fixed vs PR head | `git diff 3c90c45c20..754f5a3f96` over the PR's eight code/test files under `packages/workflows/src` and `test/unit/` is **empty** (byte-identical); only `docs/workflows.md`, `README.md`, `CHANGELOG.md` differ, by documentation lines that came from other commits already on `main` |
| Fixed vs baseline | `git diff --stat 7bb38b2bdc..754f5a3f96` = exactly the PR's 11 files, 731 insertions / 206 deletions — nothing else differs, so the A/B builds isolate the PR |
| Runtime | bun 1.3.14, node v26.7.0, tmux 3.7b, macOS darwin-arm64; host config `~/.atomic/agent` (no global `atomic` binary invoked at any point) |

## Attribution

- **@flora131 (Norin Lavaee) authored all of the PR's work.** On the GitHub PR head `fix/2454-tool-block-inspection`: `590f1282da` *"fix(workflows): refine ctx.tool graph inspection"* and `3c90c45c20` *"fix(workflows): preserve multiline tool error previews"*, on top of their merged #2454 commit `662faa4f4c` *"feat(workflows): inspect ctx.tool graph nodes"*.
- The local `pr-2457` branch carries the same two changes reworked onto newer `main` as `3fb9067899` and `754f5a3f96`, both authored by Norin Lavaee; their diffs over the PR's code and test files are byte-identical to the PR head's (verified above).
- **No commits were added in this verification session.** This session only built binaries, drove live sessions, and wrote this evidence file (left uncommitted). Nothing was squashed, amended, re-authored, pushed, merged, or commented.

## Commands run

Setup and identity:

```sh
git status && git log --oneline -10 && git branch --show-current && git remote -v
git fetch origin main --quiet && git merge-base origin/main pr-2457   # -> 7bb38b2bdc
git diff --stat 662faa4f4c..3c90c45c20                               # PR as posted
git diff --stat 7bb38b2bdc..754f5a3f96                               # local rework vs baseline
git diff --stat 3c90c45c20..754f5a3f96 -- <PR's 11 files>            # code files empty, docs-only residue
gh pr view 2457 --repo bastani-inc/atomic --json title,author,state,commits
```

Fixed binary (worktree HEAD `754f5a3f96`, in tmux session `build-fixed`):

```sh
cd /Users/tonystark/Documents/projects/atomic-pr-2454-followup/packages/coding-agent
git -C ../.. rev-parse HEAD        # 754f5a3f960f945579201202146b0d1ada94f387
npm run build:binary               # exit 0; "Pi 0.84.2 model-data and OAuth runtime assets verified."
./dist/atomic --version            # 0.0.0
```

Pre-fix binary (scratch checkout of the parent commit, in tmux session `build-prefix`):

```sh
git worktree add --detach /tmp/pr2457-prefix 7bb38b2bdc
cd /tmp/pr2457-prefix && npm ci --ignore-scripts
# first attempt with a symlinked node_modules failed: tsgo escaped through the
# symlink and type-checked the WORKTREE's emitted dist; a real npm ci fixed it.
cp /Users/tonystark/Documents/projects/atomic-pr-2454-followup/packages/natives/native/atomic_natives.darwin-arm64.node \
   /tmp/pr2457-prefix/packages/natives/native/     # gitignored cargo artifact; scratch build can't produce it.
   # justified: git diff 7bb38b2bdc..pr-2457 -- packages/natives is empty, so both
   # builds load the byte-identical native module.
cd packages/coding-agent && npm run build:binary   # exit 0
./dist/atomic --version                            # 0.0.0
```

Demo workflow (identical file in `/tmp/pr2457-demo` and `/tmp/pr2457-demo2`, deleted afterwards):

```ts
import { workflow } from "@bastani/workflows";
import { Type } from "typebox";

export default workflow({
	name: "tool-block-probe",
	description: "Drives ctx.tool node inspection: success, multiline failure, oversized payload.",
	outputs: {
		failureOk: Type.Boolean({ description: "Whether the deliberate failure node returned an outcome." }),
	},
	run: async (ctx) => {
		await ctx.tool("probe-success", { phase: "success" }, async () => ({ ok: true, rows: 3 }));
		const failure = await ctx.tool(
			"e2e-failure",
			{ phase: "failure" },
			async () => {
				throw new Error(
					"AssertionError: expected 1 to equal 2\n  at foo (/x/y.ts:3:1)\n  at bar (/x/y.ts:9:2)\nRR-LAST-ERROR-LINE",
				);
			},
			{ failureMode: "return" },
		);
		await ctx.tool("oversized-result", { phase: "oversized" }, async () => ({ payload: "x".repeat(400) }));
		return { failureOk: failure.ok };
	},
});
```

Live driving (tmux, 120×34 unless noted; fixed session `pr2457-fixed`, pre-fix session `pr2457-prefix`; every binary path is the worktree/scratch build, never a global install):

```text
/workflow reload                       # fixed session: import OK after adding the outputs schema
/workflow inputs tool-block-probe      # "Workflow has no declared inputs."
/workflow tool-block-probe             # run 89b74b99-… started in background, completed in ~1s
/workflow connect 89b74b99-ea18-4a61-a7e9-975bfe99dbae
Enter                                  # on focused probe-success  -> collapsed block
ctrl+o / ctrl+o                        # expand / collapse round trip
Escape, Down ×2, Enter                 # oversized-result collapsed
Escape, Up, Enter                      # e2e-failure collapsed (multiline), ctrl+o expanded
tmux resize-window -x 60               # narrow-frame multiline check
ctrl+x, ctrl+x                         # tool block -> graph -> main chat
grep -icE "attach|steer|interrupt|resume|composer" <full scrollback>   # -> 0 matches
# pre-fix session: same launch + connect (via picker), Enter, Down, Enter, then
diff <(capture before ctrl+o) <(capture after ctrl+o)                  # -> IDENTICAL
```

## Captures — fixed build (HEAD `754f5a3f96`)

Launch accepted and run completed (pane tail):

```text
│   started in background                                                                                             │
│ ▸ /workflow connect 89b74b99-ea18-4a61-a7e9-975bfe99dbae  see agents working · chat with and steer each stage        │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
 Run 89b74b99-ea18-4a61-a7e9-975bfe99dbae is active in the background.
╭ BACKGROUND  1 run  ✓ 1 complete ─────────────────────────────────────────────────────────────────────────────────────╮
│   ✓  89b74b99-ea18-4a61-a7e9-975bfe99dbae                                                                            │
│     tool-block-probe · complete · 0s                                                                                 │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

Graph viewer (`/workflow connect …`) — three durable tool nodes, footer advertises `↵ open tool detail`:

```text
 ╭──────────────╮
 │ ORCHESTRATOR │  tool-block-probe                                                                           ✓ 2  ✗ 1
 ╰──────────────╯

                                              ╭─── probe-success ────╮
                                              │     durable tool     │
                                              │     ✓ completed      │
                                              │         root         │
                                              ╰──────────────────────╯
                                                          │
                                              ╭──── e2e-failure ─────╮
                                              │     durable tool     │
                                              │       ✗ failed       │
                                              │        1 dep         │
                                              ╰──────────────────────╯
                                                          │
                                              ╭── oversized-result ──╮
                                              │     durable tool     │
                                              │     ✓ completed      │
                                              │        1 dep         │
                                              ╰──────────────────────╯

 ╭───────╮
 │ GRAPH │                            ctrl+x return to main chat  ·  ↵ open tool detail  ·  ↑↓←→ navigate  ·  / stages
 ╰───────╯
```

**Enter on `probe-success` — the collapsed tool message block** (flat, one-cell inset, no border, bounded preview, expand hint):

```text
 ✓ probe-success {"phase":"success"}
   ✓ {"ok":true,"rows":3}
   (ctrl+o to expand)
```

**`ctrl+o` — expanded** (full bounded fields):

```text
 ✓ probe-success {"phase":"success"}
 ARGS      {"phase":"success"}
 RESULT    {"ok":true,"rows":3}
 SOURCE    async () => ({
                 ok: true,
                 rows: 3
               })
 TIMING    startedAt=1786920357846 · endedAt=1786920357846 · duration=0ms (0s)
 MARKERS   —
   (ctrl+o to collapse)
```

**Second `ctrl+o` — collapsed again, byte-for-byte identical to the first collapsed frame:**

```text
 ✓ probe-success {"phase":"success"}
   ✓ {"ok":true,"rows":3}
   (ctrl+o to expand)
```

**Oversized node (400-char payload) — collapsed preview carries the truncation marker:**

```text
 ✓ oversized-result {"phase":"oversized"}
   ✓ {"payload":"xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx…
   (ctrl+o to expand)
```

**Failed node — collapsed multiline error: every line present, two-space continuation inset, last line `RR-LAST-ERROR-LINE` survives:**

```text
 ✗ e2e-failure {"phase":"failure"}
   ✗ AssertionError: expected 1 to equal 2
     at foo (/x/y.ts:3:1)
     at bar (/x/y.ts:9:2)
   RR-LAST-ERROR-LINE
   (ctrl+o to expand)
```

**Failed node expanded** (`ERROR` keeps the real newlines; `SOURCE`, `TIMING`, `MARKERS` present):

```text
 ✗ e2e-failure {"phase":"failure"}
 ARGS      {"phase":"failure"}
 ERROR     AssertionError: expected 1 to equal 2
             at foo (/x/y.ts:3:1)
             at bar (/x/y.ts:9:2)
           RR-LAST-ERROR-LINE
 SOURCE    async () => {
                   throw new Error(
                     "AssertionError: expected 1 to equal 2\n  at foo (/x/y.ts:3:1)\n  at bar (/x/y.ts:9:2)\nRR-LAST-ERR
           OR-LINE"
                   );
                 }
 TIMING    startedAt=1786920357887 · endedAt=1786920357888 · duration=1ms (0s)
 MARKERS   —
   (ctrl+o to collapse)
```

**Same failed node at 60 columns (`tmux resize-window -x 60`) — all four error lines still kept:**

```text
 ╭──────────────╮
 │ ORCHESTRATOR │  tool-block-probe               ✓ 2  ✗ 1
 ╰──────────────╯
 ✗ e2e-failure {"phase":"failure"}
   ✗ AssertionError: expected 1 to equal 2
     at foo (/x/y.ts:3:1)
     at bar (/x/y.ts:9:2)
   RR-LAST-ERROR-LINE
   (ctrl+o to expand)
```

Return path and control-surface check:

```text
# ctrl+x from the tool block -> graph (footer back to "ctrl+x return to main chat"),
# ctrl+x again -> main chat, with the run summary printed:
 - oversized-result: passed; summary was truncated
 - Final result: {"failureOk": false}

 Connected to 89b74b99-ea18-4a61-a7e9-975bfe99dbae. h hide · ctrl+x leave graph · return to main chat · esc close.

# grep -icE "attach|steer|interrupt|resume|composer" over the session's full scrollback:
0
```

## Captures — pre-fix build (parent commit `7bb38b2bdc`, same workflow, same terminal size)

Graph viewer renders identically (the PR does not change the graph):

```text
 ╭──────────────╮
 │ ORCHESTRATOR │  tool-block-probe                                                                           ✓ 2  ✗ 1
 ╰──────────────╯

                                              ╭─── probe-success ────╮
                                              │     durable tool     │
                                              │     ✓ completed      │
                                              │         root         │
                                              ╰──────────────────────╯
                                              ... e2e-failure / oversized-result identical ...
 ╭───────╮
 │ GRAPH │                            ctrl+x return to main chat  ·  ↵ open tool detail  ·  ↑↓←→ navigate  ·  / stages
 ╰───────╯
```

**Enter on `probe-success` — the OLD behavior: a rounded standalone panel, always fully expanded, no collapsed state, no expand hint:**

```text
            ╭ TOOL probe-success ──────────────────────────────────────────────────────────────────────────╮
            │NAME      "probe-success"                                                                     │
            │STATUS    completed                                                                           │
            │ARGS      {"phase":"success"}                                                                 │
            │RESULT    {"ok":true,"rows":3}                                                                │
            │SOURCE    async () => ({                                                                      │
            │                ok: true,                                                                     │
            │                rows: 3                                                                       │
            │              })                                                                              │
            │TIMING    startedAt=1786920640166 · endedAt=1786920640189 · duration=23ms (0s)                │
            │MARKERS   —                                                                                   │
            ╰──────────────────────────────────────────────────────────────────────────────╯
```

**Enter on `e2e-failure` — the old panel renders the same error as escaped `\n` JSON strings, not as preserved multiline rows:**

```text
            ╭ TOOL e2e-failure ────────────────────────────────────────────────────────────────────────────╮
            │NAME      "e2e-failure"                                                                       │
            │STATUS    failed                                                                              │
            │ARGS      {"phase":"failure"}                                                                 │
            │RESULT    {"ok":false,"error":{"name":"Error","message":"AssertionError: expected 1 to equal 2│
            │          \n  at foo (/x/y.ts:3:1)\n  at bar (/x/y.ts:9:2)\nRR-LAST-ERROR-LINE"},"attempts":1,│
            │          "cached":false}                                                                     │
            │ERROR     "AssertionError: expected 1 to equal 2\n  at foo (/x/y.ts:3:1)\n  at bar (/x/y.ts:9:│
            │          2)\nRR-LAST-ERROR-LINE"                                                             │
            │SOURCE    async () => {                                                                       │
            │                  throw new Error(                                                            │
            │                    "AssertionError: expected 1 to equal 2\n  at foo (/x/y.ts:3:1)\n  at bar (│
            │          /x/y.ts:9:2)\nRR-LAST-ERROR-LINE"                                                   │
            │                  );                                                                          │
            │                }                                                                             │
            │TIMING    startedAt=1786920640413 · endedAt=1786920640414 · duration=1ms (0s)                 │
```

**`ctrl+o` on the pre-fix tool detail is a no-op** (diff of the pane before/after the keypress):

```text
$ diff /tmp/prefix-before-ctrlo.txt /tmp/prefix-after-ctrlo.txt
IDENTICAL: ctrl+o is a no-op on the pre-fix tool detail
```

## What the A/B proves

Only the PR's 11 files differ between the two builds (`git diff --stat 7bb38b2bdc..754f5a3f96`). With everything else held constant:

1. **The UI change is real and causal**: pre-fix Enter opens a bordered, always-expanded `╭ TOOL … ╰` standalone panel; post-fix Enter opens a flat, inset, collapsed agent-chat-style message block — no `renderRoundedBoxLines` border anywhere in the post-fix capture.
2. **Collapsed-by-default with bounded preview** including the `…` truncation marker on the 400-character payload.
3. **`ctrl+o` expand/collapse round trip** works post-fix (full fields, then a byte-identical collapsed frame) and does nothing pre-fix.
4. **Multiline error preservation** — the exact regression the PR names: all four error lines survive in the collapsed block with the two-space inset, at 120 and at 60 columns, and the expanded `ERROR` field renders real newlines where the old panel printed escaped `\n` strings.
5. **Read-only, non-attachable surface**: zero occurrences of attach/steer/interrupt/resume/composer in the fixed session's scrollback; `escape`/`ctrl+x` return from block → graph → main chat.

Scope note: the PR's payload-bound hardening (bounded key discovery, boundary types, truncation-key collision) is exercised here only through the display path (oversized payload, hostile-free); its trap-counting proof lives in `test/unit/workflow-tool-payload-bounds.test.ts`, which was not re-run in this session. The TUI behavior above is the part driven live, end to end, from a real build of this worktree.

flora131:
Valid point — fixed in c16aed3a8a. `toolExpandKey` now takes `KeybindingsLike | undefined` instead of `unknown`, and the call site narrows `piKeybindings` with `isKeybindingsLike` once at the payload boundary, matching the existing `_promptKeybindings()` pattern in `graph-view-input.ts`. No behavior change: a missing manager or `getKeys` still falls back to `ctrl+o`.

PR #2519, opened 2026-08-18
fix(rpc): preserve queue control lane semantics

### LINKED ISSUE #2516: Interactive engine: clear_queue / resume_queued_messages time out at 30s because they queue behind the in-flight prompt frame
### Summary

`clear_queue` and `resume_queued_messages` are dispatched on the engine's **ordinary** input lane, so they queue behind the in-flight `prompt` frame — but on the host they are **not** in `LONG_LIVED_COMMANDS`, so they carry the 30s default deadline. Any turn that takes longer than 30s to unwind after Escape therefore produces spurious red timeouts, and repeated Ctrl+C escalates into killing the engine child.

Observed on `@bastani/atomic` `0.9.14-alpha.4`.

```
Error: Timeout waiting for response to resume_queued_messages. Stderr:
Error: Interactive engine clear queue failed: Timeout waiting for response to clear_queue. Stderr:
Error: Agent process stopped
```

The empty `Stderr:` is the tell — the engine is alive and healthy, it just has not reached the frame yet.

### Root cause

`modes/rpc/rpc-input-scheduler.js` splits engine input into two lanes:

```text
concurrent lane (serviced immediately, even mid-turn)
  abort · abort_bash · abort_compaction · abort_retry
  cancel_login_provider · pause_queued_messages
  bash · user_bash · refresh_models
  extension_ui_response

ordinary lane (strictly one frame at a time)
  prompt  ←── an agent turn occupies this for its whole duration
  clear_queue
  resume_queued_messages
```

`modes/rpc/rpc-command-timeouts.js` splits host requests into two deadline classes: `LONG_LIVED_COMMANDS` (unbounded — `prompt`, every `abort*`, `steer`, `follow_up`, `bash`, `compact`, …) and everything else at `DEFAULT_REQUEST_TIMEOUT_MS = 30_000`.

The queue-control trio is split across both axes:

| command                   | engine lane | host deadline | can it land mid-turn? |
| ------------------------- | ----------- | ------------- | --------------------- |
| `pause_queued_messages`   | concurrent  | 30s           | yes                   |
| `resume_queued_messages`  | ordinary    | 30s           | **no**                |
| `clear_queue`             | ordinary    | 30s           | **no**                |

Pause can always get through. Un-pause and clear cannot — they sit behind the very `prompt` frame they are trying to unwind, while a 30s timer runs against them.

This is not a stuck engine. Escape is cooperative by design, and `rpc-command-timeouts.js` already documents exactly this reasoning for exempting `abort`:

> A host-side deadline would surface as `Timeout waiting for response to abort…` — a red error about the host's own impatience while the engine is still working — and would not stop anything.

That reasoning applies verbatim to the two commands that unwind the queue after an abort.

### Secondary issue: host/engine queue state diverges

`modes/interactive-engine/isolated-runtime.js` (`clearQueue`) empties the local `steeringMessages` / `followUpMessages` arrays and *then* sends the remote `clear_queue` through `dispatchBestEffort`. When the remote half times out, the host has already cleared its copy and the engine has not — the two halves disagree, and the only surface for the failure is the `Interactive engine clear queue failed: …` diagnostic.

### Reproduction

1. Start a named workflow that reaches a stage awaiting human input (any HiL prompt).
2. In a single assistant turn, issue two slow tool calls in one parallel block — e.g. a `workflow` control call (`quit`/`interrupt`) against that run, plus a network fetch.
3. Press Escape while both are in flight, then type/queue a message.

The `workflow` control call has to tear down a stage parked on a pending human prompt, and the fetch has its own retry budget, so the `prompt` frame takes well over 30s to unwind. Both tool calls correctly report `Tool execution was interrupted after it started`, and meanwhile the queue commands blow their deadline.

Repeated Ctrl+C then escalates past cooperative abort into killing the child, and `modes/rpc/rpc-client.js` rejects every pending request with `rpcTransportError("Agent process stopped")`.

### Why `workflow` control calls make this likely

A workflow control call is about the longest-unwinding thing an ordinary frame can hold: `quit`, `interrupt`, and `send` reach into a durable graph whose stages may be parked on their own HiL prompts or inside a `ctx.tool` callback with no `timeoutMs`. And the moment a user reaches for Escape and starts typing is precisely when a stage *looks* stuck on a prompt. The two conditions co-occur by construction, which matches the scattered reports of `workflow`-adjacent engine stalls.

Adjacent, already-fixed history in `CHANGELOG.md` — the queued-message stall that was "most visible after answering an `ask_user_question` questionnaire" — suggests this corner (HiL + queued input + interrupt) is worth a second pass as a whole.

### Expected

Escape and Ctrl+C settle the queue without red errors, however long the turn takes to unwind cooperatively. A user who waits should never need a second Ctrl+C, and should never lose the engine child.

### Suggested fix

**Primary — put the queue-control commands on the lane their sibling already uses.** In `modes/rpc/rpc-input-scheduler.js`:

```diff
 const INTERRUPT_COMMANDS = new Set([
     "abort",
     "cancel_login_provider",
     "abort_compaction",
     "abort_retry",
     "abort_bash",
     "pause_queued_messages",
+    "resume_queued_messages",
+    "clear_queue",
 ]);
```

These mutate queue state, not turn state; there is no ordering reason for them to sit behind a `prompt`. `clearQueue()` returning the drained queue snapshot is unaffected, and `resumeQueuedMessages()` already awaits the outstanding pause request before releasing (`modes/interactive-engine/remote-queue-pause.js`), so the pause→resume ordering is preserved by the promise, not by the lane.

**Fallback, if ordering behind `prompt` turns out to be load-bearing** — add both to `LONG_LIVED_COMMANDS` in `modes/rpc/rpc-command-timeouts.js`. That stops the host reporting its own impatience, exactly as it already does for `abort`. Failure detection is unchanged: engine exit, transport violations, generation replacement, and explicit stop reject pending requests independently of any timer.

**Also** — have `clearQueue` reconcile (or not pre-clear) the host-side arrays when the remote `clear_queue` fails, so a timeout cannot leave the two halves disagreeing about what is queued.

### Workaround

After Escape, wait rather than pressing again. The first abort is already unbounded and is working; the extra Ctrl+C presses are what escalate to killing the child. The 30s errors are cosmetic if the engine survives.

flora131:
## Summary
Fixes queue-control handling for issue #2516 by routing control requests through the interrupt lane and preserving authoritative queue state when remote clearing fails.

## Changes

- Route RPC queue controls through the interrupt lane so they are admitted ahead of ordinary queued work.
- Restore the host queue only when a failed `clear_queue` leaves it empty.
- Avoid duplicating entries when a `queue_update` already supplied the engine's authoritative queue.
- Add focused scheduler and isolated-runtime regression coverage.
- Update the coding-agent changelog.

## Notes

Focused tests, the package test suite, and `npm run check` pass. The standard push hook also ran the full suites; unrelated pre-existing environment/base failures remain in semver pin, workflow scope-guard, and native release publisher contract tests.

Closes #2516

---
<a href="https://app.blacksmith.sh/bastani-inc/codesmith/atomic/pr/2519"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789687747&installation_model_id=10562&pr_number=2519&repository=bastani-inc%2Fatomic&return_to=https%3A%2F%2Fgithub.com%2Fbastani-inc%2Fatomic%2Fpull%2F2519&signature=f9691a24c9e15a6956f123ffa279b4dc5bfb3121dfe93555e303534537915e13"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

<h3>Greptile Summary</h3>

Queue clearing now reconciles local state with engine state across authoritative updates, overlapping requests, and engine replacement. The previously reported stale-empty-update rollback was disproved: an empty authoritative update remains empty after the request fails. The previously reported timing-dependent overlapping rollback and lost-original-queue behaviors were disproved in both failure orders. The previously reported retired-engine rollback was also disproved: a request from the replaced engine cannot restore its old queue snapshot.

No blocking failure remains in the queue-clear reconciliation behavior.

No accepted blocking findings remain. Focused runtime checks exercised authoritative empty updates, overlapping failures in both settlement orders, mixed success and failure, and rejection after engine replacement.

**What T-Rex did**
- The runtime harness emitted an authoritative empty queue update after an optimistic clear and then rejected clear\_queue, leaving both steering and follow-up queues empty.
- Two overlapping clearQueue() calls with no later queue update were issued; in both oldest-first and newest-first orders, the queue stayed cleared until the final rejection and then restored the original messages, showing rollback behavior is independent of settlement order.
- The focused queue-clear recovery test completed with seven tests passing and demonstrated that replacement clears the pending rollback record so retired requests cannot restore old messages.
- A test of overlapping clear requests with no later update finished with empty host queues in both orders, confirming that one successful engine clear suppresses rollback of stale messages.
- The requested verification ran, but local artifact references were not uploaded.

<a href="https://app.greptile.com/trex/runs/19522213/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

<sub>Reviews (5): Last reviewed commit: ["fix(rpc): invalidate clears from retired..."](https://github.com/bastani-inc/atomic/commit/8115519ca9221a56980a9111fb6285d1b8908844) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54573901)</sub>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Queue-control RPC frames now remain reachable during long-running prompts, and failed remote queue clears restore host-queued messages ([#2516](https://github.com/bastani-inc/atomic/issues/2516)).
+
 ## [0.9.14-alpha.4] - 2026-08-18
 
 ### Changed
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixed
 
-- Queue-control RPC frames now remain reachable during long-running prompts, and failed remote queue clears restore host-queued messages ([#2516](https://github.com/bastani-inc/atomic/issues/2516)).
+- Queue-control RPC frames now remain reachable during long-running prompts. A failed remote `clear_queue` restores host-queued messages only when no later `queue_update` has already supplied engine truth, including an empty queue ([#2516](https://github.com/bastani-inc/atomic/issues/2516)).
 
 ## [0.9.14-alpha.4] - 2026-08-18
 
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixed
 
-- Queue-control RPC frames now remain reachable during long-running prompts. A failed remote `clear_queue` restores host-queued messages only when no later `queue_update` has already supplied engine truth, including an empty queue ([#2516](https://github.com/bastani-inc/atomic/issues/2516)).
+- Queue-control RPC frames now remain reachable during long-running prompts. When all overlapping remote `clear_queue` calls fail, Atomic restores the pre-clear host queue only if no later `queue_update` has supplied engine truth, including an empty queue ([#2516](https://github.com/bastani-inc/atomic/issues/2516)).
 
 ## [0.9.14-alpha.4] - 2026-08-18
 
--- a/packages/coding-agent/src/modes/interactive-engine/isolated-runtime.ts
+++ b/packages/coding-agent/src/modes/interactive-engine/isolated-runtime.ts
@@ -408,7 +408,15 @@ export class IsolatedInteractiveRuntime extends AgentSessionRuntime {
 					const queued = { steering: [...this.steeringMessages], followUp: [...this.followUpMessages] };
 					this.steeringMessages = [];
 					this.followUpMessages = [];
-					this.dispatchBestEffort("clear queue", this.client.requestInternal({ type: "clear_queue" }));
+					const clearRequest = this.client.requestInternal({ type: "clear_queue" });
+					this.dispatchBestEffort(
+						"clear queue",
+						clearRequest.catch((error: Error) => {
+							this.steeringMessages = [...queued.steering, ...this.steeringMessages];
+							this.followUpMessages = [...queued.followUp, ...this.followUpMessages];
+							throw error;
+						}),
+					);
 					return queued;
 				},
 			},
--- a/packages/coding-agent/src/modes/interactive-engine/isolated-runtime.ts
+++ b/packages/coding-agent/src/modes/interactive-engine/isolated-runtime.ts
@@ -22,6 +22,16 @@ import type { EngineKeybindingState, InteractiveEngineCommand, InteractiveEngine
 import { RemoteCommandCatalog, type RemoteCommandsListener } from "./remote-command-catalog.ts";
 import { RemoteModelCatalog } from "./remote-model-catalog.ts";
 import { RemoteQueuePause } from "./remote-queue-pause.js";
+
+type QueueSnapshot = { steering: string[]; followUp: string[] };
+
+type PendingQueueClear = {
+	snapshot: QueueSnapshot;
+	queueUpdateGeneration: number;
+	count: number;
+	succeeded: boolean;
+};
+
 /**
  * Owns Atomic's local interactive host facade and child-process engine.
  *
@@ -38,13 +48,10 @@ export class IsolatedInteractiveRuntime extends AgentSessionRuntime {
 	private readonly activeBashRequestIds = new Map<string | symbol, string>();
 	private steeringMessages: string[] = [];
 	private followUpMessages: string[] = [];
-	/**
-	 * Bumped by every authoritative queue_update and every optimistic local
-	 * clear. A failed clear_queue restores its snapshot only while the
-	 * generation it captured is still current, so neither an engine update nor
-	 * a newer clear can be overwritten by a stale rollback.
-	 */
-	private queueGeneration = 0;
+	/** Bumped by every authoritative queue_update. */
+	private queueUpdateGeneration = 0;
+	/** Clears started before the next queue_update share one rollback snapshot. */
+	private pendingQueueClear: PendingQueueClear | undefined;
 	private engineCallbackActive = false;
 	private readonly queuePause: RemoteQueuePause;
 	private autoCompactionEnabled = true;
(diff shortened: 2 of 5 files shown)


flora131:
TMUX EVIDENCE (issue #2516)

This evidence came from the dedicated tmux session `atomic-2516-evidence:0.0`, using a real logged-in CLI from the worktree (not a mock or isolated harness).

# Issue 2516 tmux evidence

- **Timestamp (UTC):** 2026-08-18T23:20:02Z
- **Session target:** `atomic-2516-evidence:0.0`
- **Launch command:**
  ```sh
  cd /Users/tonystark/Documents/projects/atomic-issue-2516/packages/coding-agent && exec env ATOMIC_CODING_AGENT_DIR="$HOME/.atomic/agent" bun src/bun/cli.ts --provider xai --model grok-4.6
  ```
- **Exercise:** Started `sleep 45; printf "issue-2516 slow turn complete\n"`, pressed Escape after the command had run for about 13 seconds, then submitted `Queued follow-up after Escape: reply exactly QUEUE_CONTROL_LANE_OK and do not run tools.` Waited 37 seconds after queuing, which exceeds the 30-second host deadline.
- **Pane state after the wait:** `command=bun dead=0 pid=36174`
- **Capture command:** `tmux capture-pane -t atomic-2516-evidence:0.0 -p -S -`

## Raw captured pane text

```text

   ██████▙                  ▟██████    Atomic v0.0.0
    ██████▙                ▟██████░░   (xai) grok-4.6 xhigh
     ██████▙              ▟██████░░    ~/Documents/projects/atomic-issue-2516/packages/coding-agent
      ██████▙            ▟██████░░
       ████████████████████████░░      We question,
        ██████▛░░░░░░░░▜██████░░       we break away from what is accepted.
         ██████▛      ▜██████░░        Engineering matters.
          ██████▛    ▜██████░░
           ██████▛  ▜██████░░
            ░████████████░░░
              ░░░░░░░░░░░░

RESOURCES ~/Documents/projects/atomic-issue-2516/AGENTS.md · 18 skills · 14 prompts



 Run only this exact Bash command, then wait for it to finish: sleep 45; printf "issue-2516 slow turn complete\\n"


 Thinking...

 I'll run that exact command and wait for it to finish.


 $ sleep 45; printf "issue-2516 slow turn complete\n" (timeout 60s)

 Command aborted

 Took 13s


 Request aborted


 Queued follow-up after Escape: reply exactly QUEUE_CONTROL_LANE_OK and do not run tools.


 Thinking...

 QUEUE_CONTROL_LANE_OK




                                                                                                                      ↑26k • ↓105 • R26k • CH99.3% • $0.065 (sub) • 5.1%/500k (auto)
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
❯
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
(xai) grok-4.6 xhigh • ~/Documents/projects/atomic-issue-2516/packages/coding-agent (fix/2516-queue-control-lane)
```

## Verdict

Pass. Escape aborted the slow Bash turn, the queued follow-up ran and returned `QUEUE_CONTROL_LANE_OK`, and the pane remained live after 37 seconds. The capture has none of the four failure strings.

Issue #2627, opened 2026-08-24
intercom: blocking ask hangs 10 minutes when target session disconnects after delivery — add broker peer-disconnect release


flora131:
## Summary

A blocking `intercom.ask` (and `contact_supervisor`'s blocking wait, which shares the same reply-waiter slot) hangs for the full 10-minute reply timeout when the **target session disconnects after delivery succeeded**. Delivery is verified exactly once, at send time; after that, nothing releases the waiter when the peer unregisters from the broker. The asker's turn cannot complete until the waiter settles, so the whole session appears stalled.

This is the dominant silent-hang mode for asks targeting `subagent-chat-*` sessions (the runtime alias for unnamed subagent sessions): the subagent finishes its run and disconnects, and the asker rides out `DEFAULT_REPLY_TIMEOUT_MS = 10 * 60 * 1000` for a reply that can never arrive.

## Observed behavior

Two parallel `intercom.ask` calls in one assistant turn, both targeting `subagent-chat-*` sessions:

- One ask atomically wins the single-slot waiter; the other correctly fails fast with `"Already waiting for a reply"` (working as designed since the atomic check-and-reserve fix).
- The winning ask delivered successfully, then both target sessions disconnected without replying. Broker listing confirmed neither target remained in the group.
- The winner blocked on `wait.promise` with only three exits: a correlated reply (impossible — peer gone), caller abort, or the 10-minute timeout. The turn — and visually *both* tool calls — stalled until timeout.

The same gap makes mutual asks (A asks B while B asks A) and asks to any child that exits mid-wait degrade into full-timeout hangs instead of prompt errors.

## Root cause

- `packages/intercom/intercom-tool.ts` (`case "ask"`): `sendResult.delivered` is checked once right after `connectedClient.send(...)`. A target that dies *after* that point is indistinguishable from a slow replier.
- `packages/intercom/reply-waiter.ts`: `ReplyWaiterSlot` settles only via correlated reply, per-call `cancel` (abort/send-failure), `rejectCurrent` (slot-wide teardown), or the per-waiter timeout.
- `packages/intercom/index-heavy.ts`: `rejectCurrent` fires only on the **asker's own** broker disconnect / session shutdown-replacement. There is no path keyed on the **target's** lifecycle.
- The broker (`packages/intercom/broker/broker.ts`) observes every session unregister/socket close, but discards that knowledge with respect to in-flight `expectsReply` messages. It never notifies senders that a pending question's recipient is gone.

[...]

PR #2640, opened 2026-08-24
feat(intercom): concurrent-ask call-site audit, edge cases, and docs

### LINKED ISSUE #2628: intercom: support concurrent blocking asks via correlation-keyed multi-waiter registry (blocked by #2627)
## Summary

Support **N concurrent blocking `intercom.ask` calls** per session by replacing the single-slot `ReplyWaiterSlot` with a correlation-keyed waiter registry. Today, when sibling tool calls in one assistant message issue two asks concurrently (the default parallel tool execution mode), exactly one wins the atomic check-and-reserve and every other call returns a structured `"Already waiting for a reply"` error. After this change, each ask gets its own waiter, all questions go out immediately, and the turn resolves when each reply (or per-waiter timeout/abort/peer-disconnect release) settles — parallel waiting instead of one-wins/one-errors.

**Blocked by #2627.** Without the broker peer-disconnect release, N outstanding asks means N independent chances per turn of a silent 10-minute dead-peer hang, with the turn stalled on max(all of them). With #2627 landed, a dead peer fails in ~1s, making concurrent waiters cheap to hold open. Do not land this first.

## Current behavior and why it exists

- `packages/intercom/reply-waiter.ts` — `ReplyWaiterSlot` is a single-slot waiter with atomic synchronous check-and-reserve. Concurrent losers get a structured busy refusal instead of a rejected promise.
- The single slot was itself a bug fix: the prior multi-path behavior produced losers holding already-rejected promises that sat unhandled, and failure paths that tore down *other* calls' waiters. Any multi-waiter design must preserve those two guarantees (no unhandled rejections; cleanup scoped to the owning call).
- Reply correlation is **not** the blocker: `routeIncomingReply` (`packages/intercom/reply-routing.ts`) already matches on exact sender + `replyTo === questionId`, so multiple in-flight waiters are distinguishable with zero protocol changes.

## Proposed design

1. **Keyed registry.** Replace the single slot with `Map<questionId, ReplyWaiterRecord>`. Admission becomes per-question: generate `questionId`, insert atomically, no busy refusal for ordinary concurrent asks. Keep the existing per-record shape — per-waiter timeout (`DEFAULT_REPLY_TIMEOUT_MS`), per-call abort via the tool signal, pre-attached rejection handler so an unawaited pending promise can never become an unhandled rejection, and `cancel` scoped to the owning call.
2. **Reply routing.** Route an inbound reply to the exact waiter whose `(from, replyTo)` matches; unmatched replies fall through to normal message delivery exactly as today. First settle wins per waiter; late/duplicate settlement is a no-op via the existing settled-guard.
3. **Peer-disconnect release (from #2627).** The broker's `peer_disconnected` frame carries `replyTo`, so it rejects exactly the affected waiters when a target dies — this generalizes to N waiters for free because the release is correlation-keyed, not slot-wide.
4. **Slot-wide teardown becomes reject-all.** Session shutdown/replacement and own-broker-disconnect (`rejectCurrent` in `packages/intercom/index-heavy.ts`) iterate and reject every pending waiter with the teardown error. Registry entries are removed on any settle path (resolve, timeout, abort, cancel, disconnect release, teardown).
5. **Concurrency cap.** Add a small per-session cap on outstanding waiters (suggest 4–8, config-overridable). At the cap, return the existing structured busy-style refusal naming the cap (`"Too many pending asks (N); reply-wait slots are full"`). This bounds worst-case turn stall and broker pending-question state. Duplicate concurrent asks to the *same* target remain allowed — distinct `questionId`s correlate independently.
6. **Cross-tool audit — the singular-pending-ask assumptions.** Everything that reasons about "the pending ask" as singular must become per-waiter or explicitly keep a stricter policy:
   - `hasReplyWaiter()` guards in `packages/intercom/intercom-tool.ts` (`case "ask"`) — replaced by cap check.
   - `contact_supervisor` (`packages/intercom/contact-supervisor-tool.ts`) shares the slot today. Decide and document: either it joins the registry like any waiter, or supervisor escalation remains exclusive (one outstanding supervisor wait, and/or mutually exclusive with peer asks). Exclusive-supervisor is the conservative default; do not silently change escalation semantics.
   - Parent-ask handoff (`packages/intercom/parent-ask-handoff.ts`) and foreground detach handshake (`packages/intercom/foreground-detach-handoff.ts`) — verify they either operate per-waiter or are unaffected because they intercept before waiter admission.
   - Workflow busy-target reservation and post-mortem ask scheduling (`workflow-reply-tracker.ts`, `late-stage-message-router.ts`) — confirm target-side admission handles multiple concurrent inbound asks from one sender.
7. **UI/rendering.** Multiple simultaneous "waiting for reply" states must render sanely (tool-call result renderers, session-list overlay).

## Explicitly not in scope

- **Queueing** a losing ask behind an occupied slot (serialized blocking doubles worst-case stall; rejected in #2627 discussion).
- **Non-blocking/scheduled ask** — that contract already exists as `send` + reply hints.
- Changing the 10-minute default reply timeout or the broker protocol beyond what #2627 introduces.

## Acceptance criteria

- [ ] Two (and up to cap) concurrent `intercom.ask` tool calls in one assistant turn each deliver immediately and each resolve with their own correlated reply; no `"Already waiting for a reply"` for ordinary concurrency under the cap.
- [ ] Replies arriving out of order (second question answered first) resolve the correct waiters.
- [ ] Two concurrent asks to the same target correlate independently by `questionId`.
- [ ] Per-waiter timeout, abort, send-failure cancel, and #2627 peer-disconnect release settle **only** the owning waiter — never a sibling's (scoped-cleanup guarantee preserved).
- [ ] No pending waiter promise can become an unhandled rejection, including waiters settled while their owner is between awaits.
- [ ] Session shutdown/replacement and own-broker-disconnect reject all pending waiters; registry is empty afterward (no leaks; assert via test).
- [ ] Cap enforcement returns a structured, actionable refusal; cap is configurable.
- [ ] `contact_supervisor` interaction decided, implemented, and documented (registry member vs. exclusive).
- [ ] Unit tests: registry admission/settle/teardown matrix, out-of-order replies, same-target concurrency, cap boundary, reply-vs-timeout-vs-disconnect races, unhandled-rejection guard.
- [ ] Integration test: parallel sibling tool calls issuing two asks to two live sessions, both replied, turn completes with both results.
- [ ] Docs updated: `packages/intercom/README.md`, `packages/intercom/skills/intercom/SKILL.md` (retire the "Already waiting" pattern), `packages/coding-agent/docs/intercom.md`; `packages/intercom/CHANGELOG.md` `[Unreleased]` → `### Added`/`### Changed`.

## References

- #2627 — broker peer-disconnect release (hard prerequisite)
- `packages/intercom/reply-waiter.ts` — single-slot waiter to be replaced
- `packages/intercom/reply-routing.ts` — existing exact sender/thread correlation
- `packages/intercom/intercom-tool.ts`, `packages/intercom/contact-supervisor-tool.ts` — admission call sites
- `packages/intercom/index-heavy.ts` — slot-wide teardown call site
- `packages/intercom/parent-ask-handoff.ts`, `packages/intercom/foreground-detach-handoff.ts` — singular-ask assumptions to audit

flora131:
Part 4/4 of the stacked implementation for #2627 and #2628.

Audits every singular-pending-ask assumption (contact_supervisor policy, parent-ask handoff, foreground detach handshake, reply tracking, pending/reply actions, UI rendering), handles the concurrency edge cases from the #2628 amendment (parallel subagents, parallel children contacting their parent, mixed-session fan-out), adds focused tests, and updates docs/skill/changelog.

Closes #2628.

---
<a href="https://app.blacksmith.sh/bastani-inc/codesmith/atomic/pr/2640"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790181608&installation_model_id=10562&pr_number=2640&repository=bastani-inc%2Fatomic&return_to=https%3A%2F%2Fgithub.com%2Fbastani-inc%2Fatomic%2Fpull%2F2640&signature=af1dbb1baaa04e858fd085557152cf12a66473f359cc4a0f1ab0b888ee53fa07"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

<h3>Greptile Summary</h3>

This change expands Intercom reply handling for concurrent asks and exact reply correlation. Runtime execution reproduced a case where a stale or mistyped reply ID is redirected to a different active request, causing that unrelated request to be marked answered after delivery. This should be corrected before merging so explicit reply correlation remains reliable.

Not safe to merge until an explicit unknown reply ID is rejected instead of selecting another active reply context.

A focused executable reproduction exercised both fallback paths and observed an unrelated current-turn or pending request being selected and removed after a stale reply ID was supplied.

**Files Needing Attention:** packages/intercom/reply-tracker.ts needs to distinguish an omitted reply ID from an explicit reply ID that is no longer pending. Coverage in test/unit/intercom-workflow-reply-tracker.test.ts should cover stale reply IDs with other active contexts.

**What T-Rex did**
- T-Rex produced a proof for the posted P1 finding and attached reproduction artifacts to support the review.
- T-Rex produced a second proof for another posted P1 finding, adding its artifacts for review.
- T-Rex performed general-contract-validation-proof by documenting the reproduction source, before and after outputs, and an artifact-check log to verify contract behavior.

<a href="https://app.greptile.com/trex/runs/20410982/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

1. General comment 

   <a href="#">P1</a> **Stale replyTo is redirected to and clears an unrelated reply context**

   - **Bug**
     - When `replyTo` is supplied but does not identify a pending ask, `resolveReplyTarget` does not reject it. It instead returns the current turn context, or, if no current turn exists, the only pending ask. The focused executable reproduction shows `replyTo: "stale-ask"` selecting `current-turn` and `unrelated-pending-ask`; invoking `markReplied` with each returned target removes that unrelated context. In the actual reply caller, the resolved target is sent with `replyTo: target.message.id` and then cleared on successful delivery.
   - **Cause**
     - `packages/intercom/reply-tracker.ts:54-63` only returns or validates when the exact pending ask exists. A missing exact match falls through into generic fallback resolution at `:64-97`, which was designed for omitted correlation rather than an explicitly supplied, stale correlation ID.
   - **Fix**
     - If `options.replyTo` is provided and no matching pending ask remains after pruning, throw a correlation-specific error instead of applying current-turn or pending-ask fallback. Preserve fallback behavior only when `replyTo` is absent. Add focused coverage for stale `replyTo` with a current turn and with one unrelated pending ask, asserting no context is cleared.

</details>

`````markdown
### Issue 1
packages/intercom/reply-tracker.ts:54-63
**Unknown reply ID selects an unrelated thread**

When `replyTo` is provided but does not match a pending ask, this branch falls through to the generic current-turn or single-pending-ask selection. A stale or mistyped ID can therefore send the reply using another ask’s ID and, after successful delivery, clear that unrelated ask. Reject an explicit `replyTo` that is not pending; only use fallback selection when `replyTo` is absent.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["docs(intercom): add live concurrency ver..."](https://github.com/bastani-inc/atomic/commit/64ab1dd2ba3d77a796a3d2216b063265801bec02) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56326161)</sub>


flora131:
## Live E2E intercom concurrency verification

Ran five real interactive scenarios from this checkout's dev bundle in separately named tmux-driven Atomic sessions. An isolated `ATOMIC_CODING_AGENT_DIR` forced the broker to launch from this branch rather than reuse an installed broker.

| Scenario | Result | Observed live behavior |
| --- | --- | --- |
| A — peer disconnect | **PASS** | Killing only `e2e2628-a-target` released its exact waiter in **0.299 s** with `Session "e2e2628-a-target" disconnected before replying`. |
| B — concurrent asks | **PASS** | Two sibling `intercom.ask` calls were admitted in one turn. C replied immediately; B deliberately slept **6 s**. Both replies correlated to the right call with no `Already waiting for a reply`. |
| C — parallel children | **PASS** | Handoff phase: exactly one child claim won; two losers ended with deterministic `interrupted` outcomes before emitting asks, so zero emitted asks were orphaned. Destination phase: two real parallel child sessions appeared together in the parent receiver's pending list and each received the reply tied to its exact message ID. |
| D — selective disconnect | **PASS** | One batch sent duplicate asks to B plus one to C. Only C was killed. C's wait failed by name; both B IDs remained pending. B replied to the second at `16:54:24.119Z`, slept 6 s, replied to the first at `16:54:42.185Z`, and ended with no unresolved asks. |
| E — cross-tool concurrency | **PASS** | One child message emitted `intercom.ask` plus two `contact_supervisor` calls. At `16:57:00.974Z`, exactly one supervisor handoff returned `Parent ask claimed`; the losing supervisor call and peer ask returned structured `Cancelled` outcomes. No deadlock, crash, or unhandled/uncaught/rejection broker-log entry. |

### A. Peer-disconnect release

Scenario A

### B. Concurrent asks and out-of-order replies

Scenario B

### C. Parallel children and parent-side correlation

Scenario C

### D. Fan-out with selective disconnect

Scenario D

### E. Cross-tool concurrency

Scenario E

PR #2667, opened 2026-08-25
fix(workflows): name resolution awaits an in-flight reload

### LINKED ISSUE #2666: Workflow name resolution races past an in-flight /workflow reload and runs the pre-reload module
## Summary

Workflow name resolution does not await an in-flight `/workflow reload`. A run dispatched immediately after a reload can resolve against the pre-reload registry and execute the old module.

## Mechanism

`packages/workflows/src/extension/extension-runtime-state.ts`, `ensureWorkflowResourcesLoaded()` (lines ~503–512 at `879dc4eb4`):

```ts
async function ensureWorkflowResourcesLoaded(): Promise<void> {
	while (!pi.disableAsyncDiscovery && discoveryRef.current === null) {
		...
	}
}
```

The loop waits only while **no discovery exists at all** (`discoveryRef.current === null`). Once any first discovery has been applied, the function returns immediately — even when `lazyDiscoveryPromise` holds an in-flight reload (tracked by `trackLazyDiscovery`, requested by `reloadWorkflowResources`, same file ~430–441). Every name-resolution path that calls `ensureWorkflowResourcesLoaded()` (slash command dispatch in `workflow-command-registration.ts:102`, durable resume, tool control) can therefore read the registry the reload is about to replace.

Two aggravating factors:

- `pi.sendUserMessage` in the extension bindings is fire-and-forget (`packages/coding-agent/src/core/agent-session-extension-bindings.ts` ~184–192): the internal promise is not returned, so an extension that injects `/workflow reload` followed by `/workflow <name>` has no way to sequence them.
- No reload-completion signal reaches extensions: `message_start`/`message_end` exclude `role: "custom"` messages, and the reload report only goes to the UI reporter.

## Observed live

Found while building the atomic-remote bridge (a Claude Code plugin commanding a running Atomic session) on atomic 0.9.15, macOS, Node 22.22.1. The bridge writes a workflow TS into the project's `.atomic/workflows/`, injects `/workflow reload`, then injects `/workflow <name>`. Observed notification order in the RPC session:

1. `Reloaded workflow resources. 10 workflow(s), generation 2.` — a reload report that had read the file **before** it was rewritten (coalesced with the startup warmup),
2. the run executes the **pre-reload** module (failed with "Workflow run completed without creating any workflow stages or durable tool nodes", which only the old file version could produce),
3. `Reloaded workflow resources. 10 workflow(s), generation 3.` — the requested reload applying after the run had already started.

Re-running the same command immediately afterwards succeeded, confirming a pure ordering race.

## Reproduction

Deterministic unit test (no sleeps): gate `pi.refreshWorkflowResources` so a second reload stays in flight, then call `ensureWorkflowResourcesLoaded()` and assert it stays pending until the reload applies, and that the registry it resolves against contains the newly discovered workflow. On `main` the ensure promise settles immediately while the reload is still in flight.

The test is included in the linked PR (`test/unit/workflow-resolution-awaits-inflight-reload.test.ts`).

## Downstream mitigation currently shipped

The atomic-remote bridge works around this with a flat `RELOAD_SETTLE_MS = 5000` delay between injecting `/workflow reload` and `/workflow <name>`. It papers over the race rather than fixing it, and fails on any project whose discovery takes longer than the delay. With the engine fix, that mitigation becomes unnecessary for engines that include it (the bridge keeps it, version-gated, for older engines).

## Environment

- atomic 0.9.15 (installed build), repo at `879dc4eb4`
- macOS (darwin arm64), Node 22.22.1

uppifyagency:
Closes #2666

Submitted for maintainer review — accept, request changes, or reject as you see fit.

## Problem

A workflow run dispatched immediately after `/workflow reload` can resolve its name against the pre-reload registry and execute the old module. Observed live on atomic 0.9.15 through the atomic-remote bridge: the bridge writes a workflow file, injects `/workflow reload`, then `/workflow <name>`; the session showed reload "generation 2" (a report that had read the file before it was rewritten), the run executing the pre-reload module, and the requested reload applying as "generation 3" only after the run had started. Full trace in #2666.

## Mechanism

`ensureWorkflowResourcesLoaded()` in `packages/workflows/src/extension/extension-runtime-state.ts` waited only while **no discovery existed at all** (`discoveryRef.current === null`). Once any first discovery had applied, it returned immediately, even when `lazyDiscoveryPromise` held an in-flight reload — so every name-resolution path (slash dispatch, tool control, durable resume) could read the registry the reload was about to replace.

## Fix

Resolution now awaits an in-flight reload as well as a missing first discovery:

- if `lazyDiscoveryPromise` is pending, await it and re-check (a further reload requested meanwhile is awaited too);
- a reload that **fails while a registry already exists** does not throw at the unrelated resolution caller — the reload's own caller already reports the failure, and resolution proceeds on the retained registry (previous throw semantics are kept for the no-registry-at-all case);
- generation/supersession checks are unchanged.

No API change; behavior-only. Changelog entry added under `packages/workflows/CHANGELOG.md` → `[Unreleased]` → `Fixed`.

## Tests

- New `test/unit/workflow-resolution-awaits-inflight-reload.test.ts`, deterministic (gated `refreshWorkflowResources`, no sleeps):
  - red on `main`: `ensureWorkflowResourcesLoaded()` settles while a reload is still in flight; green with the fix, and the registry seen after the await contains the newly discovered workflow;
  - a failed in-flight reload with a retained registry does not reject resolution.
- `test/unit/workflow-reload-rediscovery.test.ts` (coalescing, session boundaries, failure retention): 10/10 green with the fix.
- Full `npm run test:unit`: at the fix commit in a space-free worktree, 696/697 files green (6979 passed, 24 skipped); the one failure is `test/unit/install-shell.test.ts` with `spawnSync /bin/sh ETIMEDOUT` under full-suite load, 25/25 green rerun in isolation — unrelated to this change.
- `npm run check`: biome (`--error-on-warnings`), root `tsc --noEmit`, the coding-agent `tsgo` pass, and the shrinkwrap check all pass. The `npm run check` chain as a whole currently fails on pristine `main` too: its pi-ai build step refetches models.dev, and today's snapshot drops the `openai-completions` stream key that `packages/ai/src/providers/cloudflare-ai-gateway.ts:19` references (TS2353). Reproduced in a clean `main` worktree; unrelated to this change.

Test-environment note: a checkout path containing a space (e.g. `…/PROGETTI ANTYGRAVITY/atomic`) makes 19 pre-existing prompt-callsite/replay tests fail on `main` too (`normalizedPromptCallsiteFrame` truncates the frame at the space). The evidence above was produced in a space-free worktree; happy to file that separately if useful.

## Live verification

Deliberately not performed: the installed Atomic runs the official published build, which cannot include this fix, and this change is submitted for review rather than run as a local divergence. The downstream mitigation (a flat `RELOAD_SETTLE_MS = 5000` in the atomic-remote bridge between the two injections) stays in place, version-gated, until a release containing this fix is the oldest supported engine.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
<a href="https://app.blacksmith.sh/bastani-inc/codesmith/atomic/pr/2667"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790249138&installation_model_id=10562&pr_number=2667&repository=bastani-inc%2Fatomic&return_to=https%3A%2F%2Fgithub.com%2Fbastani-inc%2Fatomic%2Fpull%2F2667&signature=c39b40443d5e3e4663adfc60793f680f123307825dbaeff56e3a5f10258762ed"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

<h3>Greptile Summary</h3>

## Summary
Workflow name resolution now waits for an explicitly requested reload before reading the registry, including when background discovery is disabled. Reload failures continue to retain the existing registry for later resolution.

The focused workflow reload regression coverage passed all four scenarios, including the disabled-discovery path.

## Merge safety
Safe to merge.

No blocking failure remains.

The reload-resolution behavior is covered for successful reloads, failed reloads, and disabled background discovery, with no accepted blocking findings.

**What T-Rex did**
- I inspected ensureWorkflowResourcesLoaded() and confirmed that an explicit reload publishes lazyDiscoveryPromise before disabled background discovery is considered.
- I executed the unit test suite with Vitest against workflow-resolution-awaits-inflight-reload.test.ts, and all four tests passed, including the explicit-reload and disabled-discovery scenarios.
- I reviewed how ensureWorkflowResourcesLoaded() uses lazyDiscoveryPromise and verified that an explicit reload publishes it, while the disabled-discovery flag prevents starting a new discovery when none is pending, so the resolver awaits the explicit reload and observes the replacement registry.

<a href="https://app.greptile.com/trex/runs/20732813/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

<sub>Reviews (4): Last reviewed commit: ["Merge remote-tracking branch &#39;origin/mai..."](https://github.com/bastani-inc/atomic/commit/cc9555efbfcd2eca6ec3ab7056762007c87ecc00) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56620429)</sub>


flora131:
Thanks @uppifyagency  — we'll take it over from here to resolve the conflicts/comments, run CI, and get this merged. We'll make sure you receive proper attribution for your work so you get the credit.

Issue #2716, opened 2026-08-27
[PROPOSAL] Workflows SDK must use the exact extension/package design: real compiled specifier, no virtual-only module, no ambient bridge


lavaman131:
## Summary

The workflows SDK is the only surface in the ecosystem that consumers cannot typecheck like a normal package, and it is currently breaking every consumer's `tsc` run. The fix should not be a packaging patch: **the workflows SDK should adopt the exact same design extensions and packages already use** — a real published specifier resolved by `tsc` to a compiled `.js` + `.d.ts` closure, with the jiti loader aliasing that same specifier at runtime. No virtual-only specifier, no ambient declaration bridge, no triple-slash injection, no raw implementation `.ts` in the published artifact.

## What is broken today (0.9.16-alpha.6 and alpha.7)

The workflows `src/` tree was bundled into `dist/builtin/workflows/src/index.bundle.mjs` and the loose files deleted (0.9.14-alpha.6 was the last release shipping the full tree). Two sets of surviving files still import the deleted paths:

**1. Hard errors in every consumer's `tsc` run.** `dist/index.d.ts` begins with `/// <reference path="./builtin/workflows/ambient.d.ts" />`. That pulls `ambient.d.ts` -> `@bastani/atomic/workflows/builtin` -> `builtin/index.d.ts`, whose relative imports (`./goal.js`, `./ralph.js`, `./tournament.js`, ...) resolve to the still-shipped raw implementation `.ts` files (TS prefers source over `.d.ts`). Those files import `../src/shared/types.js`, `../src/authoring/workflow.js`, `../src/shared/workflow-artifacts.js` — all deleted. `skipLibCheck` cannot shield them because they are `.ts`, not `.d.ts`.

**2. Silent `any`-poisoning.** `dist/builtin/workflows/src/authoring.d.ts` imports `./shared/workflow-authoring-types.js` and `./shared/workflow-heartbeat-contract.js`, but `shared/` now ships only `authoring-contract.d.ts`. Under `skipLibCheck` the unresolved imports do not error — the exported types collapse to `any`, so workflow authors silently lose type safety (e.g. `ctx.parallel(...)` results, `stage.prompt(...)` turns) and then fail their own `noImplicitAny`.

## Minimal repro (46 errors from a 2-line consumer)

```jsonc
// package.json
{
  "name": "repro", "private": true, "type": "module",
  "dependencies": { "@bastani/atomic": "0.9.16-alpha.7" },
[...]

Issue #1208, opened 2026-06-03
Workflow-providing packages cannot type-check `@bastani/workflows` (no externally-resolvable SDK types)


lavaman131:
## Summary

Atomic packages register workflows through the `atomic.workflows` manifest config,
and those workflows are **discovered and loaded seamlessly at runtime**. However,
workflow source files in an *installed* package **cannot be type-checked**, because
the `@bastani/workflows` SDK specifier they import has **no types resolvable outside
Atomic's own workspace**.

As a result, third-party workflow packages are forced to hand-author a
`declare module "@bastani/workflows"` ambient shim (or a `tsconfig` `paths` alias into
Atomic's internal `dist`) purely to make `tsc` pass. This directly contradicts the
documentation, which states no shim is required.

The runtime is **not** the problem — it already works. The gap is purely **static type
resolution** for packages that depend on `@bastani/atomic` from npm/git rather than
living inside the Atomic monorepo.

## Environment

- `@bastani/atomic`: `0.8.24-alpha.1`
- `moduleResolution`: `NodeNext` (also reproduces on `Bundler`)
- Package depends on `@bastani/atomic` only (the supported, documented setup).

## Background: why every *other* package resource is seamless

Atomic packages declare resources in the `atomic` manifest, e.g.:

```jsonc
"atomic": {
  "extensions": ["./src/index.ts"],
  "skills": ["./skills"],
  "workflows": ["./workflows"]
}
```

Atomic discovers each resource by path and loads it with jiti. The reason most
resources are seamless for TypeScript authors:

- **Extensions** import the **real, published, typed** package:
  `import { … } from "@bastani/atomic"` resolves to `@bastani/atomic`'s shipped
  `dist/index.d.ts`. Fully typed, zero config.
- **Skills / prompt templates / themes** are asset directories — no SDK import, so
  no typing concern.

Workflows are the **only** resource whose source files import a **virtual** SDK
specifier that is not the installed package.

## Where the seam is: the workflow SDK is a runtime-only virtual module

Workflow files are authored as:

```ts
import { defineWorkflow, Type } from "@bastani/workflows";
export default defineWorkflow("…")./* .input/.output/.run */.compile();
```

At load time, Atomic's workflow module loader registers `@bastani/workflows` (and its
`builtin/*` submodules) as **in-memory jiti virtual modules** and maps them to the
bundled SDK surface:

- `dist/builtin/workflows/src/extension/workflow-module-loader.ts`
[...]

PR #2718, opened 2026-08-27
fix(build): ship the intercom broker's import closure so intercom and subagent work in installed packages

flora131:
## Problem

In the published `0.9.16-alpha.7` package, `dist/builtin/intercom/` was missing `group.ts`, `session-target.ts`, `source-ownership.ts`, and `types.ts`.

`INSTALLED_KEEP_PREFIXES.intercom` kept `broker/` but none of the sibling modules the broker imports. The broker is spawned as its own jiti process with plain filesystem resolution — no host alias, no `virtualModules` fallback — so it crashed at `require.resolve` on **every** startup:

```
Cannot find module '../session-target.js' imported from
  .../dist/builtin/intercom/broker/send-handler.ts
  code: 'MODULE_NOT_FOUND'
```

That made the `intercom` tool completely unusable, and because subagent results are delivered over intercom, it made the `subagent` tool completely unusable too — for every agent, in an installed package.

It shipped because intercom is lazy and tool-driven: the broker only spawns on first intercom use, so CLI-boot smoke tests never touch it. `send-handler.ts` gained the `../session-target.js` import in 518a952071 and nothing caught the missing sibling.

## Change

The copy step now **derives** the closure instead of listing filenames. `packages/coding-agent/scripts/derive-import-closure.ts` parses each kept raw `.ts` file with `@babel/parser` and walks its relative import edges transitively; `pruneInstalledPackage` keeps anything in that closure in addition to the existing keep-list. A new cross-directory import inside `broker/` is pulled into the shipped tree with no keep-list edit.

Edges followed: `import`/`export … from`, `export *`, dynamic `import()`, type-position `import("…")`, and relative `require()`.

**The walker fails closed everywhere it cannot prove an edge is safe.** It throws — rather than silently pruning a target — when a specifier is unresolved, escapes the package root, resolves beneath an excluded path, or is a non-literal `import()`/`require()` argument. Packaging cannot prove a computed edge is safe, so it refuses to guess.

Scope is deliberately narrow. Only `intercom: ["broker/"]` participates, because it is the one kept raw tree that runs directly off the filesystem in its own process. **Workflows is excluded on purpose**: its installed authoring surface prunes raw sources for #1208, and tracing it would resurrect them. Its shipped tree is path-identical at 278 files after this change.

Exclusions are unchanged and shared with the copy step: `node_modules`, `test`/`tests`, `*.test.ts`, `*.spec.ts`, `*.map`, and the usual dot-directories.

## Verification

Captured on the implementation run at `1c79fe535f`:

- `bun run copy-builtin-packages` → intercom **26** files (the four missing modules present); workflows **278**, `diff` against baseline empty
- Scan over the built tree: `total unresolved: 0` relative specifiers under `dist/builtin/intercom`
- Negative case, at the compiled-binary level: deleting `session-target.ts` from the built tree reproduces the literal `Cannot find module '../session-target.js'` through `Module._resolveFilename`; restoring via the copy step passes again
- End-to-end through a binary built with `npm run build:binary --workspace=@bastani/atomic`: a real prompt drove the `intercom` tool to `Connected: Yes`, and a subagent completed with `Delivered single subagent result via intercom` and zero `intercom_result_error`. Credential-free legs ran on a crabbox linux runner; the model-driven legs ran locally on darwin-arm64, because `.crabbox.yaml` `env.allow` is `[CI, NODE_OPTIONS]` and no credential may reach the runner. Full transcripts are posted as a comment below.
- `npm run test:unit -- test/unit/builtin-copy-import-closure.test.ts` — 5 passed
- `npm run test:unit -- test/unit/intercom-broker-runtime-graph.test.ts` — 24 passed
- `npm run check` — passes

`test/unit/builtin-copy-import-closure.test.ts` covers the derivation against fixtures (transitive closure, exclusions, escape, excluded-path, non-literal specifiers) and asserts against the real built tree that every relative import in every shipped raw `.ts` file resolves, and that the four intercom modules are present. The tree assertion fails when any of the four is removed.

Only scoped suites plus `npm run check` were run, per the run's standing instruction; full CI has not run on both platforms.

## Known, non-blocking

Review flagged two P3s that are latent today and left for follow-up rather than widened into this fix:

- `deriveImportClosure` recurses only into `.ts`. A `.mts`/`.cts`/`.tsx` dependency is copied but not re-entered, so its own imports could be pruned. No such file exists in any participating package today.
- The `CallExpression`/`Import`-callee fallback for dynamic `import()` is unreachable on current Babel majors.

Separately, `dist/builtin/workflows/builtin/*.ts` still has 33 pre-existing unresolved relative imports of the same class. #2716 replaces that surface in a separate stacked change.

---
<a href="https://app.blacksmith.sh/bastani-inc/codesmith/atomic/pr/2718"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790388723&installation_model_id=10562&pr_number=2718&repository=bastani-inc%2Fatomic&return_to=https%3A%2F%2Fgithub.com%2Fbastani-inc%2Fatomic%2Fpull%2F2718&signature=58f75532530a1d5699462c59bada1f6a44e9dc6a3210ca4c39fdb4726fe56d41"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

<h3>Greptile Summary</h3>

This change ensures installed Intercom packages retain the complete set of statically referenced runtime TypeScript files, including dependencies reached through `require()` and TypeScript import types. The potential failure where installed Intercom omits broker dependencies was disproved by running the import-closure harness against the real package: it found and retained the 20-file closure, and it rejected computed and excluded-path imports as intended. No defects were found; the change is safe to merge.

The installed Intercom import closure retains the tested runtime dependencies and safely rejects unsupported dynamic or excluded import edges.

No review findings remain after an executable check exercised the real Intercom dependency closure, installed-file projection, and failure paths for computed and excluded imports.

**Files Needing Attention:** No files need follow-up attention.

**What T-Rex did**
- Ran the end-to-end Intercom packaging walker harness before installing the projection against the real packages tree and observed a 20-file runtime closure with all relative edges resolving.
- Reran the harness after the projection installation and observed that the installed projection retained those 20 files, while computed import() and require() paths and imports into excluded paths were rejected.
- Validated the derive-import-closure.ts region (lines 145-188) against the harness to confirm expected behavior.

<a href="https://app.greptile.com/trex/runs/20946248/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(build): fail closed on builtin impor..."](https://github.com/bastani-inc/atomic/commit/1c79fe535f06aee23c112bad8587566f589a1f1a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57349344)</sub>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed installed packages where the intercom broker crashed at startup, leaving both the `intercom` and `subagent` tools entirely non-functional, and fixed the workflows SDK `types` condition pointing to a source file that is not shipped.
+
 ## [0.9.16-alpha.6] - 2026-08-26
 
 ### Changed
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixed
 
-- Fixed installed packages where the intercom broker crashed at startup, leaving both the `intercom` and `subagent` tools entirely non-functional, and fixed the workflows SDK `types` condition pointing to a source file that is not shipped.
+- Fixed installed packages where the intercom broker failed to start on every launch, leaving both the `intercom` and `subagent` tools entirely non-functional.
 
 ## [0.9.16-alpha.6] - 2026-08-26
 
--- a/packages/coding-agent/scripts/copy-builtin-packages.ts
+++ b/packages/coding-agent/scripts/copy-builtin-packages.ts
@@ -47,7 +47,6 @@ const packagesRoot = resolve(packageRoot, "..");
 const workflowsDistDir = join(distBuiltinDir, "workflows");
 const workflowsAuthoringTsconfig = join(packageRoot, "tsconfig.workflows-types.json");
 const workflowsAmbientDts = join(workflowsDistDir, "ambient.d.ts");
-const workflowsAuthoringDeclaration = "src/authoring.d.ts";
 const workflowsAmbientReference = '/// <reference path="./builtin/workflows/ambient.d.ts" />';
 // Raw authoring-surface sources that are type-only at runtime (all importers use
 // `import type`), so deleting their `.ts` copies is behaviorally inert; only the
@@ -96,7 +95,7 @@ const INSTALLED_KEEP_PREFIXES: Record<BuiltinPackageDirName, readonly string[]> 
 		"ambient.d.ts",
 		INSTALLED_EXTENSION_ENTRIES.workflows,
 		WORKFLOWS_SDK_BUNDLE_ENTRY,
-		workflowsAuthoringDeclaration,
+		"src/authoring.d.ts",
 		"src/shared/authoring-contract.d.ts",
 		"builtin/",
 		"skills/",
@@ -310,7 +309,6 @@ function pointWorkflowsSdkAtBundle(pkg: Record<string, unknown>): void {
 		| undefined;
 	if (exportsField?.["."]) {
 		exportsField["."].default = `./${WORKFLOWS_SDK_BUNDLE_ENTRY}`;
-		exportsField["."].types = `./${workflowsAuthoringDeclaration}`;
 	}
 }
 
--- a/packages/coding-agent/scripts/copy-builtin-packages.ts
+++ b/packages/coding-agent/scripts/copy-builtin-packages.ts
@@ -15,6 +15,11 @@ import {
 	WORKFLOWS_SDK_BUNDLE_ENTRY,
 	type BuiltinPackageDirName,
 } from "../src/core/builtin-install-layout.ts";
+import {
+	deriveImportClosure,
+	INSTALLED_IMPORT_CLOSURE_ROOTS,
+	shouldSkipBuiltinCopyEntry,
+} from "./derive-import-closure.js";
 
 interface BuiltinCopy {
 	label: string;
@@ -42,6 +47,7 @@ const packagesRoot = resolve(packageRoot, "..");
 const workflowsDistDir = join(distBuiltinDir, "workflows");
 const workflowsAuthoringTsconfig = join(packageRoot, "tsconfig.workflows-types.json");
 const workflowsAmbientDts = join(workflowsDistDir, "ambient.d.ts");
+const workflowsAuthoringDeclaration = "src/authoring.d.ts";
 const workflowsAmbientReference = '/// <reference path="./builtin/workflows/ambient.d.ts" />';
 // Raw authoring-surface sources that are type-only at runtime (all importers use
 // `import type`), so deleting their `.ts` copies is behaviorally inert; only the
@@ -90,7 +96,7 @@ const INSTALLED_KEEP_PREFIXES: Record<BuiltinPackageDirName, readonly string[]> 
 		"ambient.d.ts",
 		INSTALLED_EXTENSION_ENTRIES.workflows,
 		WORKFLOWS_SDK_BUNDLE_ENTRY,
-		"src/authoring.d.ts",
+		workflowsAuthoringDeclaration,
 		"src/shared/authoring-contract.d.ts",
 		"builtin/",
 		"skills/",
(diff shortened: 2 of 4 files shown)


flora131:
## Compiled-binary E2E — intercom broker packaging fix

Commit `1c79fe535f`. Credential-free legs on crabbox linux runner **`silver-crab`** (lease **`cbx_730c3fb3fdbd`**, x86-64, stopped after the run). Credentialed model-driven legs on local **darwin-arm64**, because `.crabbox.yaml` `env.allow` is `[CI, NODE_OPTIONS]` and no credential may reach the runner. `amber-krill`/`cbx_717bd17dd85c` was never touched.

**Binary built with the project's own script** (`npm run build:binary --workspace=@bastani/atomic`):

```
packages/coding-agent/dist/atomic: ELF 64-bit LSB executable, x86-64, version 1 (SYSV),
dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 3.2.0,
BuildID[sha1]=2c2e51f3b6fae6722da4f7b69751a2e9467ab063, not stripped
```

**Compiled-binary broker started and served real traffic** (crabbox):

```
8381 atomic /tmp/crabbox/cbx_730c3fb3fdbd/atomic/packages/coding-agent/dist/atomic \
  --atomic-internal-intercom-broker .../dist/builtin/intercom/broker/broker.ts
REAL_DELIVERED_FRAME  {"type":"delivered","messageId":"e2e-8391-1787794985003"}
REAL_RECEIVED_MESSAGE {"type":"message","from":{"name":"sender",...},"message":{...,"content":{"text":"credential-free broker E2E"}}}
```

**Real prompt using the `intercom` tool through the compiled binary** (local, darwin-arm64):

```json
{"type":"tool_execution_end","toolName":"intercom","result":{"content":[{"type":"text",
"text":"**Intercom Status:**\nConnected: Yes\nSession ID: 1dffdf2d-...\nGroup: compiled-binary-e2e-isolated\nActive sessions in group: 1"}],"isError":false}}
```

**Subagent completed and its result was delivered over intercom** (local):

```json
{"agent":"hello-demo","task":"Return exactly RELAYED_THROUGH_INTERCOM","status":"ok",
 "envelope":"RELAYED_THROUGH_INTERCOM","finalOutput":"RELAYED_THROUGH_INTERCOM"}
```

```
Delivered single subagent result via intercom.
$ grep -c 'intercom_result_error' /tmp/e2e-subagent-relay.jsonl   -> 0
```

**Negative case — remove `session-target.ts`, same E2E fails:**

```
Atomic startup error: Cannot find module '../session-target.js' imported from
  .../dist/builtin/intercom/broker/send-handler.ts
    at Function._resolveFilename (node:internal/modules/cjs/loader:1383:15)
    at Function.resolve (node:internal/modules/helpers:157:19)
    at jitiResolve (.../node_modules/jiti/dist/jiti.cjs:1:176713)
  code: 'MODULE_NOT_FOUND'
```

Restored via `bun run copy-builtin-packages` → `RESTORED_COMPILED_BINARY_PASS` and `RESTORED_NODE_JITI_PASS`.

**Branch coverage, stated honestly.** The compiled host selects `bun-binary` and self-spawns the same executable with `--atomic-internal-intercom-broker`, but still dynamically imports the on-disk shipped `broker.ts` graph — so missing siblings **do** break the real binary. The `node`+`jiti` leg additionally covers the npm-package path where 0.9.16-alpha.7 actually crashed.

**Scoped suites on linux:** `builtin-copy-import-closure` 5/5, `intercom-broker-runtime-graph` 24/24. No full suite was run, per standing amendment.

PR #3138, opened 2026-09-20
fix(clipboard): keep remote WSL copies off the host Windows clipboard

flora131:
## Summary

Follow-up to #3137 addressing the review findings that were posted after it merged, including the comment outside the diff.

## Changes

- **Remote WSL wrote twice** (P1, `clipboard.ts:192-196`): a display-less SSH/Mosh session inside WSL without `WT_SESSION` wrote the copied text to the remote host's Windows clipboard via PowerShell and then emitted OSC 52 to the connected client. The WSL branch now requires `!remote`, so a remote session uses only the OSC 52 fallback and the text reaches one endpoint. Upstream pi ([6dff740](https://github.com/earendil-works/pi/commit/6dff740fa)) has the same double write, so this is an Atomic-side change.
- **Assertion convention** (P2 + outside-diff comment): `cache-warmer.test.ts` and `clipboard.test.ts` now use `node:assert/strict`. Both files were converted whole rather than only the lines #3137 added, so neither file mixes two assertion styles; both are already Atomic-adapted, so upstream drift is unchanged in kind. The one remaining `../src/utils/clipboard.ts` specifier becomes `.js`, matching the fix in b8f40fb0.
- New regression tests: `remote %s into WSL skips the Windows clipboard and emits OSC 52 for the client` for `SSH_CONNECTION`, `SSH_CLIENT`, and `MOSH_CONNECTION`. All three fail against the pre-fix source.
- Changelog entry under `[Unreleased] › Fixed` in `packages/coding-agent/CHANGELOG.md`.

## Notes

- `npm run check` (biome + both typecheck passes + shrinkwrap) passes; also ran in the pre-commit hook.
- `packages/coding-agent`: `clipboard.test.ts` (25) and `cache-warmer.test.ts` (6) pass.
- No user-guide changes: the docs do not describe the WSL clipboard backend.

Refs #3137

---
<a href="https://app.blacksmith.sh/bastani-inc/codesmith/atomic/pr/3138?ref=codesmith_pr_footer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792516732&installation_model_id=10562&pr_number=3138&ref=codesmith_pr_footer&repository=bastani-inc%2Fatomic&return_to=https%3A%2F%2Fgithub.com%2Fbastani-inc%2Fatomic%2Fpull%2F3138&signature=11577f0436facf0ca3bcdf4f625230f7c36f70bb008d409c5d66b69713b04673"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

<h2><a href="https://app.greptile.com/api/retrigger?id=66667233"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

Safe to merge; there are no outstanding blocking issues.

<h3>Summary</h3>

- Prevents display-less remote SSH and Mosh sessions in WSL from using Windows clipboard interop.
- Routes those remote clipboard operations through OSC 52 and adds regression coverage plus a changelog entry.

<sub>Reviews (2) · Last reviewed commit: ["test(clipboard): pin OSC 52 after a succ..."](https://github.com/bastani-inc/atomic/commit/f3f38db8ca1c0c2d7adbdcd47666cf363c7d2655)</sub>


flora131:
Re the outside-diff findings (remote WSLg writes the display clipboard *and* emits OSC 52): declining the suggested `!copied` guard. That is the existing, intended remote-session behavior, not something this PR introduced.

- `if (copied && !remote) return;` has been in place since c4cb45d34 and matches upstream pi. In an SSH/Mosh session the user is at the client, which only the terminal can reach, so OSC 52 is emitted even after a successful local write. The existing test `remote native success emits OSC 52 after native write` pins exactly this; requiring `!copied` would fail it and regress every remote user with a forwarded or remote display.
- It differs from the Windows interop case fixed in 66f6fc5: PowerShell was a last-resort fallback that a remote session gained nothing from. The display backends are the primary path.

f3f38db8 spells this out in the comment above the condition and adds `remote WSLg session writes the display clipboard and still emits OSC 52 for the client` so the behavior is not mistaken for the interop double write again.

Issue #2824, opened 2026-09-03
Workflow BACKGROUND panel needs manual collapse and bounded overflow navigation


sina85:
## Summary

The persistent `BACKGROUND` workflow widget grows by roughly three terminal rows per top-level run at widths of 80 columns or more. With several concurrent workflows it consumes most of the fixed editor/status dock and leaves too little room for the main transcript. The widget cannot be collapsed, minimized, focused, or scrolled.

A screenshot with seven running workflows demonstrates the problem: the panel occupies about 22 rows below the editor while the main chat is still active.

## Verified current behavior

- `packages/workflows/src/tui/widget.ts` renders every qualifying top-level run in the wide layout.
- The only compact path is the automatic `< 80` column breakpoint.
- The mounted widget has no input handler or user-controlled expansion state.
- `test/unit/widget-rendering.test.ts` intentionally requires more than four concurrent runs to render without truncation.
- #2134 removed the old four-run cap. Its PR notes documented the consequence: 40 concurrent runs produced a 121-line panel.

### Manual compact mode

Add a dedicated configurable keyboard action that toggles the `BACKGROUND` widget between expanded and compact modes.

Compact mode should remain one line and show at least:

- total visible workflow count;
- status/attention counts already carried by the current narrow summary;
- the largest elapsed duration among visible runs, so long-running work remains glanceable.

Starting or completing a workflow must not force a manually compacted panel open again.

### Bounded expanded mode

Keep the renderer's complete run list so the existing no-truncation contract remains valid, but bound the mounted widget to the available terminal height. If all rows do not fit, users must still be able to reach every run through a focused scrollable view.

Follow Atomic's existing navigation conventions:

- `PageUp` / `PageDown` for pages;
- arrows for rows or lines;
- `Home` / `End` for boundaries;
- mouse wheel where supported;
- an explicit hint when more runs exist outside the visible window.

Do not assign `Ctrl+D` or `Ctrl+U` as defaults. In the main editor those already mean exit/delete-forward and delete-to-line-start. Any new action should participate in Atomic's namespaced, user-configurable keybinding system rather than hard-coding a chord.

The existing F2 workflow graph may be reused as the full focused/scrollable surface if that provides a cleaner ownership model than making the below-editor widget focusable.

## Acceptance criteria

- A keyboard action toggles the persistent panel between expanded and compact modes.
- Compact mode shows `N` workflows and the maximum elapsed duration.
[...]

PR #3153, opened 2026-09-20
feat(workflows): show resumable failures as warning-colored `failed · resumable`

### LINKED ISSUE #2565: workflows: show resumable failures as warning-colored `failed · resumable`
## What do you want to change?

Distinguish **resumable workflow stops** from true terminal failures in the TUI without adding a new workflow state.

When a top-level workflow has a failed outcome but remains eligible for resume, show it as **`failed · resumable`** with the existing warning/yellow tone instead of the error/red tone. If the run already has a more specific label, such as `blocked` or `budget_exceeded`, keep that label, append the resumable cue where useful, and use the same warning tone.

Apply this rule across run-level workflow UI surfaces, including:

- main-chat lifecycle notices;
- the background workflow widget;
- workflow status lists and run details;
- graph headers and other run-level summaries;
- workflow resume/session pickers.

Keep true, non-resumable failures red. Do not recolor ordinary failed stage or tool nodes merely because the containing run can resume; those still represent real node failures.

## Why?

A workflow may stop because it reached an expected safety cap rather than because its work became unrecoverable. Examples include:

- a bounded review or verification loop reaching its maximum iteration count;
- a workflow reaching another author-defined loop cap;
- a duration, token, or cost budget being exhausted and waiting for a human to raise the budget.

Some capped workflows may be resumed by higher-level automation, while others wait for human action. In both cases, the durable work remains available and the run can continue. Showing `WORKFLOW FAILED` in red makes the stop look final and suggests that work was lost or cannot continue.

The UI already has warning semantics for paused and blocked work. A warning-colored resumable cue states the difference without expanding the state machine or adding a second class of workflow status for users to learn.

Related: #2212 defines budget exhaustion as a resumable `budget_exceeded` outcome rather than a true failure.

## How? (optional)

Keep this as a presentation-only change:

1. Reuse the existing authoritative resume eligibility rules (`isWorkflowRunResumable(...)` and the run's durable/resumable metadata) rather than inferring the result from an error string or one specific cap.
2. Centralize the run-level label/tone decision so notices, widgets, lists, graph headers, details, and pickers do not drift.
3. For a resume-eligible failed run, return the label `failed · resumable` and the warning tone. Preserve specific non-failure labels such as `blocked` or `budget_exceeded` while making their resumability clear.
4. Leave the persisted status, lifecycle transitions, resume behavior, and automatic follow-up behavior unchanged.

### Acceptance criteria

- [ ] A resume-eligible failed workflow renders as `failed · resumable` in warning/yellow across all run-level workflow UI surfaces.
- [ ] A resume-eligible named stop, such as `blocked` or `budget_exceeded`, keeps its specific label, makes resumability clear, and uses the warning tone.
- [ ] A non-resumable failed workflow continues to render as `failed` in error/red.
- [ ] Failed stage and tool nodes keep their current error/red treatment.
- [ ] The change adds no persisted workflow status and does not alter engine or auto-resume behavior.
- [ ] Focused tests cover the shared presentation rule and the main run-level renderers.

MarkAronov:
Closes #2565

## Problem

A failed workflow run is painted red `failed` on every run-level surface: the `/workflow status` list, the run detail screen, the graph header pill, the resume and connect pickers, and the main-chat notice. Red reads as terminal, but most failed runs can be resumed with `/workflow resume`, which restarts from the failed stage with the completed stages replayed from their checkpoints. Nothing on screen says so, so people start over.

## Solution

Runs that are eligible to resume show as `failed · resumable` in the warning tone. Red stays for runs that cannot be resumed. The specific words survive with the cue added (`blocked · resumable`, `budget_exceeded · resumable`), an actively resuming run shows its active state, and stage and tool nodes stay red. Presentation only: what a run can do is unchanged.

Per @flora131's guidance on the issue, eligibility reuses the existing resume checks rather than the engine's `resumable` flag alone, the persisted `resumable` field keeps its meaning, and mixed summaries read `N resumable and N terminal failure(s)`.

### How it works

- **Helper** (`tui/run-outcome-presentation.ts`): `runOutcomePresentation({ status, resumable, budgetExceeded? })` returns `{ status, cue, label, tone }` from one table, so every surface renders the same words and tone. `status` and `cue` come back separately so a surface can interleave its own fields.
- **Eligibility** is two existing checks in the order `/workflow resume` runs them: `isWorkflowRunResumable` on the run's resume candidate, then `resolveResumeStage`, the restart-point check, which needs a failed stage, a system-owned budget stop, or a tool frontier. A run with no restart point (one that failed before creating a stage, for instance) stays red with the existing guidance to start a new workflow. Surfaces that render per frame ask only for failed, blocked and crashed rows, so running and completed runs never pay for the probe; the pickers pass their per-revision candidate cache.
- **Reusing the restart-point check** needed only a move. `resolveResumeStage` and its three helpers were private to the extension runtime factory but use nothing from it beyond the process-global durable backend. They now live in `durable/resume-restart-point.ts` beside `tool-resume-frontier.ts`, with the backend passed the way that sibling takes it, and the runtime calls the exported function. No behaviour change.
- **Persisted structs.** `RunDetail` and the lifecycle notice are saved into the session and re-rendered without a store, so eligibility is computed when they are built and stored in a new `resumeEligible` field. `resumable` on both stays the engine or author claim; the presentation reads only `resumeEligible`.
- **Surfaces** are migrated one at a time to call the helper: run detail badge and hint row, resume picker, status list rows and counts, graph header pill (which moves from raw to effective status, so an active blocked run's header recolours from accent to warning), connect picker icon (which now builds the candidate cache for every picker intent, not only resume), and the lifecycle notice (`WORKFLOW FAILED · RESUMABLE` in warning; the model-facing `content` is unchanged).

## Decisions worth a second opinion

- **Two registry-dependent refusals are not part of eligibility.** `/workflow resume` also refuses when the workflow definition is not loaded (`workflow_not_found`) or the source run's inputs no longer validate. Both need the workflow registry, which no surface holds, and both describe the environment rather than the run's state, so a run in that situation still reads `failed · resumable`. Threading the registry through every surface and both persisted structs would be a much larger change.
- **The resume picker keeps its filter.** Rows change colour and label only; which runs it offers is unchanged, since that is resume behaviour rather than presentation. A row that fails the restart-point check shows red there, and resume refuses it as it does today.
- **The cue means eligible, not guaranteed.** The docs sentence says so.

## Commits

Draft while these land, one idea per commit:

1. **`refactor(workflows): move the resume restart-point check out of the runtime`** (done)
2. the helper and its tests, including the regression pair for a failure with and without a restart point
3. run detail carries and shows `resumeEligible`
4. live surfaces call the helper: resume picker, status list, header pill, connect picker
5. status list bucket split with the agreed wording
6. lifecycle notice
7. changelog and docs

## Deliberately not in this PR

- The background widget (`tui/widget.ts`), until #2700, #3027 and #2824 settle; a follow-up as agreed on the issue.
- Any change to resume behaviour or to the meaning of the persisted `resumable` field.
- Stage and tool node colours.

## Tests

So far, for commit 1: the nine unit files that reach `resumeFailedRun`, plus two neighbours that mention its messages.

```
test/unit/workflow-task-tail-liveness.test.ts              (pins the insufficient_state message)
test/unit/executor-run-9, executor-run-10, executor-continuation-tool-parents
test/unit/workflow-active-blocked-claim, workflow-ended-blocked-resume
test/unit/workflow-initialization-cancellation, workflow-resume-unavailable
test/unit/workflow-run-control-completed-resume
test/unit/stage-runner-controlled-pause, workflow-tool-graph
                                                            147 passed
```

Each later commit adds its own tests; the full `npm run test:unit` runs before this leaves draft.

## Verification

Based on current `main`. `tsc --noEmit` and the coding-agent `tsgo` pass, biome clean.

---
<a href="https://app.blacksmith.sh/bastani-inc/codesmith/atomic/pr/3153?ref=codesmith_pr_footer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792539205&installation_model_id=10562&pr_number=3153&ref=codesmith_pr_footer&repository=bastani-inc%2Fatomic&return_to=https%3A%2F%2Fgithub.com%2Fbastani-inc%2Fatomic%2Fpull%2F3153&signature=e39a4a7d216c03ea0bd40944c2ed26539940709e0219c9a2bf9c574110cca40d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


flora131:
@MarkAronov Your proposed eligibility boundary makes sense: use the existing run-eligibility checks plus a usable restart point. Keep definition availability and input validation outside the presentation check. Please document that the cue reflects resumable run state; the current definition and inputs can still prevent resume from starting.

Keep the resume-picker filtering unchanged. There’s value in retaining visibility of runs that pass the existing filter but lack a restart point. Show those as red `failed`, with clear guidance to inspect the retained work or start a new workflow.

Keep resume execution and persisted `resumable` semantics unchanged, with presentation eligibility computed separately. Thanks for the helper and run-detail updates.

Please include the BACKGROUND widget in this PR so its labels and colors stay consistent with chat notices and the other run-level views. The earlier widget hold is lifted: #3094 merged the focused pending-input feature, and we have closed the older #2700 and #3027 PRs as superseded. Work against current main, preserve the pending-input behavior, and account for the compact-indicator work in #3169 when resolving overlap. There’s no need to wait for the older reconciliation work or absorb its unrelated changes. Keep the manual-collapse work in #2824 separate, and include focused coverage for the widget’s labels and colors.

